PR #139, opened 2014-12-15
Add iN synthesis with two queries (no quantifiers)

rsas:
This patch teaches Souper to infer iN integers (``-souper-infer-iN`` option) using two solver queries without quantifiers. The first query is a SAT query ``LHS == constant`` asking for a model for ``constant`` whereas the second query is a ``LHS != constant_model`` query. If the result of the latter is UNSAT, we have synthesized a valid integer.

Note that STP doesn't support SMTLIBv2 option ``:produce-models``, so we have to use z3 or cvc4.

### DIFF
--- a/lib/SMTLIB2/Solver.cpp
+++ b/lib/SMTLIB2/Solver.cpp
@@ -89,16 +89,30 @@ struct SMTLIBParser {
     return false;
   }
 
-  APInt parseModel(std::string &ErrStr) {
-    if (!consumeExpected('(', ErrStr))
-      return APInt();
-    if (!consumeExpected('(', ErrStr))
-      return APInt();
-    if (!consumeSExpr(ErrStr))
-      return APInt();
+  bool consumeBvName(std::string &ErrStr) {
+    while (Begin != End && (*Begin != ' ' && *Begin != '\n' && *Begin != '\r' &&
+                            *Begin != '\t'))
+      ++Begin;
+    if (Begin == End) {
+      ErrStr = "unexpected EOF";
+      return false;
+    }
+    return true;
+  }
 
-    if (!consumeExpected('(', ErrStr))
-      return APInt();
+  APInt parseBinaryBitVector(std::string &ErrStr) {
+    ErrStr.clear();
+    const char *NumBegin = Begin;
+    while (Begin != End && (*Begin == '0' || *Begin == '1'))
+      ++Begin;
+    const char *NumEnd = Begin;
+    unsigned Width = NumEnd - NumBegin;
+
+    return APInt(Width, StringRef(NumBegin, Width), 2);
+  }
+
+  APInt parseDecimalBitVector(std::string &ErrStr) {
+    ErrStr.clear();
     if (!consumeExpected('_', ErrStr))
       return APInt();
     if (!consumeExpected('b', ErrStr))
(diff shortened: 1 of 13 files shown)


lib/SMTLIB2/Solver.cpp
```diff
@@ -89,16 +89,35 @@ struct SMTLIBParser {
...
+      ++Begin;
+    if (Begin == End) {
+      ErrStr = "unexpected EOF";
+      return false;
+    }
+    return true;
+  }
+
+  APInt parseBinaryBitVector(std::string &ErrStr) {
+    ErrStr.clear();
```
jeroenk: I think it should still be checked that a 'b' follow the '#'?

lib/SMTLIB2/Solver.cpp
```diff
@@ -133,6 +152,42 @@ struct SMTLIBParser {
...
+    while (Begin != End && ((*Begin >= '0' && *Begin <= '9') ||
+           (*Begin >= 'a' && *Begin <= 'f') ||
+           (*Begin >= 'A' && *Begin <= 'F')))
+      ++Begin;
+    const char *NumEnd = Begin;
+
+    if (!consumeExpected(')', ErrStr))
+      return APInt();
+    if (!consumeExpected(')', ErrStr))
+      return APInt();
```
jeroenk: The above four lines now seem duplicated in the three routines to parse literals. Should these occur in parseModel instead?

lib/SMTLIB2/Solver.cpp
rsas: ``b`` is being consumed while checking for ``x`` before moving to ``parseBinaryVector``. I agree that we should explicitly check if the consumed token was ``b`` and not something else.

lib/SMTLIB2/Solver.cpp
rsas: Yes, they are.

pcc:
LGTM

lib/SMTLIB2/Solver.cpp
```diff
@@ -126,12 +140,47 @@ struct SMTLIBParser {
...
+    if (!consumeExpected('(', ErrStr))
+      return Res;
+    if (!consumeSExpr(ErrStr))
+      if (!consumeBvName(ErrStr))
+        return Res;
+    if (consumeExpected('#', ErrStr)) {
+      if (consumeExpected('x', ErrStr))
+        Res = parseHexBitVector(ErrStr);
+      else
+        Res = parseBinaryBitVector(ErrStr);
```
rsas: @jeroenk are you OK with assuming 'b' was consumed here? Otherwise, it requires bigger changes in the whole parsing code.

jeroenk: I think it should be pretty straightforward by introducing a peek method, but if the others are happy with the current version, it's probably best to get the current version merged first, before making further changes to the parsing code.

regehr:
Seems like we're OK with this then? Want to sqash and merge then Raimondas?

rsas:
Ok

rsas:
I did a couple of merges with the ``master`` already and my commits are in between. What's the easiest way now to squash the commits and keep intact with the master without doing everything manually?

rsas:
It required some merging during rebasing, but now it LGTM

PR #142, opened 2014-12-19
Simple (slow!) instruction synthesis + pass support

regehr:
I think the interesting part of this patch is the getValue() function in the Pass


lib/Extractor/Candidates.cpp
```diff
@@ -354,6 +354,9 @@ Inst *ExprBuilder::build(Value *V) {
       ; // fallthrough to return below
     }
   } else if (auto GEP = dyn_cast<GetElementPtrInst>(V)) {
+
+    return makeArrayRead(V);
```
pcc: The dead code below (and `buildGEP`) should be removed. When we do support GEP as a native instruction, it seems  that `buildGEP` would look very different.

lib/Extractor/Solver.cpp
```diff
@@ -50,27 +62,122 @@ class BaseSolver : public Solver {
...
+              { IC.getConst(APInt(I->Width, -1)), I }));
+          Guesses.push_back(IC.getInst(Inst::Sub, LHS->Width,
+              { IC.getConst(APInt(I->Width, 0)), I }));
+          if (I->Width == 16 || I->Width == 32 || I->Width == 64)
+            Guesses.push_back(IC.getInst(Inst::BSwap, LHS->Width, {I}));
+          if (I->Width == 8 || I->Width == 16 || I->Width == 32 ||
+              I->Width == 64 || I->Width == 256) {
+            Guesses.push_back(IC.getInst(Inst::CtPop, LHS->Width, {I}));
+            Guesses.push_back(IC.getInst(Inst::Cttz, LHS->Width, {I}));
+            Guesses.push_back(IC.getInst(Inst::Ctlz, LHS->Width, {I}));
```
pcc: Are bswap/ctpop/cttz/ctlz likely enough to be worth guessing every time?

regehr: Definitely not, but keep in mind that this code only runs when the user asks for it and also it's intended to be provisional and will get ripped out when we have a real synthesis strategy.

lib/Extractor/Candidates.cpp
regehr: OK I'll pull out the current GEP code and we can replace it later.

regehr:
I'll bring this back after Raimondas's synthesis patch lands

PR #144, opened 2014-12-19
default build type for update_deps.sh is now Release instead of Debug

regehr:
also, CMake for Souper now defaults to same build type

this is going to fail the CI but I did test it


regehr:
Peter, any objections to this? I don't want to merge if it's going to break the ci thing.

pcc:
I've dealt with the CI now, sorry for the delay. LGTM after squashing.

PR #146, opened 2014-12-20
Z3 can produce models

jeroenk:
The additional command line argument requires at least z3 v 4.3.2 (the latest stable version).


jeroenk:
I did a bit of digging into the spec and it seems that the proper way to represent boolean literal is to use either #x... as Z3 does, which should be a hexadecimal representation, or #b..., a binary representation. It's actually not clear to me where the (_ bv... ...) notation is coming from.

So closing this one as this means Raimondas approach is the correct one.

PR #150, opened 2015-01-07
Fix iN synthesis test case execution

rsas:
The iN synthesis test cases that were tested separately were not executed during ``make check``. The ``REQUIRES:`` directive requires the features to be separated by a comma. Also, ``FileCheck`` was missing the ``check-file``.


regehr:
LGTM

PR #149, opened 2015-01-05
Enable running of valgrind from make check

jeroenk:
Fixes #119 

I'm not following llvm's approach, as suggested by John; llvm achieves a similar effect through a hand-rolled Makefile in the test directory.

PR #151, opened 2015-01-07
Redirect diagnostic output of solver to /dev/null

jeroenk:
First attempt at doing something else with the diagnostic output produced by a solver (related to Issue #149).

I considered not redirecting stderr at all. For ```makeInternalSolverProgram``` this is easily doable. However for ```makeExternalSolverProgram``` this seems impossible, as ```ExecuteAndWait``` only accepts file names as targets for redirection and not file descriptors, where a nullptr target implies redirection to /dev/null.

### DIFF
--- a/lib/SMTLIB2/Solver.cpp
+++ b/lib/SMTLIB2/Solver.cpp
@@ -252,7 +252,8 @@ public:
     }
     ::close(OutputFD);
 
-    int ExitCode = Prog(Args, InputPath, OutputPath, Timeout);
+    int ExitCode =
+        Prog(Args, InputPath, OutputPath, /*ErrorPath=*/"/dev/null", Timeout);
 
     if (Keep) {
       llvm::errs() << "Solver input saved to " << InputPath << '\n';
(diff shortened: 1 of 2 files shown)


regehr:
Peter and Jeroen, did you two come to an agreement on whether this patch is a good idea?

lib/SMTLIB2/Solver.cpp
```diff
@@ -252,7 +252,9 @@ class ProcessSMTLIBSolver : public SMTLIBSolver {
     }
     ::close(OutputFD);
 
-    int ExitCode = Prog(Args, InputPath, OutputPath, Timeout);
+    SmallString<64> ErrorPath("/dev/null");
+
+    int ExitCode = Prog(Args, InputPath, OutputPath, ErrorPath, Timeout);
```
pcc: Can't this just be

    ..., OutputPath, /*ErrorPath=*/"/dev/null", Timeout);

?

jeroenk: Sure, that can be changed. More important point: are you happy with /dev/null?

pcc: Yes, that is fine, if that is a current limitation of the llvm api.

jeroenk:
Pull request updated following @pcc 's comment

pcc:
LGTM

Issue #154, opened 2015-01-09
possible bug


regehr:
I think we have a nasty bug. The checked_add_4() function in the code below is -- I believe -- a solid implementation of a 32-bit addition that checks for signed overflow. Souper proves that it always returns 0, which is nonsense. Here's the optimization:

```
%0 = block 3
%1:i32 = var
%2:i1 = slt 0:i32, %1
blockpc %0 0 %2 1:i1
%3:i32 = subnsw 2147483647:i32, %1
%4:i32 = var
%5:i1 = slt %3, %4
blockpc %0 0 %5 1:i1
%6:i1 = slt %1, 0:i32
blockpc %0 1 %6 1:i1
%7:i32 = subnsw 2147483648:i32, %1
%8:i1 = slt %4, %7
blockpc %0 1 %8 1:i1
%9:i32 = phi %0, 1:i32, 1:i32, 0:i32
cand %9 0:i32
```

I believe something has gone wrong here. The full code is below.  It is supposed to print "-202 1", but when compiled with souper it prints "-202 0".

```
#define MIN ((int)(1U<<31))
#define MAX ((int)((1U<<31)-1))

int checked_add_4(int a, int b, int *rp) {
  if (b > 0 && a > MAX - b) {
    *rp = (a + MIN) + (b + MIN);
    return 1;
  }
  if (b < 0 && a < MIN - b) {
    *rp = (a - MIN) + (b - MIN);
    return 1;
  }
  *rp = a + b;
  return 0;
}

#ifdef TEST

#include <limits.h>
#include <assert.h>
#include <stdio.h>

int main (void)
{
  assert(MIN==INT_MIN);
  assert(MAX==INT_MAX);
  volatile int x = INT_MAX - 100;
  volatile int y = INT_MAX - 100;
  int r, o;
  o = checked_add_4(x, y, &r);
  printf ("%d %d\n", r, o);
  return 0;
}

#endif
```

regehr:
I'm invoking sclang like this:

```
sclang -O3 foo.c
```

Here are my environment variables:

```
SOUPER_DEBUG=1
SOUPER_INFER_NOP=1
SOUPER_NO_INFER_I1=1
SOUPER_SOLVER=-z3-path=/usr/bin/z3
```

rsas:
Which branch are you using? You are guessing a nop, right? I cannot reproduce the result. If I try to infer %9 in the opt above, no RHS is found.

regehr:
[...]

PR #156, opened 2015-01-12
Added preconditions to the UBInstrs collected through blockpc

chenyang78:
In the original blockpc commit, we actually didn't consider
UB instructions on blockpc(s), because EB.getUBInstCondition was
invoked before blockpcs were processed. The problem was exposed
by the commit 22989c0 ("Handle instruction UB on the RHS correctly"),
where the order of getUBInstCondition and getBlockPCs were reversed.

Consequently, souper incorrectly produced the following smt query (simplified)
for the bug-induction opt file:

  (and (blockpcs ...) (and (ub-expr on "subnsw 2147483647:i32, %1")
                           (ub-expr on "subnsw 2147483648:i32, %1")))

This is wrong because those two instructions are conditional: they can
be considered only if the relevant blockpcs are true. So, the correct
form should be:

  (and (blockpcs ...) (and (relevant blockpc => ub-expr on "subnsw 2147483647:i32, %1")
                            ...))

where "=>" stands for "imply".

This patch fixed this issue (issue #154)

### DIFF
--- a/unittests/Extractor/ExtractorTests.cpp
+++ b/unittests/Extractor/ExtractorTests.cpp
@@ -56,7 +56,8 @@ struct ExtractorTest : testing::Test {
     CS = ExtractCandidates(&*M->begin(), IC, EBC, Opts);
     for (auto &B : CS.Blocks) {
       for (auto &R : B->Replacements) {
-        assert(R.Mapping.LHS->Width == 1);
+        if (R.Mapping.LHS->Width > 1)
+          continue;
         std::vector<Inst *>Guesses { IC.getConst(APInt(1, false)),
                                      IC.getConst(APInt(1, true)) };
         for (auto I : Guesses) {
@@ -75,6 +76,8 @@ struct ExtractorTest : testing::Test {
   bool hasCandidate(std::string Expected) {
     for (auto &B : CS.Blocks) {
       for (auto &R : B->Replacements) {
+        if (R.Mapping.LHS->Width > 1)
+          continue;
         std::string Str;
         llvm::raw_string_ostream SS(Str);
         R.print(SS, /*printNames=*/true);


lib/Extractor/KLEEBuilder.cpp
```diff
@@ -74,6 +74,10 @@ struct ExprBuilder {
   std::vector<Inst *> &ArrayVars;
   std::vector<Inst *> PhiInsts;
   UniqueNameSet ArrayNames;
+  // A flag that is used to control if the UBInst under process has
+  // any precondition, i.e., blockpc.
+  bool UBInstHasPrecondition;
+  ref<Expr> UBInstPrecondition;
```
regehr: Doesn't klee's Ref class have an isNull method?  In that case do you still need the bool?

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -778,8 +791,19 @@ void ExprBuilder::setBlockPCMap(const BlockPCs &BPCs) {
...
+    //   (1) UBInstExpr collected through blockpc, and;
+    //   (2) UBInstExpr collected through pc/lhs/rhs
+    // For the first case, UBInst(s) is conditional, i.e.,
+    // they are dependent on the fact that blockpc(s) are true.
+    if (I != PCMap.end()) {
+      UBInstHasPrecondition = true;
+      UBInstPrecondition = I->second;
+    }
+    ref<Expr> PE = getInstMapping(BPC.PC);
+    UBInstHasPrecondition = false;
```
regehr: Yeah that is a little gross, but fine with me if it's the best solution.

regehr:
Thanks Yang. OK with me but maybe Peter and Raimondas can take a quick look.

lib/Extractor/KLEEBuilder.cpp
rsas: There is a [isNull()](https://github.com/klee/klee/blob/master/include/klee/util/Ref.h#L99) method.

lib/Extractor/KLEEBuilder.cpp
rsas: I don't see a better way of doing it now, but maybe @pcc has an idea. Otherwise, fine with me.

rsas:
Yang, could you please add these two test cases too? They were both wrongly ``LGTM`` before your patch.

```
%0 = block 5
%1:i64 = var
%2:i1 = eq 18446744073709551615:i64, %1
blockpc %0 0 %2 0:i1
%3 = block 2
%4:i32 = var
%5:i1 = ult %4, 65:i32
blockpc %3 0 %5 0:i1
%6:i32 = and 63:i32, %4
%7:i1 = eq 0:i32, %6
blockpc %3 0 %7 0:i1
blockpc %3 1 %5 0:i1
%8 = block 2
%9:i64 = var
%10:i32 = subnsw 64:i32, %6
%11:i64 = zext %10
%12:i64 = phi %3, %11, 0:i64
%13:i64 = shl %9, %12
%14:i1 = eq 18446744073709551615:i64, %13
blockpc %8 0 %14 0:i1
blockpc %0 3 %5 1:i1
%15:i64 = var
%16:i32 = sub 64:i32, %4
%17:i64 = zext %16
%18:i64 = shl %15, %17
%19:i1 = eq 18446744073709551615:i64, %18
blockpc %0 3 %19 0:i1
blockpc %0 4 %5 1:i1
%20:i32 = var
%21:i64 = xor 18446744073709551615:i64, %1
%22:i64 = ctlz %21
%23:i32 = trunc %22
%24:i32 = add %20, %23
%25:i64 = xor 18446744073709551615:i64, %13
%26:i64 = ctlz %25
%27:i32 = trunc %26
%28:i32 = phi %8, %27, 64:i32
%29:i64 = xor 18446744073709551615:i64, %18
%30:i64 = ctlz %29
%31:i32 = trunc %30
%32:i32 = phi %0, %24, %20, %28, %31, 64:i32
infer %32
result %10
```
```
%0 = block 5
%1:i64 = var
%2:i1 = eq 18446744073709551615:i64, %1
blockpc %0 0 %2 0:i1
%3 = block 2
%4:i32 = var
%5:i1 = ult %4, 65:i32
blockpc %3 0 %5 0:i1
%6:i32 = and 63:i32, %4
%7:i1 = eq 0:i32, %6
blockpc %3 0 %7 0:i1
blockpc %3 1 %5 0:i1
%8 = block 2
%9:i64 = var
%10:i32 = subnsw 64:i32, %6
%11:i64 = zext %10
%12:i64 = phi %3, %11, 0:i64
%13:i64 = shl %9, %12
%14:i1 = eq 18446744073709551615:i64, %13
blockpc %8 0 %14 0:i1
blockpc %0 2 %5 1:i1
%15:i64 = var
%16:i32 = sub 64:i32, %4
%17:i64 = zext %16
%18:i64 = shl %15, %17
%19:i1 = eq 18446744073709551615:i64, %18
blockpc %0 2 %19 0:i1
blockpc %0 3 %5 1:i1
%20 = block 2
blockpc %20 0 %2 1:i1
%21:i32 = phi %3, %6, 64:i32
%22:i64 = xor 18446744073709551615:i64, %13
%23:i64 = ctlz %22
%24:i32 = trunc %23
%25:i32 = phi %8, %24, 64:i32
%26:i1 = eq %21, %25
blockpc %20 1 %26 1:i1
%27:i32 = var
%28:i64 = xor 18446744073709551615:i64, %1
%29:i64 = ctlz %28
%30:i32 = trunc %29
%31:i32 = add %27, %30
%32:i64 = xor 18446744073709551615:i64, %18
%33:i64 = ctlz %32
%34:i32 = trunc %33
%35:i32 = add 64:i32, %27
%36:i32 = phi %20, %35, %21
%37:i32 = phi %0, %31, %25, %34, 64:i32, %36
infer %37
result %4
```

rsas:
Yang, thanks for fixing this!

chenyang78:
Thanks for the review, John & Raimondas. I will update the request tonight based on your comments.

PR #158, opened 2015-01-13
Added preconditions to the UBInstrs collected through blockpc

chenyang78:
In the original blockpc commit, we actually didn't consider
UB instructions on blockpc(s), because EB.getUBInstCondition was
invoked before blockpcs were processed. The problem was exposed
by the commit 22989c0 ("Handle instruction UB on the RHS correctly"),
where the order of getUBInstCondition and getBlockPCs were reversed.

Consequently, souper incorrectly produced the following smt query (simplified)
for the bug-induction opt file:

  (and (blockpcs ...) (and (ub-expr on "subnsw 2147483647:i32, %1")
                           (ub-expr on "subnsw 2147483648:i32, %1")))

This is wrong because those two instructions are conditional: they can
be considered only if the relevant blockpcs are true. So, the correct
form should be:

  (and (blockpcs ...) (and (relevant blockpc => ub-expr on "subnsw 2147483647:i32, %1")
                            ...))

where "=>" stands for "imply".

This patch fixed this issue (issue #154)

### DIFF
--- a/lib/Extractor/KLEEBuilder.cpp
+++ b/lib/Extractor/KLEEBuilder.cpp
@@ -74,6 +74,8 @@ struct ExprBuilder {
   std::vector<Inst *> &ArrayVars;
   std::vector<Inst *> PhiInsts;
   UniqueNameSet ArrayNames;
+  // Holding the precondition, i.e. blockpc, for the UBInst under process.
+  ref<Expr> UBInstPrecondition;
 
   ref<Expr> makeSizedArrayRead(unsigned Width, StringRef Name, Inst *Origin);
   ref<Expr> addnswUB(Inst *I);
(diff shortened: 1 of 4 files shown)


chenyang78:
I am sorry that I closed the original issue via my commit by accident...

regehr:
LGTM

regehr:
Yang can you refresh? Then I'll merge unless Peter has objections.

pcc:
I don't have any specific objections, other than a vague worry about the increasing complexity of this conversion process, but I can't see a better solution at the moment.

regehr:
I am hoping this part of Souper is more or less done and then we won't have to touch it much from here on.

chenyang78:
Done. Thanks.

PR #160, opened 2015-01-16
Fix signs.opt specification encoding

rsas:
The souper encoding of ``(a<0 && b>=0) || (a>=0 && b<0)`` was not correct.


regehr:
LGTM

PR #163, opened 2015-01-23
Handle llvm.assume

jeroenk:
llvm.assume is turned into a path condition. We need to be careful here not to add the pc to a replacement in case the pc occurs after the instruction we are looking to replace. Intuitively, if the pc occurs after the instruction, then the instruction can be part of the expression forming the assumption, and in that case we would replace the assumption by llvm.assume(true), which is not very useful.

I still need to write some tests. I guess that having both at least one Extractor and one Infer test makes sense. However, it's not completely clear to me what the Extractor tests are checking for. Could someone explain?

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -69,6 +69,15 @@ void CandidateReplacement::print(llvm::raw_ostream &Out,
 
 namespace {
 
+struct CandidatePC {
+  CandidatePC(Inst *LHS, Inst *RHS, llvm::Instruction *I)
+      : Mapping(LHS, RHS), Origins(std::vector<llvm::Instruction *>(1, I)) {}
+  InstMapping Mapping;
+  std::vector<llvm::Instruction *> Origins;
+};
+
+typedef std::vector<CandidatePC> CandidatePCs;
+
 struct ExprBuilder {
   ExprBuilder(const ExprBuilderOptions &Opts, Module *M, const LoopInfo *LI,
               InstContext &IC, ExprBuilderContext &EBC)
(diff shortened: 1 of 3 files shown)


regehr:
Extractor tests are testing the part of Souper that takes LLVM IR and makes Souper out of it, so there we're either looking for the correct Souper IR to be formed or else the proper error message to be emitted.

jeroenk:
I've added two Infer tests (both a positive and a negative one), which is what made most sense to me.

In general I'm wondering whether there's an easier way to avoid insertion of path conditions in case the assume occurs after the instruction we're trying to optimise (compared to my current approach).

regehr:
Jeroen, is it OK if we back-burner this until we get some of our bugs sorted out? I really don't want to increase complexity until we get things back under control.

jeroenk:
Fine with me John.

regehr:
Well, since we're back-burnering blockpc and UB until we can figure out how to do it right, I'd like to revisit this patch. 

The only concern is I don't want to make things worse for ourselves when we go back and re-engineer some of the Souper guts. Yang and Raimondas, does it look OK to take this?

chenyang78:
I think it's fine to include the patch since it improves Souper in certain way. Because the change is localized, it should be easy to rewrite it (if needed) along the way of our re-design.

jeroenk:
@chenyang78 I don't quite see how to do this. In fact, this is the reason why I went for the solution I went for.

However, in any case I would like to keep this on the back burner for now. The reason is that the problem I ran into while doing this are related to control-flow. So, an easier solution might present itself depending on what choices we make there.

regehr:
should we revive this patch?

jeroenk:
I guess we may, but I doubt all the preconditions to do so have been satisfied. We may still want to wait until the UB issues have been addressed (and possibly any BlockPC issues, if any remain).

I also cannot say that I'm extremely confident with all the PC and BlockPC handling, so I'm not sure whether I'm the right person to pick this up.

regehr:
going to close for now-- but we want this at some point!

PR #165, opened 2015-01-25
Make UB exploitation optional (default=true)

rsas:
Optionally turn off UB exploitation via command line ``-souper-exploit-ub``. Default is ``true``.


regehr:
Can you also add an sclang flag?

rsas:
Sure.

regehr:
Do we want this option hidden? Either way, LGTM-- please squash and merge.

rsas:
Why would we hide it? I gave a try on your bump branch, gzip compiles and passes all test cases resulting in 16 optimizations.

Issue #112, opened 2014-11-07
how to do synthesis


regehr:
Code synthesis is the next major piece of work for Souper. For each iN in a program, Souper wants to see if it can be computed more cheaply. These are its building blocks:
1. synthesized constants, including undef
2. variables from the LHS
3. instructions
Souper already knows how to do 1 (though not undef yet). 2 is easy. 3 is the main trick. Here's how we'll do it.

We start out with a cost function. Although you can imagine many ways to do this, to a first approximation the cost of a RHS will be the number of instructions. So 1 and 2 in the list above represent zero-cost replacements. The set of cost-1 replacements is just the collection of LLVM instructions that input one or more synthesized constants and LHS variables, and output the desired iN, under the constraint that the LLVM type system is happy. Mostly the type system just wants bitwidths to match up, so for example if we're synthesizing an i32, there's no way that an icmp instruction can produce it. Another constraint is that we don't want to try to synthesize anything that will trivially constant-fold, such as a zext with a constant input or an add with two constant inputs.

Given a cost function, we batch up solutions in such a way that our first query will succeed iff there exists a cost-0 replacement, our second query will succeed iff there's a cost-1 replacement, etc. We issue these queries sequentially and stop when we find a solution (which is guaranteed to be the cheapest one) or else we get bored. It is easy to see that as the cost increases, the number of possible solutions grows exponentially. We need a way to prevent the query size from exploding. This slide deck outlines an encoding that will probably work for us:

http://www.eecs.berkeley.edu/~jha/pres/pldi11.pdf

The idea is to push the exponential search into the solver (which is doing a huge search anyway, so why not get it to consider a few more bits of search space).

rsas:
I think my patch #167 is ready for the initial review. Please give the synthesis a try and let me know if you encounter any problems. You must use z3 or any other solver that produces models.

rsas:
Implemented in 21ed266.

PR #167, opened 2015-01-28
Add instruction synthesis (-souper-infer-inst)

rsas:
This patch teaches Souper to synthesize program instructions from a given
component library (aka "synthesis of loop-free programs", see #112).
Currently, all instructions except of Phi are supported. ZExt/SExt insts
can extend from i1 only. Due to scalability challenges, all UB versions
(NSW, NUW, NW) are disabled. Also, when using all available components
in the library, Souper will very likely run out of memory. This ought
to change in the future as our synthesis algorithm improves. Until then,
the users are encouraged to specify the components to synthesize from via
the API or command line. Here are some API usage examples,
see lib/Extractor/Solver.cpp:

```
// Use three components
std::vector<Inst::Kind> UserComps = { Inst::And, Inst::Const, Inst::Mul };
InstSynthesis IS(/*Comps=*/&UserComps);

// Use three components, but the output should contain 2 components at most
InstSynthesis IS(/*Comps=*/&UserComps, /*MaxCompNum=*/2);

// Use all components (not recommended)
InstSynthesis IS();
```

An easier way to experiment with synthesis is to use command line arguments
that override the API, e.g.:

```
-souper-synthesis-comps=and,const,mul,const,xor
-souper-synthesis-comp-num=3
```

For constant synthesis, use ``-souper-synthesis-comps=const``

For nop synthesis, use ``-souper-synthesis-comp-num=0`` (comps are ignored)

Turn on verbose debugging using ``-souper-debug-synthesis``

Activate synthesis using ``-souper-infer-inst``

The synthesized instructions might contain useless instructions, that are
yet to be eliminated from the final program.

Have a look at test cases in ``test/Infer/*-syn.opt`` and ``test/Infer/HackersDelight/``.
Running synthesis tests takes longer, thus, pass ``-DTEST_SYNTHESIS=1`` to CMake
to activate them.


regehr:
Cool!

rsas:
I started to experiment with synthesis using gzip and identified a bug in my constant synthesis mode that requires more thinking. I would like to leave this pull request open for general comments and fix the bugs incrementally.

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,956 @@
...
+      Query = IC.getInst(Inst::And, 1, {Query, Copy});
+    }
+    // Solve the synthesis constraint.
+    // Each solution corresponds to a syntactically distinct and well-formed
+    // straight-line program obtained by composition of given components
+    ModelInsts.clear();
+    ModelVals.clear();
+    InstMapping Mapping(Query, IC.getConst(APInt(1, true)));
+    // Negate the query to get a SAT model.
+    // Don't use original BPCs/PCs, the are useless
```
jeroenk: "the are useless" -> "they are useless"

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,956 @@
...
+        llvm::outs() << "synthesis succeeded\n";
+      RHS = Cand;
+      return EC;
+    }
+
+    if (DebugSynthesis)
+      llvm::outs() << "didn't work for all inputs, counterexample(s):\n";
+    // Parse input counterexamples from the model
+    std::map<Inst *, Inst *> Replacements;
+    for (unsigned J = 0; J != ModelInsts.size(); ++J) {
```
jeroenk: Generic loop counters like i, j, ..., are written in lower case in other parts of the code base.

include/souper/Infer/InstSynthesis.h
```diff
@@ -0,0 +1,294 @@
...
+  /// Number of components + N
+  unsigned M = 0;
+  /// A mapping from a location variable to a concrete component instance,
+  /// namely created instruction
+  std::map<LocVar, Inst *> CompInstMap;
+  /// Default component inst width
+  unsigned DefaultInstWidth = 0;
+  /// LocInst width is fixed
+  unsigned const LocInstWidth = 32;
+  /// A mapping from a location variable's string representation to it's location.
```
jeroenk: it's -> its

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,956 @@
...
+  // Init a new set of path conditions.
+  std::vector<InstMapping> NewPCs;
+  addConstraints(NewPCs, IC);
+
+  // Create the main wiring query (aka connectivity contraint)
+  Inst *WiringQuery = getConnectivityConstraint(IC);
+
+  // Initial concrete input set S.
+  // With every new input set that proves a synthesised program is invalid,
+  // we'll have to copy WiringQuery and replace it's inputs with the new
```
jeroenk: it's -> its

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,956 @@
...
+  Inst *Ret = IC.getConst(APInt(1, true));
+
+  if (DebugSynthesis)
+    llvm::outs() << "width constraints:\n";
+  std::vector<LocInst> Tmp(P.begin(), P.end());
+  Tmp.push_back(O);
+  // Inputs
+  for (auto const &In : I) {
+    unsigned Width = CompInstMap[In.first]->Width;
+    // Compare with component inputs and it's output
```
jeroenk: it's -> its

lib/Infer/InstSynthesis.cpp
rsas: Hmm, it's looks like it's not consistent across the code base (see e.g. lib/Extractor/Solver.cpp, lib/Extractor/KLEEBuilder.cpp).

rsas:
Thanks for the comments @jeroenk!

rsas:
Constant and nop synthesis represent two special cases where the const components' outputs or the inputs are directly wired to the output. In case of invalid consts/inputs we must tell the solver explicitly to not use these in the next round. Also, during constant synthesis, just use one const component that matches the output width and don't wire inputs to the output. These were two issues that were causing bugs while performing const/nop synthesis in gzip. Now it's fixed.

rsas:
I'm doing a number of improvements in the constraint part, thus I'm closing this pull request for now. If you want to experiment with synthesis, feel free to checkout out my inst-synthesis branch.

falk-hueffner:
I tried the inst-synthesis branch. First, I needed some changes to build:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index d03cd40..2b75289 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,6 +192,7 @@ add_library(souperPass SHARED
   ${SOUPER_PARSER_FILES}
   ${SOUPER_SMTLIB2_FILES}
   ${SOUPER_TOOL_FILES}
+  ${SOUPER_INFER_FILES}
   lib/Pass/Pass.cpp
 )
 
diff --git a/lib/Infer/InstSynthesis.cpp b/lib/Infer/InstSynthesis.cpp
index d8b758b..353e8cd 100644
--- a/lib/Infer/InstSynthesis.cpp
+++ b/lib/Infer/InstSynthesis.cpp
@@ -1158,8 +1158,8 @@ Inst *InstSynthesis::createJunkFreeInst(Inst::Kind Kind, unsigned Width,
 void InstSynthesis::getInputVars(Inst *I, std::vector<Inst *> &InputVars) {
   if (I->K == Inst::Var)
     InputVars.push_back(I);
-  for (auto I : I->orderedOps())
-    getInputVars(I, InputVars);
+  for (auto Iz : I->orderedOps())
+    getInputVars(Iz, InputVars);
 }
 
 std::string InstSynthesis::getLocVarStr(const LocVar &Loc,
@@ -1256,8 +1256,8 @@ bool InstSynthesis::hasInputs(Inst *I) {
   if (I->K == Inst::Var)
     return true;
   bool Res = false;
-  for (auto I : I->orderedOps())
-    Res |= hasInputs(I);
+  for (auto Iz : I->orderedOps())
+    Res |= hasInputs(Iz);
   return Res;
 }
 ```

I'm using z3. But now I get strange results, for example:
```
$ cat test.c
_Bool f(unsigned i) {
    return i == 1 || i == 2 || i == 4 || i == 8 || i == 16;
}
$ ./sclang -mllvm -souper-infer-inst -c -Ofast -mllvm -souper-synthesis-comps=const,ctpop,and,eq /tmp/test.c && objdump -dr test.o

test.o:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <f>:
   0:	31 c0                	xor    %eax,%eax
   2:	c3                   	retq   
```
That does not seem right...

**Comment by regehr** [comment]
In general we should shun variable shadowing; I'd be in favor of a compiler flag making sure that it is a warning or even an error.

**Comment by rsas** [comment]
Thanks for the comments @falk-hueffner! It's nice to meet you here again (say "Hi" to Rolf) :-)

Back to your example. We still don't have instruction replacement support in the Souper Pass, but it will be added soon. You should experiment with ``souper`` until then:

```
$ clang -c -emit-llvm test.c -Ofast -o - | souper $SOUPER_SOLVER -souper-infer-inst -souper-synthesis-comps=const,ctpop,and,eq
; Listing valid replacements.
; Using solver: Z3 + internal cache

; Static profile 1
; Function: f
%0:i32 = var
%1:i32 = add 4294967295:i32, %0
%2:i1 = ult %1, 16:i32
pc %2 1:i1
%3:i16 = trunc %1
%4:i16 = lshr 32907:i16, %3
%5:i16 = and 1:i16, %4
%6:i1 = ne 0:i16, %5
%7:i32 = ctpop %0
%8:i1 = eq 1:i32, %7
cand %6 %8
```
Is it the result you expected?

**Comment by regehr** [comment]
Raimonads, can you add a tweak to the Pass as part of your patch that stops Souper from mis-optimizing in synthesis mode?

**Comment by regehr** [comment]
Does it also synthesize the rigtht thing when ctpop isn't available?

**Comment by rsas** [comment]
@regehr I will. What is "the right thing"?

**Comment by regehr** [comment]
((x != 0) && !(x & (x - 1)))

**Comment by regehr** [comment]
As an additional note, the job of deciding whether a particular synthesis result is profitable or not is a hard one that will take us some time to (approximately) solve. So for now, we just want to manually look at synthesis results.

**Comment by rsas** [comment]
This is "the right thing" according to the synthesis not using ctpop:

```
$ clang -c -emit-llvm test.c -Ofast -o - | souper $SOUPER_SOLVER -souper-infer-inst -souper-synthesis-comps=const,const,const,ne,ne,eq,and,and,sub
; Static profile 1
; Function: f
%0:i32 = var
%1:i32 = add 4294967295:i32, %0
%2:i1 = ult %1, 16:i32
pc %2 1:i1
%3:i16 = trunc %1
%4:i16 = lshr 32907:i16, %3
%5:i16 = and 1:i16, %4
%6:i1 = ne 0:i16, %5
%7:i32 = sub 68159584:i32, %0
%8:i32 = and %0, %7
%9:i1 = eq %0, %8
cand %6 %9
```

**Comment by regehr** [comment]
Here is Alive's version of Souper's optimization, it is moderately clever but saves just one instruction:

```
%1a = add i32 4294967295, %0
%2a = icmp ult i32 %1a, 16
br %2a, label %true, label %false
true:
%3 = trunc i32 %1a to i16
%4 = lshr i16 32907, %3
%5 = and i16 1, %4
%result = icmp ne i16 0, %5
false:
  =>
%1b = add i32 4294967295, %0
%2b = icmp ult i32 %1b, 16
br %2b, label %true, label %false
true:
%7 = sub i32 68159584, %0
%8 = and i32 %0, %7
%result = icmp eq i32 %0, %8
false:
```

It would be much more pleasing if Souper could eliminate the branch, but I think that it doesn't know how to do this. The problem is that Souper always optimizes starting at a single instruction but this function has two return instructions and Souper cannot consider both of them at the same time:

```
define zeroext i1 @f(i32 %i) #0 {
  %switch.tableidx = add i32 %i, -1
  %1 = icmp ult i32 %switch.tableidx, 16
  br i1 %1, label %switch.lookup, label %3

switch.lookup:                                    ; preds = %0
  %switch.cast = trunc i32 %switch.tableidx to i16
  %switch.downshift = lshr i16 -32629, %switch.cast
  %2 = and i16 %switch.downshift, 1
  %switch.masked = icmp ne i16 %2, 0
  ret i1 %switch.masked

; <label>:3                                       ; preds = %0
  ret i1 false
}
```

**Comment by jeroenk** [comment]
I see two possibilities of handling this: either run the mergereturn pass before running souper, or change souper's harvester to handle returns in a special way, i.e., generate a phi node that represents each possible return. Of course, the latter removed the mapping to a single instruction that we currently have.

**Comment by regehr** [comment]
Jereon, good ideas, I've added these to my list, but want to get synthesis integrated and supported by the opt pass before acting on these. In the meantime I rewrote the specification as follows:

```
define zeroext i1 @f2(i32 %i) #0 {
  %switch.tableidx = add i32 %i, -1
  %1 = icmp ult i32 %switch.tableidx, 16
  %switch.cast = trunc i32 %switch.tableidx to i16
  %switch.downshift = lshr i16 -32629, %switch.cast
  %2 = and i16 %switch.downshift, 1
  %switch.masked = icmp ne i16 %2, 0
  %r = select i1 %1, i1 %switch.masked, i1 false
  ret i1 %r
}
```
Using a command line similar to the one Raimondas shows above, Souper synthesizes this:
```
%8:i32 = sub %0, 1:i32
%9:i32 = and %0, %8
%10:i1 = eq 0:i32, %9
```

How does that look, @falk-hueffner? It may not be optimal due to the limited set of componenst, but it's pretty decent.

Does anyone know of a solution requiring only 2 LLVM instructions?

regehr:
Ach.... my flattening of the LLVM introduced undefined behavior, the synthesis isn't correct.

PR #169, opened 2015-02-07
use "exists" for environment vars instead of checking for non-zero

regehr:
value

support first and last optimization via env vars


regehr:
Raimondas can you take a quick look?

rsas:
LGTM, please rebase to master before merging.

PR #170, opened 2015-02-10
Make blockpc exploitation optional

rsas:


regehr:
Please add sclang support.

rsas:
Done.

lib/Extractor/Candidates.cpp
```diff
@@ -35,6 +36,11 @@
 #include <unordered_set>
 #include <tuple>
 
+static llvm::cl::opt<bool> ExploitBPCs(
+    "souper-exploit-bpcs",
```
chenyang78: How about souper-exploit-blockpcs? I think it doesn't hurt to use a complete name here.

rsas: Agrred.

rsas:
Is it good to squash?

chenyang78:
LGTM. Thanks.

PR #172, opened 2015-02-11
fix a use-of-unitialized variable bug that is triggered by cold external cache

regehr:
and a command line such as this:

souper  -stp-path=/usr/local/bin/stp -souper-no-infer -souper-external-cache 9gLRvwcI9i.bc


rsas:
LGTM

PR #175, opened 2015-02-14
Inst synthesis

rsas:
The synthesis code base grows and I think it's time to review this patch so that others also understand the core algorithm and its peculiarities. Please comment on everything that is unclear so I can improve both the code and the documentation. Please read the commit message from 8b5e5fd that contains instructions for synthesis invocation and usage. The main difference from that commit is that now there is no API and the synthesis is iterative, i.e., it starts with zero components (nop synthesis) and proceeds until it reaches the max num of components specified via cmd or the LHS cost minus one. If no max components were specified, it will try to use the whole component library (or all components specified via cmd).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,6 +138,15 @@ add_library(souperKVStore STATIC
   ${SOUPER_KVSTORE_FILES}
 )
 
+set(SOUPER_INFER_FILES
+  lib/Infer/InstSynthesis.cpp
+  include/souper/Infer/InstSynthesis.h
+)
+
+add_library(souperInfer STATIC
+  ${SOUPER_INFER_FILES}
+)
+
 set(SOUPER_INST_FILES
   lib/Inst/Inst.cpp
   include/souper/Inst/Inst.h
@@ -183,6 +192,7 @@ add_library(souperPass SHARED
   ${SOUPER_PARSER_FILES}
   ${SOUPER_SMTLIB2_FILES}
   ${SOUPER_TOOL_FILES}
+  ${SOUPER_INFER_FILES}
   lib/Pass/Pass.cpp
 )
 
@@ -222,7 +232,7 @@ add_executable(parser_tests
   unittests/Parser/ParserTests.cpp
 )
 
-set_target_properties(souper internal-solver-test lexer-test parser-test souper-check souperExtractor souperInst souperKVStore souperParser souperSMTLIB2 souperTool souperPass kleeExpr
+set_target_properties(souper internal-solver-test lexer-test parser-test souper-check souperExtractor souperInfer souperInst souperKVStore souperParser souperSMTLIB2 souperTool souperPass kleeExpr
   PROPERTIES COMPILE_FLAGS "${LLVM_CXXFLAGS}")
 set_target_properties(souperClangTool clang-souper
   PROPERTIES COMPILE_FLAGS "${CLANG_CXXFLAGS} ${LLVM_CXXFLAGS}")
@@ -231,7 +241,8 @@ set_target_properties(extractor_tests inst_tests parser_tests
 
 target_link_libraries(kleeExpr ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperClangTool souperExtractor souperTool ${CLANG_LIBS} ${LLVM_LIBS} ${LLVM_LDFLAGS})
-target_link_libraries(souperExtractor souperInst kleeExpr)
+target_link_libraries(souperExtractor souperInfer souperInst kleeExpr)
+target_link_libraries(souperInfer ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperInst ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperKVStore ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperParser souperInst ${LLVM_LIBS} ${LLVM_LDFLAGS})
@@ -242,13 +253,14 @@ target_link_libraries(souper souperExtractor souperKVStore souperParser souperSM
 target_link_libraries(internal-solver-test souperSMTLIB2)
 target_link_libraries(lexer-test souperParser)
 target_link_libraries(parser-test souperParser)
-target_link_libraries(souper-check souperTool souperExtractor souperKVStore souperSMTLIB2 souperParser souperInst ${HIREDIS_LIBRARY})
+target_link_libraries(souper-check souperTool souperExtractor souperKVStore souperSMTLIB2 souperParser ${HIREDIS_LIBRARY})
 target_link_libraries(clang-souper souperClangTool souperExtractor souperKVStore souperParser souperSMTLIB2 souperTool kleeExpr ${CLANG_LIBS} ${LLVM_LIBS} ${LLVM_LDFLAGS} ${HIREDIS_LIBRARY})
 target_link_libraries(extractor_tests souperExtractor ${GTEST_LIBS})
 target_link_libraries(inst_tests souperInst ${GTEST_LIBS})
 target_link_libraries(parser_tests souperParser ${GTEST_LIBS})
 
 set(TEST_SOLVER "" CACHE STRING "Solver command line argument to use in tests, e.g. -stp-path=/path/to/stp")
+set(TEST_SYNTHESIS "" CACHE STRING "Enable additional, computationally intensive synthesis tests")
 
 configure_file(
   ${CMAKE_SOURCE_DIR}/test/lit.site.cfg.in
--- a/include/souper/Infer/InstSynthesis.h
+++ b/include/souper/Infer/InstSynthesis.h
@@ -1,0 +1,297 @@
+// Copyright 2015 The Souper Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SOUPER_INST_SYNTHESIS_H
+#define SOUPER_INST_SYNTHESIS_H
+
+#include "souper/Inst/Inst.h"
+#include "souper/SMTLIB2/Solver.h"
+
+#include <map>
+#include <set>
+#include <vector>
+#include <utility>
+
+/// This class is based on the "Synthesis of Loop-free Programs" approach
+/// by Gulwani et al. presented at PLDI'11.
+/// Paper:  http://dl.acm.org/citation.cfm?id=1993506
+/// Slides: http://www.eecs.berkeley.edu/~jha/pres/pldi11.pdf
+/// The implementation is derived from the description in the paper, but it
+/// also contains additional steps that were necessary to apply this synthesis
+/// method in Souper.
+
+namespace souper {
+
+/// A location variable identifies an input variable, the output variable, or
+/// input/output of a library component. The encoding has the following
+/// semantics: If the first integer is 0, it's an input var or a constant
+/// where the second integer describes its number. If the second integer is 0,
+/// it's an output var and the first integer describes the number of a
+/// component. The output variable is encoded by second integer being 0 and
+/// the first integer being equal to the number of library components plus one.
+/// Otherwise, it's a component input where the first integer describes the
+/// number of a component and the second integer describes it's input number.
+/// Note that constant components (no inputs, one output) are not treated as
+/// ordinary library components, but rather as inputs.
+/// For example:
+/// 0_1, 0_2, 0_3 describe the input variables 1, 2, and 3.
+/// 1_0, 2_0, 3_0 describe the outputs of components 1, 2, and 3.
+/// 1_1, 2_2, 3_1 describe the first input of component one, the second input of
+/// component two, and the first input of component three.
+/// 4_0 describes the final output (assuming there are only three components)
+typedef std::pair<unsigned, unsigned> LocVar;
+
+/// A mapping from a location variable to its concrete value (Inst::Const).
+/// The values of location variables decide the interconnections between the
+/// various components. To describe a program, we have to determine which
+/// component goes on which location (line-number), and from which location
+/// (line-number of program input) does it get its input arguments. E.g.,
+/// the location mappings
+/// 0_1 -> 0
+/// 1_1 -> 0
+/// 1_0 -> 1
+/// 2_0 -> 1
+/// describe a program with one input (0_1) and one component with one
+/// input (1_1) located at line-number 0, producing output (1_0) at line-number 1
+/// which is the output of the program (2_0).
+/// Initially, the values of location variables are symbolic and it is the job
+/// of the SMT solver to figure out a valid wiring representing a valid program
+typedef std::pair<LocVar, Inst *> LocInst;
+
+/// A component is a fixed-width instruction kind
+struct Component {
+  Inst::Kind Kind;
+  unsigned Width;
+  std::vector<unsigned> OpWidths;
+};
+
+/// Unsupported components kinds
+static const std::set<Inst::Kind> UnsupportedCompKinds = {
+  Inst::Var, Inst::UntypedConst,
+  // Not implemented
+  Inst::Phi,
+  // Disabled due to scalability challenges
+  Inst::AddNSW, Inst::AddNUW, Inst::AddNW,
+  Inst::SubNSW, Inst::SubNUW, Inst::SubNW,
+  Inst::MulNSW, Inst::MulNUW, Inst::MulNW,
+  Inst::ShlNSW, Inst::ShlNUW, Inst::ShlNW,
+};
+
+/// Supported component library.
+/// The width of most of the instructions is unknown in advance, therefore, we
+/// indicate this by setting the Width to 0. Per each input of distinct width,
+/// a component of that width is instantiated.
+/// Again, note that constants are treated as ordinary inputs
+static const std::vector<Component> CompLibrary = {
+  Component{Inst::Add, 0, {0,0}},
+  Component{Inst::Sub, 0, {0,0}},
+  Component{Inst::Mul, 0, {0,0}},
+  Component{Inst::UDiv, 0, {0,0}},
+  Component{Inst::SDiv, 0, {0,0}},
+  Component{Inst::UDivExact, 0, {0,0}},
+  Component{Inst::SDivExact, 0, {0,0}},
+  Component{Inst::URem, 0, {0,0}},
+  Component{Inst::SRem, 0, {0,0}},
+  Component{Inst::And, 0, {0,0}},
+  Component{Inst::Or, 0, {0,0}},
+  Component{Inst::Xor, 0, {0,0}},
+  Component{Inst::Shl, 0, {0,0}},
+  Component{Inst::LShr, 0, {0,0}},
+  Component{Inst::LShrExact, 0, {0,0}},
+  Component{Inst::AShr, 0, {0,0}},
+  Component{Inst::AShrExact, 0, {0,0}},
+  Component{Inst::Select, 0, {1,0,0}},
+  Component{Inst::Eq, 1, {0,0}},
+  Component{Inst::Ne, 1, {0,0}},
+  Component{Inst::Ult, 1, {0,0}},
+  Component{Inst::Slt, 1, {0,0}},
+  Component{Inst::Ule, 1, {0,0}},
+  Component{Inst::Sle, 1, {0,0}},
+  //
+  Component{Inst::CtPop, 0, {0}},
+  Component{Inst::BSwap, 0, {0}},
+  Component{Inst::Cttz, 0, {0}},
+  Component{Inst::Ctlz, 0, {0}}
+};
+
+class InstSynthesis {
+public:
+  // Synthesize an instruction from the specification in LHS
+  std::error_code synthesize(SMTLIBSolver *SMTSolver,
+                             const BlockPCs &BPCs,
+                             const std::vector<InstMapping> &PCs,
+                             Inst *TargetLHS, Inst *&RHS,
+                             InstContext &IC, unsigned Timeout);
+
+private:
+  /// LHS to synthesize
+  Inst *LHS;
+  /// Replacement context for inst printing
+  ReplacementContext Context;
+  /// Components to be used
+  std::vector<Component> Comps;
+  /// Constant components
+  std::vector<Component> ConstComps;
+  /// Program inputs
+  std::vector<Inst *> Inputs;
+  std::set<unsigned> InputWidths;
+  /// Input location set I
+  std::vector<LocInst> I;
+  /// Component input location set P
+  std::vector<LocInst> P;
+  /// Component output location set R
+  std::vector<LocInst> R;
+  /// Output location O
+  LocInst O;
+  /// Location variables set L (I \cup P \cup R \cup {O}).
+  std::vector<LocInst> L;
+  /// Number of input variables
+  unsigned N;
+  /// Number of components + N
+  unsigned M;
+  /// A mapping from a location variable to a concrete component instance,
+  /// namely created instruction
+  std::map<LocVar, Inst *> CompInstMap;
+  /// Location variable's width (increase it for support of >64 comps)
+  const unsigned LocInstWidth = 6;
+  /// A mapping from a location variable's string representation to its location.
+  /// Required during model parsing
+  std::map<std::string, LocInst> LocInstMap;
+  /// Invalid wirings
+  std::set<std::pair<LocVar, LocVar>> InvalidWirings;
+
+  /// Initialize components to be used during synthesis
+  void setCompLibrary();
+
+  /// Initalize input variable locations
+  void initInputVars(InstContext &IC);
+
+  /// Add extra width manipulations components (zext/sext/trunc)
+  /// to handle varying input/output widths
+  void addZSTComps();
+
+  /// Initalize components' input locations, output locations,
+  /// and components' concrete instruction instances
+  void initComponents(InstContext &IC);
+
+  /// Initalize constant components
+  void initConstComponents(InstContext &IC);
+
+  /// Initalize output location
+  void initOutput(InstContext &IC);
+
+  /// Initalize locations L
+  void initLocations();
+
+  /// Print some debug info on initial synthesis settings
+  void printInitInfo();
+
+  /// Set particular wirings to be invalid (e.g. width mismatches).
+  /// These invalid wirings are not encoded as constraints, they are
+  /// simply skipped during connectivity constraint creation
+  void setInvalidWirings();
+
+  /// Every line in the program has at most one component.
+  /// Don't wire semantically invalid variable locations,
+  /// e.g., a component's output to its inputs
+  Inst *getConsistencyConstraint(InstContext &IC);
+
+  /// In a well-formed program, every variable is initialized
+  /// before it is used. For every component, if x is an input
+  /// of that component and y is an output of that component,
+  /// then the location l_x where the input is defined, should
+  /// be earlier than the location l_y where the component is
+  /// positioned and its output is defined
+  Inst *getAcyclicityConstraint(InstContext &IC);
+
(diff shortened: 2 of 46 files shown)


rsas:
I also highly encourage to watch the synthesis in action using the ``-souper-debug-synthesis`` flag. For examples, have a look at test cases in ``test/Infer/*-syn.*`` and ``test/Infer/HackersDelight``.

chenyang78:
Fascinating! One silly question, what's nop synthesis? Thanks.

regehr:
Yang, nop synthesis is what we call it when Souper finds an optimization that doesn't require adding any instructions at all because the required value is already available. For example:

```
%0:i32 = var
%1:i32 = lshr %0, 16:i32
%2:i64 = zext %1
%3:i64 = and 255:i64, %2
%4:i32 = lshr %0, 24:i32
%5:i64 = zext %4
%6:i64 = shlnw %5, 8:i64
%7:i64 = or %3, %6
infer %7
result %2
```

rsas:
@regehr nitpicking: The synthesis in the example above found a sequence of two instructions that was in the original LHS too. For me, nop synthesis is rather the following opt where one of the inputs is wired to the output directly:

```
%0:i32 = var
%1:i64 = var
%2:i64 = lshr %1, 3:i64
%3:i32 = trunc %2
%4:i1 = ult %0, %3
%5:i32 = sub %0, %3
%6:i32 = select %4, 0:i32, %5
%7:i1 = slt 0:i32, %6
pc %7 1:i1
%8:i1 = ult %3, %0
%9:i32 = select %8, %0, %3
infer %9
result %0
```
In the synthesis code, this is how nop synthesis is defined, i.e., find an opt using zero components.

regehr:
Well... I'm not sure it matters whether the result refers to a var or to an instruction -- either way, some LLVM instructions have been elided, as opposed to being expressed using a different instruction sequence.

include/souper/Infer/InstSynthesis.h
```diff
@@ -0,0 +1,300 @@
...
+/// ordinary library components, but rather as inputs.
+/// For example:
+/// 0_1, 0_2, 0_3 describe the input variables 1, 2, and 3.
+/// 1_0, 2_0, 3_0 describe the outputs of components 1, 2, and 3.
+/// 1_1, 2_2, 3_1 describe the first input of component one, the second input of
+/// component two, and the first input of component three.
+/// 4_0 describes the final output (assuming there are only three components)
+typedef std::pair<unsigned, unsigned> LocVar;
+
+/// A mapping from a location variable to it's concrete value (Inst::Const).
```
regehr: http://www.its-not-its.info/

rsas: Argh, @jeroenk pointed to a number of these already. I thought I got rid of these, but obviously not.

rsas:
@regehr ok, let's agree then when a sequence of synthesized instructions is in the LHS, it's a nop. It can be found using 0 or more components.

lib/Extractor/Solver.cpp
```diff
@@ -111,6 +115,13 @@ class BaseSolver : public Solver {
       }
     }
 
+    if (InferInsts && SMTSolver->supportsModels()) {
+      InstSynthesis IS;
+      EC = IS.synthesize(SMTSolver.get(), BPCs, PCs, LHS, RHS, IC, Timeout);
+      if (EC || RHS)
```
regehr: Please make sure it is OK to read RHS here (that is, make sure synthesize() has set it to 0 when synthesis failed. I ran into a bug of this kind recently.

lib/Extractor/Solver.cpp
```diff
@@ -111,6 +115,13 @@ class BaseSolver : public Solver {
...
+        return EC;
+    }
+
     RHS = 0;
```
regehr: If synthesize() has set RHS to 0 when synthesis fails, this is unnecessary code.

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+
+  // Initial concrete input set S.
+  // With every new input set that proves a synthesised program is invalid,
+  // we'll have to copy WiringQuery and replace its inputs with the new
+  // concrete values from S
+  std::vector<std::map<Inst *, Inst *>> S;
+
+  // Ask the solver for a couple of arbitrary initial inputs
+  auto InputPCs = PCs;
+  for (unsigned I = 0; I < 4; ++I) {
```
regehr: Where does 4 come from?
Be careful about variables named I since we often use that to mean an Inst *.

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+    std::vector<Inst *> ModelInsts;
+    std::vector<llvm::APInt> ModelVals;
+    InstMapping Mapping(LHS, IC.createVar(LHS->Width, "foo"));
```
regehr: Is there a better name than foo?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+    MaxCompNum = CmdMaxCompNum;
+  else
+    MaxCompNum = Comps.size();
+  if (!IgnoreCost && MaxCompNum >= LHSCost)
+    MaxCompNum = LHSCost-1;
+  // MaxCompNum can be negative when e.g. LHSCost is 0
+  if (MaxCompNum > (int)Comps.size())
+    MaxCompNum = Comps.size();
+  // Iterative synthesis loop with increasing number of components
+  for (int J = 0; J <= MaxCompNum; ++J) {
```
regehr: earlier J was unsigned, is there a reason it is signed here?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+
+      // No valid wiring exists for the target comp number
+      if (!IsSat)
+        break;
+
+      std::vector<std::pair<LocInst, LocInst>> Wiring;
+      Inst *Cand = createInstFromModel(std::make_pair(ModelInsts, ModelVals),
+                                       Wiring, IC);
+      if (!Cand) {
+        llvm::errs() << "synthesis bug: creating inst from a model failed\n";
```
regehr: If this truly should never happen, I'd prefer to crash souper when it does, instead of printing something.

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+  }
+}
+
+void InstSynthesis::setDefaultWidth(Inst *LHS) {
+  // Set the DefaultWidth of component instances to the max width seen in inputs
+  for (auto In : Inputs)
+    if (In->Width > DefaultWidth)
+      DefaultWidth = In->Width;
+  // No inputs, use output width
+  if (!DefaultWidth)
```
regehr: Can you explain how no inputs can happen? Wouldn't any Souper expression have at least one input?

include/souper/Infer/InstSynthesis.h
```diff
@@ -0,0 +1,300 @@
...
+  Inst::AddNSW, Inst::AddNUW, Inst::AddNW,
+  Inst::SubNSW, Inst::SubNUW, Inst::SubNW,
+  Inst::MulNSW, Inst::MulNUW, Inst::MulNW,
+  Inst::ShlNSW, Inst::ShlNUW, Inst::ShlNW,
+};
+
+/// Supported component library.
+/// The width of most of the instructions is unknown in advance, therefore, we
+/// indicate this by setting the Width to 0. During initialization, the width
+/// will be set to the DefaultWidth (maximum width of the input vars).
```
regehr: Do we care about mixed-width expressions? When looking at the stuff that comes from gzip, I seem to see a lot of expressions that mix different widths. Is it too expensive to instantiate components at every width that might be useful?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+    for (auto const &E : ProgramWiring) {
+      llvm::outs() << E.first << "\t";
+      for (auto const &Loc : E.second)
+        llvm::outs() << getLocVarStr(Loc) << " ";
+      llvm::outs() << "\n";
+    }
+  }
+
+  if (!CompInstMap.count(OutLoc)) {
+    llvm::errs() << "synthesis bug: output not wired\n";
```
regehr: report_fatal_error()?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+      std::vector<std::pair<LocInst, LocInst>> &Wiring,
+      InstContext &IC) {
+  std::vector<Inst *> Ops;
+
+  // Create operand instructions recursively
+  for (auto const &OpLoc : OpLocs) {
+    LocVar Match = getWiringLocVar(OpLoc, ProgramWiring);
+    assert(CompInstMap.count(Match) && "unknown matching location variable");
+    if (!CompInstMap.count(Match)) {
+      llvm::errs() << "synthesis bug: component input "
```
regehr: report_fatal_error()?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+
+  return Match;
+}
+
+Inst *InstSynthesis::createJunkFreeInst(Inst::Kind Kind, unsigned Width,
+                                        std::vector<Inst *> &Ops,
+                                        InstContext &IC) {
+  switch (Kind) {
+  case Inst::Add:
+  case Inst::AddNSW:
```
regehr: But aren't the nsw/nuw variants in the list of components not supported by synthesis above?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1264 @@
...
+  case Inst::SDiv:
+  case Inst::UDivExact:
+  case Inst::SDivExact: {
+    if (Ops[1] == IC.getConst(APInt(Width, 1)))
+      return Ops[0];
+    break;
+  }
+
+  case Inst::And:
+  case Inst::Or: {
```
regehr: Do we care to look for and/or with all 1s or all 0s?

regehr:
Well generally this looks great. Since it has minimal impact on Souper code that runs in non-synthesis mode, I don't see any reason not to take this patch. It would be good for @pcc and @chenyang78 to look it over as well. 

Once this lands, I'll revive my synthesis code in the pass. I do not believe that will be a huge amount of work.

rsas:
Thanks for the review @regehr, I'll address them on Monday.

CMakeLists.txt
```diff
@@ -242,13 +252,14 @@ target_link_libraries(souper souperExtractor souperKVStore souperParser souperSM
...
 target_link_libraries(parser-test souperParser)
-target_link_libraries(souper-check souperTool souperExtractor souperKVStore souperSMTLIB2 souperParser souperInst ${HIREDIS_LIBRARY})
+target_link_libraries(souper-check souperTool souperExtractor souperKVStore souperSMTLIB2 souperParser ${HIREDIS_LIBRARY})
 target_link_libraries(clang-souper souperClangTool souperExtractor souperKVStore souperParser souperSMTLIB2 souperTool kleeExpr ${CLANG_LIBS} ${LLVM_LIBS} ${LLVM_LDFLAGS} ${HIREDIS_LIBRARY})
 target_link_libraries(extractor_tests souperExtractor ${GTEST_LIBS})
 target_link_libraries(inst_tests souperInst ${GTEST_LIBS})
 target_link_libraries(parser_tests souperParser ${GTEST_LIBS})
 
 set(TEST_SOLVER "" CACHE STRING "Solver command line argument to use in tests, e.g. -stp-path=/path/to/stp")
+set(TEST_SYNTHESIS "" CACHE STRING "Enable program synthesis tests")
```
pcc: Why are these tests disabled by default? I would strongly prefer that have at least some test coverage for all features by default without a good reason otherwise.

rsas: The only reason is testing runtime that is typically longer during synthesis. If we don't care about that, I can enable these test cases by default.

regehr: Is a compromise available where for example you run synthesis with a very limited set of components?

rsas: Yes, I can remove a couple of Hacker's Delight examples. Nevertheless, I still want to have a big synthesis test suite for both testing and performance measurements.

lib/Extractor/Solver.cpp
rsas: Thanks, I ran into a bug recently too and now I'm setting RHS to 0 explicitly since then.

lib/Extractor/Solver.cpp
rsas: Yes, but we still have to cover the path where synthesis is not invoked at all, right?

lib/Infer/InstSynthesis.cpp
rsas: Yes, ``MaxCompNum`` becomes negative for the following LHS with cost 0:

```
%0:i8 = var
infer %0
```

MaxCompNum = LHSCost-1;
=> MaxCompNum = -1;

In this case, we don't want to enter the synthesis loop at all.

lib/Infer/InstSynthesis.cpp
rsas: It happened when I was running synthesis on a test case derived from test/Infer/bswap.ll:

```
define i32 @foo(i32 %x) {
entry:
  %swap = call i32 @llvm.bswap.i32(i32 2882343476) ;input is 0xABCD1234
  ret i32 %swap ;swapped result is 0x3412CDAB
}
```
LHS %swap has no inputs.

include/souper/Infer/InstSynthesis.h
rsas: At the moment, definitely too expensive. Let's consider this once the performance of synthesis significantly improves.

lib/Infer/InstSynthesis.cpp
rsas: They are not supported yet (scalability), but once they are, this function will handle them properly.

lib/Infer/InstSynthesis.cpp
rsas: Isn't that already captured by the lines below?

CMakeLists.txt
rsas: With the current set of 29 synthesis test cases, the testing time using four cores on my machine goes up from 15.42s to 28.20s. If you are OK with that, I'll enable the test cases by default.

regehr: It sounds like testing time will be acceptable if you move a subset into the default test config -- try to pick a collection that together stress the important features.

rsas: I'm moving 13 test cases to the default test suite while leaving others, i.e., computationally intensive ones, optional.

rsas:
Thanks for the comments @chenyang78 @pcc @regehr, I addressed all of them. If you have time, please have a look. If there are no further objections, we should merge it ASAP so that @regehr can add the Pass support.

chenyang78:
I don't have any concern at the moment. Again, thanks for putting this together, Raimondas and John.

rsas:
Thanks @chenyang78. @regehr hit a tricky bug yesterday, that I haven't seen so far. I'll have to do more testing.

rsas:
@regehr I addressed the bug, you can start testing. Please let me know once you are LGTM with the patch, so I can merge.

rsas:
@regehr I squashed the commits and ran tests in Debug/Release modes. Please have a final look and merge.

include/klee/Config/config.h
```diff
@@ -13,4 +13,4 @@
 // limitations under the License.
 
 #define LLVM_VERSION_MAJOR 3
-#define LLVM_VERSION_MINOR 6
```
regehr: The souper trunk is already at LLVM_VERSION_MINOR 7 since the last bump. Raimonads could you make sure that your patch here is against the latest master?

rsas: It is. My squash includes all master commits since my initial commit in my branch. I did a test merge on my local master and it applies cleanly.

rsas: I merged to master from time to time, that's why you see these lines.

rsas: Argh, I see your point now. I thought git would be smart enough to remove these existing commits automatically during merge, but it doesn't. I'll have to fix it first.

regehr:
I am getting this from a Debug build:

```
regehr@ulam:build{inst-synthesis}$ make check
[  2%] Built target profileRuntime
[ 45%] Built target souperPass
[ 47%] Built target souperSMTLIB2
[ 50%] Built target souperInfer
[ 52%] Built target souperInst
[ 75%] Built target kleeExpr
[ 81%] Built target souperExtractor
[ 83%] Built target souperTool
[ 85%] Built target souperKVStore
[ 87%] Built target souperParser
[ 89%] Built target souper
[ 91%] Built target parser-test
[ 93%] Built target souper-check
[ 95%] Built target extractor_tests
[ 97%] Built target inst_tests
[100%] Built target parser_tests
FAIL: Souper :: Infer/signs-syn.opt (70 of 151)
******************** TEST 'Souper :: Infer/signs-syn.opt' FAILED ********************
Script:
--
/home/regehr/souper-rsas/build/souper-check -z3-path=/usr/bin/z3 -infer-rhs -souper-infer-inst -souper-synthesis-comps=const,slt,xor /home/regehr/souper-rsas/test/Infer/signs-syn.opt > /home/regehr/souper-rsas/build/test/Infer/Output/signs-syn.opt.tmp1
/home/regehr/souper-rsas/third_party/llvm/Debug/bin/FileCheck /home/regehr/souper-rsas/test/Infer/signs-syn.opt < /home/regehr/souper-rsas/build/test/Infer/Output/signs-syn.opt.tmp1
--
Exit Code: 1

Command Output (stderr):
--
/home/regehr/souper-rsas/test/Infer/signs-syn.opt:6:10: error: expected string not found in input
; CHECK: %9:i32 = xor %0, %2
         ^
<stdin>:1:1: note: scanning from here
; Failed to infer RHS
^
<stdin>:1:6: note: possible intended match here
; Failed to infer RHS
     ^

--

********************
Testing Time: 70.59s
********************
Failing Tests (1):
    Souper :: Infer/signs-syn.opt

  Expected Passes    : 134
  Unsupported Tests  : 16
  Unexpected Failures: 1
make[3]: *** [CMakeFiles/check] Error 1
make[2]: *** [CMakeFiles/check.dir/all] Error 2
make[1]: *** [CMakeFiles/check.dir/rule] Error 2
make: *** [check] Error 2
regehr@ulam:build{inst-synthesis}$ 
```

rsas:
Weird, no failures here. z3 version?

regehr:
Weird, I don't know. My z3 was a bit old (unstable from last fall) but I upgraded to unstable as of right now (it calls itself 4.4.0) and still run into the same problem.

regehr:
Here's the raw command:

```
$ /home/regehr/souper-rsas/build/souper-check -z3-path=/usr/bin/z3 -infer-rhs -souper-infer-inst -souper-synthesis-comps=const,slt,xor /home/regehr/souper-rsas/test/Infer/signs-syn.opt
; Failed to infer RHS
```

rsas:
I fetched a fresh copy of my branch and ran the same command:

```
$ ./souper-check -z3-path=`which z3` -infer-rhs -souper-infer-inst -souper-synthesis-comps=const,slt,xor ../test/Infer/signs-syn.opt 
; RHS inferred successfully
%9:i32 = xor %0, %2
%10:i1 = slt %9, 0:i32
result %10
```
Z3 version 4.4.0 from unstable branch.

rsas:
Did you update my 'inst-synthesis' branch?

lib/Extractor/Candidates.cpp
```diff
@@ -362,8 +362,10 @@ Inst *ExprBuilder::build(Value *V) {
   } else if (auto GEP = dyn_cast<GetElementPtrInst>(V)) {
     if (isa<VectorType>(GEP->getType()))
       return makeArrayRead(V); // vector operation
-    return buildGEP(get(GEP->getOperand(0)), gep_type_begin(GEP),
-                    gep_type_end(GEP));
+    // TODO: replace with a GEP instruction
+    //return buildGEP(get(GEP->getOperand(0)), gep_type_begin(GEP),
+    //                gep_type_end(GEP));
+    return makeArrayRead(V);
```
rsas: @regehr are you OK with commenting ``buildGEP`` out in the master?

rsas:
I did a clean commit and a rebase to master. The only thing I'm not sure is the ``buildGEP`` part, see comment above. @regehr please do a fresh checkout of my ``inst-synthesis`` branch and run all test cases again. On my machine building and testing in Debug/Release modes didn't show any problems.

regehr:
The test failure was from a fresh checkout and build of your branch on my Ubuntu 14.04 machine at work. I'm home now and will see what happens here.

regehr:
I see the same thing at home. You've definitely tried a Debug build?

chenyang78:
I got the same error in the Release build (haven't tried Debug version yet)

    /home/chenyang/work/souper_fork/souper_ras/build_release/souper-check -z3-path=/home/chenyang/other_programs/z3/build/z3 -infer-rhs -souper-infer-inst -souper-synthesis-comps=const,slt,xor /home/chenyang/work/souper_fork/souper_ras/test/Infer/signs-syn.opt > /home/chenyang/work/souper_fork/souper_ras/build_release/test/Infer/Output/signs-syn.opt.tmp1
    /home/chenyang/work/souper_fork/souper_ras/third_party/llvm/Release/bin/FileCheck /home/chenyang/work/souper_fork/souper_ras/test/Infer/signs-syn.opt < /home/chenyang/work/souper_fork/souper_ras/build_release/test/Infer/Output/signs-syn.opt.tmp1
    --
    Exit Code: 1
    
    Command Output (stderr):
    --
    /home/chenyang/work/souper_fork/souper_ras/test/Infer/signs-syn.opt:6:10: error: expected string not found in input
    ; CHECK: %9:i32 = xor %0, %2
             ^
    <stdin>:1:1: note: scanning from here
    ; Failed to infer RHS
    ^
    <stdin>:1:6: note: possible intended match here
    ; Failed to infer RHS
         ^

I am using z3 4.3.1.

rsas:
This test case is problematic, I removed it in my last squashed commit.

chenyang78:
I see. Thanks!

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1376 @@
...
+#include "llvm/Support/raw_ostream.h"
+#include "souper/Extractor/KLEEBuilder.h"
+#include "souper/Infer/InstSynthesis.h"
+
+using namespace souper;
+using namespace llvm;
+
+namespace {
+
+static cl::opt<bool> DebugSynthesis("souper-synthesis-debug",
```
regehr: I think it would make sense to collapse these into a single variable where level 0 means debugging is turned off, etc.

rsas: Yes, totally.

lib/Extractor/Candidates.cpp
regehr: I think it's an OK simplifying assumption for now.

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1373 @@
...
+  case Inst::BSwap: {
+    if (Ops[0] == IC.getConst(APInt(Width, 0)))
+      return IC.getConst(APInt(Width, 0));
+    break;
+  }
+
+  case Inst::Cttz:
+  case Inst::Ctlz: {
+    if (Ops[0] == IC.getConst(APInt(Width, 0)))
+      return IC.getConst(APInt(Width, Width));
```
regehr: It's a nit, but do we want to use curlies around a case that doesn't declare any local variables?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1373 @@
...
+void InstSynthesis::getInputVars(Inst *I, std::vector<Inst *> &InputVars) {
+  if (I->K == Inst::Var)
+    InputVars.push_back(I);
+  for (auto Iz : I->orderedOps())
+    getInputVars(Iz, InputVars);
+}
+
+std::string InstSynthesis::getLocVarStr(const LocVar &Loc,
+                                        const std::string Prefix) {
+  std::string Post = "";
```
regehr: a std::string is implicitly initialized to "" I believe

lib/Infer/InstSynthesis.cpp
rsas: No.

lib/Infer/InstSynthesis.cpp
rsas: You are right.

include/souper/Infer/InstSynthesis.h
```diff
@@ -0,0 +1,300 @@
...
+  std::vector<LocInst> L;
+  /// Number of input variables
+  unsigned N;
+  /// Number of components + N
+  unsigned M;
+  /// A mapping from a location variable to a concrete component instance,
+  /// namely created instruction
+  std::map<LocVar, Inst *> CompInstMap;
+  /// Location variable's width is fixed
+  const unsigned LocInstWidth = 32;
```
regehr: Is there any reason to use so many bits here? Might it be helpful to reduce this to 6 or 8?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1360 @@
...
+  // explicitly because constants are not constrainted by the inputs. Still, it
+  // can take many iterations to converge. Therefore, we limit the number
+  // of attempts per such a candidate wiring via cmd flag MaxWiringAttempts
+  // (default value: 10)
+  std::map<ProgramWiring, unsigned> NotWorkingConstWirings;
+
+  // Ask the solver for four initial concrete inputs.
+  // This number was derived experimentally giving a good overall speed-up
+  // for both small and big synthesis queries
+  unsigned InitialInputNum = 4;
```
regehr: Is there a reason for not declaring this variable in the for-loop header below?

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    MaxCompNum = Comps.size();
+
+  // Iterative synthesis loop with increasing number of components
+  for (int J = 0; J <= MaxCompNum; ++J) {
+    Inst *CompConstraint;
+    if (DebugLevel > 1)
+      llvm::outs() << "synthesizing using " << J << " component(s)\n";
+    // If synthesis using 0 components failed (aka nop synthesis),
+    // don't subsequently wire the output to the input(s)
+    if (J == 0)
```
regehr: please don't use curly braces on the else-branch but not on the if-branch

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1360 @@
...
+}
+
+void InstSynthesis::filterFixedWidthIntrinsicComps() {
+  // CtPop/BSwap/Cttz/Ctlz require specific widths
+  for (auto It = Comps.begin(); It != Comps.end();) {
+    assert(It->Width && "comp width not set");
+    if (It->Kind == Inst::BSwap && It->Width != 16 && It->Width != 32 &&
+        It->Width != 64)
+      It = Comps.erase(It);
+    else if (((It->Kind == Inst::CtPop) || (It->Kind == Inst::Ctlz) ||
```
regehr: personally I'd put more curly braces in a collection of ifs/elses that is this complicated

lib/Infer/InstSynthesis.cpp
```diff
@@ -0,0 +1,1360 @@
...
+    }
+  }
+
+  return Ret;
+}
+
+
+void InstSynthesis::addConstraints(std::vector<InstMapping> &PCs,
+                                   InstContext &IC) {
+  Inst *C = getConsistencyConstraint(IC);
```
regehr: repeated use of C for different purposes is a bit odd here-- would the code look better if you just inlined the calls that are generating the constraints inside the emplace_back calls?

regehr:
other than issues just noted, LGTM -- please look over them and then squash and merge

include/souper/Infer/InstSynthesis.h
rsas: No. I think that 6 (max 64 comps) should bu sufficient for now.

lib/Infer/InstSynthesis.cpp
rsas: No.

lib/Infer/InstSynthesis.cpp
rsas: Absolutely.

rsas:
@regehr thanks for the review! I addressed your comments in c7e4777. In addition, I removed ``addConstraints`` (I think it's better to have it in the main function) and introduced ``TrueConst`` in the main function to avoid redundant calls to ``IC.getConst(APInt(1, true))``. All test cases in Debug/Release modes with both KLEE versions are passing. Please have a short look at the last commit and then I'll merge it.

regehr:
Looks good, please squash and merge.

PR #176, opened 2015-02-25
Don't turn-off UB constraints on the RHS

rsas:
The pull request #165 was incomplete, it disabled UB on the RHS
in the -souper-exploit-ub=false mode, resulting in unsound reasoning

### DIFF
--- a/lib/Extractor/KLEEBuilder.cpp
+++ b/lib/Extractor/KLEEBuilder.cpp
@@ -918,14 +918,14 @@ CandidateExpr souper::GetCandidateExprForReplacement(
   }
   // Get UB constraints of LHS and (B)PCs
   ref<Expr> LHSPCsUB = klee::ConstantExpr::create(1, Expr::Bool);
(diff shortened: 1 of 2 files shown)


rsas:
@regehr could you please look at it?

rsas:
@regehr I just found this, can't remember why we didn't merge.

regehr:
we still want to merge this one, right? I can update it...

rsas:
can't recall why we didn't merge, there was certainly a reason for this. In general, we have a problem that we cannot treat LHS and RHS separately as long as they share the same vars in common. Do we have a problem with this not being merged?

regehr:
well, the problem only occurs if we turn off UB exploitation, but we're currently wrong about UB anyway when there are instructions on the RHS.

I want to fix the UB model once and for all but I don't want to do it until we do the Z3 changes.

so maybe let's close this for now since it doesn't obviously make things more correct?

rsas:
Agreed.

PR #178, opened 2015-02-25
Increase LocInstWidth to 8 and add a assertion

rsas:
The current ``LocInstWidth`` of 6 can be easily exceeded when using the whole component library with multiple width instances.


regehr:
LGTM

PR #179, opened 2015-02-26
Support for overflow instructions

jubitaneja:
This patch teaches souper about signed/unsigned overflow of add, sub
and mul. Extractvalue instruction can only be harvested if the value
being extracted is coming from an overflow instruction.

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -386,6 +386,29 @@ Inst *ExprBuilder::build(Value *V) {
       }
       return IC.getPhi(BI.B, Incomings);
     }
+  } else if (auto EV = dyn_cast<ExtractValueInst>(V)) {
+    Inst *L = get(EV->getOperand(0));
+    ArrayRef<unsigned> Idx = EV->getIndices();
+    // NOTE: extractvalue instruction can take set of indices. Most of the
+    // times we just pass one index value, i.e. why I am extracting 0th index
+    // always, Idx[0]. If required, we can fix it by iterating over set of
+    // indices. Additonally, we harvest extractvalue instruction only if the
+    // extracted value comes from an overflow instruction. Otherwise, we simply
+    // create a var.
+    Inst *R = IC.getConst(APInt(32, Idx[0]));
+    switch (L->K) {
+      default:
+        return makeArrayRead(V);
+      case Inst::SAddWithOverflow:
+      case Inst::UAddWithOverflow:
+      case Inst::SSubWithOverflow:
+      case Inst::USubWithOverflow:
+      case Inst::SMulWithOverflow:
+      case Inst::UMulWithOverflow: {
+        unsigned WidthExtracted = L->Ops[Idx[0]]->Width;
+        return IC.getInst(Inst::ExtractValue, WidthExtracted, {L, R});
+      }
+    }
   } else if (auto Call = dyn_cast<CallInst>(V)) {
     if (auto II = dyn_cast<IntrinsicInst>(Call)) {
       Inst *L = get(II->getOperand(0));
@@ -400,6 +423,42 @@ Inst *ExprBuilder::build(Value *V) {
           return IC.getInst(Inst::Cttz, L->Width, {L});
         case Intrinsic::ctlz:
           return IC.getInst(Inst::Ctlz, L->Width, {L});
+        case Intrinsic::sadd_with_overflow: {
+          Inst *R = get(II->getOperand(1));
+          Inst *Add = IC.getInst(Inst::Add, L->Width, {L, R});
+          Inst *Overflow = IC.getInst(Inst::SAddO, 1, {L, R});
+          return IC.getInst(Inst::SAddWithOverflow, L->Width+1, {Add, Overflow});
+        }
+        case Intrinsic::uadd_with_overflow: {
+          Inst *R = get(II->getOperand(1));
+          Inst *Add = IC.getInst(Inst::Add, L->Width, {L, R});
+          Inst *Overflow = IC.getInst(Inst::UAddO, 1, {L, R});
+          return IC.getInst(Inst::UAddWithOverflow, L->Width+1, {Add, Overflow});
+        }
+        case Intrinsic::ssub_with_overflow: {
+          Inst *R = get(II->getOperand(1));
+          Inst *Sub = IC.getInst(Inst::Sub, L->Width, {L, R});
+          Inst *Overflow = IC.getInst(Inst::SSubO, 1, {L, R});
+          return IC.getInst(Inst::SSubWithOverflow, L->Width+1, {Sub, Overflow});
+        }
+        case Intrinsic::usub_with_overflow: {
+          Inst *R = get(II->getOperand(1));
+          Inst *Sub = IC.getInst(Inst::Sub, L->Width, {L, R});
+          Inst *Overflow = IC.getInst(Inst::USubO, 1, {L, R});
+          return IC.getInst(Inst::USubWithOverflow, L->Width+1, {Sub, Overflow});
+        }
+        case Intrinsic::smul_with_overflow: {
+          Inst *R = get(II->getOperand(1));
+          Inst *Mul = IC.getInst(Inst::Mul, L->Width, {L, R});
+          Inst *Overflow = IC.getInst(Inst::SMulO, 1, {L, R});
+          return IC.getInst(Inst::SMulWithOverflow, L->Width+1, {Mul, Overflow});
+        }
+        case Intrinsic::umul_with_overflow: {
+          Inst *R = get(II->getOperand(1));
+          Inst *Mul = IC.getInst(Inst::Mul, L->Width, {L, R});
+          Inst *Overflow = IC.getInst(Inst::UMulO, 1, {L, R});
+          return IC.getInst(Inst::UMulWithOverflow, L->Width+1, {Mul, Overflow});
+        }
       }
     }
   }
--- a/lib/Extractor/KLEEBuilder.cpp
+++ b/lib/Extractor/KLEEBuilder.cpp
@@ -188,11 +188,16 @@ ref<Expr> ExprBuilder::subnuwUB(Inst *I) {
 
 ref<Expr> ExprBuilder::mulnswUB(Inst *I) {
    const std::vector<Inst *> &Ops = I->orderedOps();
-   ref<Expr> L = SExtExpr::create(get(Ops[0]), 2*I->Width);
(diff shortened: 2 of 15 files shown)


lib/Extractor/Candidates.cpp
```diff
@@ -386,6 +386,29 @@ Inst *ExprBuilder::build(Value *V) {
       }
       return IC.getPhi(BI.B, Incomings);
     }
+  } else if (auto EV = dyn_cast<ExtractValueInst>(V)) {
+    Inst *L = get(EV->getOperand(0));
+    ArrayRef<unsigned> Idx = EV->getIndices();
+    // NOTE : extractvalue instruction can take set of indices. Most of the
```
rsas: Nit: remove space before colon.

lib/Extractor/Candidates.cpp
```diff
@@ -386,6 +386,29 @@ Inst *ExprBuilder::build(Value *V) {
...
+    // times we just pass one index value, i.e. why I am extracting 0th index
+    // always, Idx[0]. If required, we can fix it by iterating over set of
+    // indices. Additonally, we harvest extractvalue instruction only if the
+    // extracted value comes from an overflow instruction. Otherwise, we simply
```
rsas: What happens if you harvest an ``extractvalue`` instruction that has multiple indices? Does it change the semantics of the instruction when you extract only the index 0?

jubitaneja: Usually, the extract value instruction is used with one index value only.
But, I have marked it as a NOTE just for the case that in future if we
needed to handle set of indices too, we can extend the implementation.
Multiple indices won't change the semantics as such.

On Thu, Feb 26, 2015 at 10:29 AM, Raimondas Sasnauskas <
notifications@github.com> wrote:

> In lib/Extractor/Candidates.cpp
> <https://github.com/google/souper/pull/179#discussion_r25446145>:
>
> > @@ -386,6 +386,29 @@ Inst *ExprBuilder::build(Value *V) {
> >        }
> >        return IC.getPhi(BI.B, Incomings);
> >      }
> > +  } else if (auto EV = dyn_cast<ExtractValueInst>(V)) {
> > +    Inst *L = get(EV->getOperand(0));
> > +    ArrayRef<unsigned> Idx = EV->getIndices();
> > +    // NOTE : extractvalue instruction can take set of indices. Most of the
> > +    // times we just pass one index value, i.e. why I am extracting 0th index
> > +    // always, Idx[0]. If required, we can fix it by iterating over set of
> > +    // indices. Additonally, we harvest extractvalue instruction only if the
> > +    // extracted value comes from an overflow instruction. Otherwise, we simply
>
> What happens if you harvest an extractvalue instruction that has multiple
> indices? Does it change the semantics of the instruction when you extract
> only the index 0?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/souper/pull/179/files#r25446145>.
>



-- 
Thanks and Regards,
Jubi Taneja

lib/Parser/Parser.cpp
```diff
@@ -508,8 +526,25 @@ bool Parser::typeCheckInst(Inst::Kind IK, unsigned &Width,
     return false;
   }
 
-  if (!typeCheckOpsMatchingWidths(OpsMatchingWidths, ErrStr))
-    return false;
+  // Note : We don't 'typeCheckOpsMatchingWidths' because the
+  // overflow instructions operands and the inst width is different
+  // The overflow instruction is a tuple of two elements (i32, i1)
+  // that makes it width (32 + 1 = 33). Likewise, for 64-bit operands,
+  // it will be (64 + 1 = 65).
```
rsas: Nit: redundant space before colon, missing dot after "is different", "that makes it width" could be rephrased.

lib/Parser/Parser.cpp
```diff
@@ -532,6 +567,33 @@ bool Parser::typeCheckInst(Inst::Kind IK, unsigned &Width,
...
+    ExpectedWidth = Ops[0]->Width + 1;
+    break;
+
+  case Inst::ExtractValue:
+    if (Ops[1]->Val.getZExtValue() == 0)
+      ExpectedWidth = Ops[0]->Ops[0]->Width;
+    else if (Ops[1]->Val.getZExtValue() == 1)
+      ExpectedWidth = 1;
+    else
+      ErrStr = "extractval inst doesn't expect index value other than 0 or 1";
```
rsas: extractvalue

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -188,11 +188,12 @@ ref<Expr> ExprBuilder::subnuwUB(Inst *I) {
 
 ref<Expr> ExprBuilder::mulnswUB(Inst *I) {
    const std::vector<Inst *> &Ops = I->orderedOps();
-   ref<Expr> L = SExtExpr::create(get(Ops[0]), 2*I->Width);
-   ref<Expr> R = SExtExpr::create(get(Ops[1]), 2*I->Width);
+   Expr::Width Width = get(Ops[0])->getWidth();
```
rsas: Is there any reason why you are building an expression first to get the width instead of writing
```unsigned Width = I->Width;```
?

rsas:
@jubitaneja @regehr other than a couple of typos and nits, the patch LGTM.

lib/Extractor/KLEEBuilder.cpp
jubitaneja: I found I->Width was not giving the correct values in one of the test
cases. I need to find the test case though. Then, I followed the
traditional approach of constructing the expression and find width from it
using getWidth(), it always gave correct values.

On Thu, Feb 26, 2015 at 10:49 AM, Raimondas Sasnauskas <
notifications@github.com> wrote:

> In lib/Extractor/KLEEBuilder.cpp
> <https://github.com/google/souper/pull/179#discussion_r25447766>:
>
> > @@ -188,11 +188,12 @@ ref<Expr> ExprBuilder::subnuwUB(Inst *I) {
> >
> >  ref<Expr> ExprBuilder::mulnswUB(Inst *I) {
> >     const std::vector<Inst *> &Ops = I->orderedOps();
> > -   ref<Expr> L = SExtExpr::create(get(Ops[0]), 2*I->Width);
> > -   ref<Expr> R = SExtExpr::create(get(Ops[1]), 2*I->Width);
> > +   Expr::Width Width = get(Ops[0])->getWidth();
>
> Is there any reason why you are building an expression first to get the
> width instead of writing
> unsigned Width = I->Width;
> ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/souper/pull/179/files#r25447766>.
>



-- 
Thanks and Regards,
Jubi Taneja

rsas: Can you please point me to that test case? The result of a ``mul`` inst is always the same width as the operands.

jubitaneja: Yes, the test case is test/Solver/overflow3.ll which is written to test
signed multiplication with overflow. In the implementation of SMULO
instruction which basically gives 1-bit overflow result, indicating whether
it overflowed or not.
  case Inst::SMulO:
    return XorExpr::create(mulnswUB(I), klee::ConstantExpr::create(1,
Expr::Bool));

I debugged and found I->Width gives me value 1 and expression width using
getWidth() gives me correct result of 32.


On Thu, Feb 26, 2015 at 11:01 AM, Raimondas Sasnauskas <
notifications@github.com> wrote:

> In lib/Extractor/KLEEBuilder.cpp
> <https://github.com/google/souper/pull/179#discussion_r25448830>:
>
> > @@ -188,11 +188,12 @@ ref<Expr> ExprBuilder::subnuwUB(Inst *I) {
> >
> >  ref<Expr> ExprBuilder::mulnswUB(Inst *I) {
> >     const std::vector<Inst *> &Ops = I->orderedOps();
> > -   ref<Expr> L = SExtExpr::create(get(Ops[0]), 2*I->Width);
> > -   ref<Expr> R = SExtExpr::create(get(Ops[1]), 2*I->Width);
> > +   Expr::Width Width = get(Ops[0])->getWidth();
>
> Can you please point me to that test case? The result of a mul inst is
> always the same width as the operands.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/souper/pull/179/files#r25448830>.
>



-- 
Thanks and Regards,
Jubi Taneja

jubitaneja: The reason is that when we pass the instruction : SMulO to mulnswUB(I)
function, the width of the instruction is 1-bit. But, we are interested in
the operands width here and on the basis of operands sign-extension, we are
computing the overflow result. So, that's why we will not use I->Width.
Instead, we will use get(Ops[0])->getWidth().

On Thu, Feb 26, 2015 at 11:21 AM, Jubi Taneja <jubitaneja@gmail.com> wrote:

> Yes, the test case is test/Solver/overflow3.ll which is written to test
> signed multiplication with overflow. In the implementation of SMULO
> instruction which basically gives 1-bit overflow result, indicating whether
> it overflowed or not.
>   case Inst::SMulO:
>     return XorExpr::create(mulnswUB(I), klee::ConstantExpr::create(1,
> Expr::Bool));
>
> I debugged and found I->Width gives me value 1 and expression width using
> getWidth() gives me correct result of 32.
>
>
> On Thu, Feb 26, 2015 at 11:01 AM, Raimondas Sasnauskas <
> notifications@github.com> wrote:
>
>> In lib/Extractor/KLEEBuilder.cpp
>> <https://github.com/google/souper/pull/179#discussion_r25448830>:
>>
>> > @@ -188,11 +188,12 @@ ref<Expr> ExprBuilder::subnuwUB(Inst *I) {
>> >
>> >  ref<Expr> ExprBuilder::mulnswUB(Inst *I) {
>> >     const std::vector<Inst *> &Ops = I->orderedOps();
>> > -   ref<Expr> L = SExtExpr::create(get(Ops[0]), 2*I->Width);
>> > -   ref<Expr> R = SExtExpr::create(get(Ops[1]), 2*I->Width);
>> > +   Expr::Width Width = get(Ops[0])->getWidth();
>>
>> Can you please point me to that test case? The result of a mul inst is
>> always the same width as the operands.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/google/souper/pull/179/files#r25448830>.
>>
>
>
>
> --
> Thanks and Regards,
> Jubi Taneja
>
>


-- 
Thanks and Regards,
Jubi Taneja

rsas: Ok, now it totally makes sense. Thanks for the explanation. A small comment above that line would be helpful though.

rsas:
Thanks @jubitaneja, LGTM

PR #185, opened 2015-06-12
Handle llvm.assume (B)

dv-extrarius:
I think this is a proper handling of llvm.assume. Ignoring any replacement that seeks to prove something already known doesn't simplify assumes in my tests. I originally had the removal of the replacement in Pass.cpp after the solver was invoked, so the replacement could be compared against the path constraint's value. However, if that test failed for a PC with a const RHS, it would prove an internal inconsistency, and thus the earlier check should speed up souper while retaining useful candidates.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ execute_process(
 set(LLVM_CXXFLAGS "${LLVM_CXXFLAGS} -fno-exceptions -fno-rtti -DLLVM_ENABLE_STATS=true")
 
 execute_process(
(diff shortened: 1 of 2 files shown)


regehr:
Hi! Could you please add some test cases? Also the CI bot is unhappy about something.

pcc:
Also please remember to sign the CLA. https://cla.developers.google.com/clas

jeroenk:
I cannot easily test this right now, but do the two test cases I wrote work as expected?

It's not clear to me how this patch solves the problem I was having, where the assume was actually originating from a position in the code after the instruction we're trying to simplify.

dv-extrarius:
I made the rookie mistake of comparing my work-in-progress to a modified "base". This patch does prevent assume instructions from being replaced, but also prevents many other candidates from being recognized. It does bring up a simple optimization though, in that if the candidate is proving something we already know, the RHS can be populated from the PC instead of invoking the solver.

PR #186, opened 2015-06-15
Constrain not-working candidates explicitly

rsas:
With each new counter-example, the synthesis algorithm should never
generate the same component wiring. However, this old invalid wiring
might still be embedded in the new synthesis result in addition to new
valid wirings. From the solver point of view, this is a valid behaviour.
However, when we are reconstructing a candidate from the resulting set
of wirings, we don't know if it's valid or not. Therefore, it makes sense
to explicitly tell the synthesis engine to not produce the invalid wiring again
so we don't have to worry about the validity of wirings during reconstruction
of candidates.


regehr:
LGTM

PR #187, opened 2015-06-27
Forbid candidates with no benefit

rsas:
The synthesis loop assumes that each component has a cost of one. This is not the case for all components in the library. Thus, for each new candidate we first check if it has some benefit. If not, we forbid it by constraining the wiring and continue searching for others.


regehr:
LGTM

PR #194, opened 2015-07-08
LLVM's local identifier can't be just a dot

jubitaneja:
Don't allow '%.' as a LLVM local identifier


pcc:
This looks like it would forbid periods in identifiers altogether. I'm pretty sure periods are valid characters in LLVM identifiers (even on their own), so it probably makes sense to allow them in souper IR as well.

jubitaneja:
Correct. I misunderstood the regular expression in LLVM's manual. It's '[%@][-a-zA-Z$._][-a-zA-Z$._0-9]*'. I am just editing it and adding '$' and '_' too.

PR #203, opened 2015-10-01
support for demanded bits analysis

jubitaneja:


include/souper/Inst/Inst.h
```diff
@@ -119,6 +120,7 @@ struct Inst : llvm::FoldingSetNode {
   static int getCost(Kind K);
   llvm::APInt KnownZeros;
   llvm::APInt KnownOnes;
+  llvm::APInt DemandedBitsVal = llvm::APInt(1, 0, false);
```
regehr: Why are you creating an APInt of width 1 here? This doesn't seem useful.

lib/Extractor/Candidates.cpp
```diff
@@ -684,17 +695,19 @@ std::tuple<BlockPCs, std::vector<InstMapping>> GetRelevantPCs(
...
                            ExprBuilderContext &EBC,
                            FunctionCandidateSet &Result) {
-  ExprBuilder EB(Opts, F.getParent(), LI, IC, EBC);
+  ExprBuilder EB(Opts, F.getParent(), LI, DB, IC, EBC);
 
   for (auto &BB : F) {
     std::unique_ptr<BlockCandidateSet> BCS(new BlockCandidateSet);
     for (auto &I : BB) {
       if (I.getType()->isIntegerTy())
+      {
```
regehr: don't just add braces for no reason

test/Infer/demanded-bits1.opt
```diff
@@ -0,0 +1,9 @@
+; REQUIRES: solver, solver-model
+
+; RUN: %souper-check %solver -infer-rhs -souper-infer-inst -souper-synthesis-comps=const %s > %t1
+; RUN: FileCheck %s < %t1
+
+; CHECK: result 240:i8
+
+%0:i8 = shl 254:i8, 3:i8
+infer %0 (ddddnnnn)
```
regehr: I think this test case is bogus. If you evaluate 254<<3 you're going to get 240 regardless of the demanded bits.

Please write thorough test cases for your new feature. Make sure that it works.

unittests/Parser/ParserTests.cpp
```diff
@@ -79,13 +79,13 @@ TEST(ParserTest, Errors) {
...
       { "%0:i4 = var (2345)\n",
-        "<input>:1:14: invalid knownbits string" },
+        "<input>:1:14: invalid bits string, expected [0|1]+ or [n|d]+" },
       { "%0:i4 = var (01xa)\n",
         "<input>:1:17: invalid knownbits string" },
       { "%0:i4 = var (xxx)\n",
         "<input>:1:1: knownbits pattern must be of same length as var width" },
       { "%0:i4 = var ()\n",
-        "<input>:1:14: invalid knownbits string" },
+        "<input>:1:14: invalid bits string, expected [0|1]+ or [n|d]+" },
```
regehr: you should mention "x"

regehr:
In addition to the missing tests for functionality, you haven't added test cases showing that the parse/print loop works. There are lots of examples already there for you to work from.

lib/Extractor/Candidates.cpp
jubitaneja: Yes, that was by mistake. I added it for some debugging purpose earlier.

include/souper/Inst/Inst.h
jubitaneja: I am just initializing the demanded bits value to 0, otherwise it will take some garbage values from the places where I am not setting its value. Width of 1 is just for initialization purpose and it gets over-written with exact width and value wherever we get some concrete values to assign it to. Similarly, I did for known zeros/ones, the only difference is I was passing them as an argument to createVar() function and thus initialized the function arguments value to llvm::APInt(1, 0, false)

PR #205, opened 2015-11-03
teach souper about some more data flow analysis like nonZero, nonNega…

jubitaneja:
…tive and powerOfTwo


lib/Extractor/KLEEBuilder.cpp
```diff
@@ -143,14 +149,26 @@ ref<Expr> ExprBuilder::makeSizedArrayRead(unsigned Width, StringRef Name,
   std::vector<unsigned> ZeroBits, OneBits;
   UpdateList UL(Arrays.back().get(), 0);
   ref<Expr> Var = ReadExpr::create(UL, klee::ConstantExpr::alloc(0, Expr::Int32));
-  if (Origin && Origin->K == Inst::Var && (Origin->KnownZeros.getBoolValue() ||
-                                           Origin->KnownOnes.getBoolValue())) {
```
regehr: this'll look better if you put the whole or-expression on the second line

lib/Extractor/Candidates.cpp
```diff
@@ -45,6 +45,10 @@ static llvm::cl::opt<bool> HarvestKnownBits(
     "souper-harvest-known-bits",
     llvm::cl::desc("Perform known bits analysis (default=true)"),
     llvm::cl::init(true));
+static llvm::cl::opt<bool> HarvestMoreKnownBits(
+    "souper-harvest-more-known-bits",
+    llvm::cl::desc("Perform more known bits analysis to find if non-zero(default=false)"),
+    llvm::cl::init(false));
```
regehr: Please just put all known bits until control of a single flag

lib/Extractor/KLEEBuilder.cpp
jubitaneja: I think you meant the OR expression to look like as its in line : 153 below. The line is deleted for which you pointed a change. (?)

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -143,14 +149,26 @@ ref<Expr> ExprBuilder::makeSizedArrayRead(unsigned Width, StringRef Name,
...
-    ref<Expr> NotZeros = NotExpr::create(klee::ConstantExpr::alloc(Origin->KnownZeros));
-    ref<Expr> VarOrNotZero = OrExpr::create(Var, NotZeros);
-    ZeroBitsMap[Origin] = EqExpr::create(VarOrNotZero, NotZeros);
-    ref<Expr> Ones = klee::ConstantExpr::alloc(Origin->KnownOnes);
-    ref<Expr> VarAndOnes = AndExpr::create(Var, Ones);
-    OneBitsMap[Origin] = EqExpr::create(VarAndOnes, Ones);
+  if (Origin && Origin->K == Inst::Var) {
+    if (Origin->KnownZeros.getBoolValue() || Origin->KnownOnes.getBoolValue()) {
```
jubitaneja: Does this look ok?

regehr:
Here's a case where we need a better error message, or better yet just support the flags in either order.

```
pc33:souper-jubi regehr$ cat more-knownbits1.opt
%0:i32 = var
%1:i32 = var (z) (00000000000000000000000000000000)
%2:i32 = shlnw %0, %1
%3:i32 = and %2, 2147483648:i32
%4:i1 = eq %3, 0:i32
infer %4
pc33:souper-jubi regehr$ ./build/souper-check $SOUPER_SOLVER -infer-rhs -print-replacement < more-knownbits1.opt 
<stdin>:2:19: expected inst, block, cand, infer, result, pc, or blockpc
pc33:souper-jubi regehr$ 
```

test/Tool/more-knownbits11.opt
```diff
@@ -0,0 +1,12 @@
...
+; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
+; RUN: FileCheck %s < %t
+
+; CHECK: LGTM
+%0:i8 = var (p)
+%1:i8 = var (z)
+%2:i8 = udivexact %0, %1
+%3:i8 = ctpop %2
+%4:i1 = eq %3, 1:i8
+cand %4 1:i1
```
regehr: this one might not be testing what you think it is testing, it is still LGTM when the (z) is removed

test/Tool/more-knownbits5.opt
```diff
@@ -0,0 +1,18 @@
...
+
+; CHECK: %0:i16 = var (p)
+; CHECK-NEXT: %1:i16 = var (z)
+; CHECK-NEXT: %2:i16 = udivexact %0, %1
+; CHECK-NEXT: %3:i16 = ctpop %2
+; CHECK-NEXT: %4:i1 = eq 1:i16, %3
+; CHECK-NEXT: cand %4 1:i1
+
+%0:i16 = var (p)
+%1:i16 = var (z)
```
regehr: this one works even without the (z)

test/Tool/more-knownbits11.opt
jubitaneja: I added the 'z' flag just to make sure that denominator of division operation is never zero. The intention was basically to test power of two feature only. I have fixed the test case by removing 'z' flag. Similarly in other test case too.

regehr:
I'm getting an apparent infinite loop in parser_tests. It just hangs up, both in a Debug and Release build.

PR #207, opened 2015-11-22
add a new option for dynamic profiling of all sites that souper might

regehr:
optimize, not just the ones that it has been able to optimize. this is
useful because it permits us to find the highest profile sites and
then run expensive synthesis only on those, instead of doing synthesis
first.

this profile-all mode needs to run last: it can't run as a peephole
pass.  unfortunately it isn't possible to make a single pass that is
scheduled either as a peephole or running last, so this patch compiles
two versions of the pass and then sclang chooses the right one

also, it turns out that running profile code from an LLVM function
pass was incorrect and only happened to work: it is illegal to add new
functions from a function pass. so turn the function pass into a
module pass.

finally, fix a different bug in the profile instrumentation where we
had failed to put the atomic increment below any phi nodes that happen
to be there.

### DIFF
--- a/lib/Pass/Pass.cpp
+++ b/lib/Pass/Pass.cpp
@@ -19,6 +19,7 @@
 #include "llvm/IR/Module.h"
 #include "llvm/Pass.h"
 #include "llvm/IR/LegacyPassManager.h"
+#include "llvm/IR/Module.h"
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Transforms/IPO/PassManagerBuilder.h"
@@ -28,6 +29,7 @@
 #include "souper/SMTLIB2/Solver.h"
 #include "souper/Tool/GetSolverFromArgs.h"
 #include "souper/Tool/CandidateMapUtils.h"
+#include "set"
 
 using namespace souper;
 using namespace llvm;
@@ -68,11 +70,17 @@ static bool dumpAllReplacements() {
   return DebugSouperPass && (DebugLevel > 1);
 }
 
-struct SouperPass : public FunctionPass {
+#ifdef DYNAMIC_PROFILE_ALL
+static const bool DynamicProfileAll = true;
+#else
+static const bool DynamicProfileAll = false;
+#endif
+
+struct SouperPass : public ModulePass {
   static char ID;
 
 public:
-  SouperPass() : FunctionPass(ID) {
+  SouperPass() : ModulePass(ID) {
     if (!S) {
       S = GetSolverFromArgs(KV);
       if (StaticProfile && !KV)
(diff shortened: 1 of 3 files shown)


lib/Pass/Pass.cpp
```diff
@@ -133,27 +150,32 @@ struct SouperPass : public FunctionPass {
...
     InstContext IC;
     ExprBuilderContext EBC;
     CandidateMap CandMap;
-
-    LoopInfo *LI = &getAnalysis<LoopInfoWrapperPass>().getLoopInfo();
-
-    FunctionCandidateSet CS = ExtractCandidatesFromPass(&F, LI, IC, EBC);
+    LoopInfo *LI = &getAnalysis<LoopInfoWrapperPass>(*F).getLoopInfo();
+    if (!LI)
+      report_fatal_error("foo!");
```
rsas: Is there a better error message?

lib/Pass/Pass.cpp
```diff
@@ -133,27 +150,32 @@ struct SouperPass : public FunctionPass {
...
     new AtomicRMWInst(AtomicRMWInst::Add, CntVar,
                       ConstantInt::get(C, APInt(64, 1)), Monotonic, CrossThread,
                       BI);
   }
 
-  bool runOnFunction(Function &F) {
+  std::set <Function *> DoneFuncs;
+
+  bool runOnFunction(Function *F) {
     bool changed = false;
```
rsas: Can't remember if there was a naming policy for local vars, shouldn't it be ``Changed``?

lib/Pass/Pass.cpp
```diff
@@ -250,6 +269,20 @@ struct SouperPass : public FunctionPass {
 
     return changed;
   }
+
+  bool runOnModule(Module &M) {
+    bool changed = false;
```
rsas: Ditto.

rsas:
LGTM, but couldn't test it yet.

lib/Pass/Pass.cpp
regehr: Ouch how did that slip through??

regehr:
Issues fixed-- ok to merge?

lib/Pass/Pass.cpp
```diff
@@ -133,27 +150,32 @@ struct SouperPass : public FunctionPass {
...
     std::string FunctionName;
-    if (F.hasLocalLinkage()) {
+    if (F->hasLocalLinkage()) {
       FunctionName =
-        (F.getParent()->getModuleIdentifier() + ":" + F.getName()).str();
+        (F->getParent()->getModuleIdentifier() + ":" + F->getName()).str();
     } else {
-      FunctionName = F.getName();
+      FunctionName = F->getName();
     }
```
rsas: Are we still following the code style policy "no curly braces for single lines"? If yes, please edit both ``if`` and ``else`` path.

rsas:
Except of the code style question, LGTM.

lib/Pass/Pass.cpp
regehr: Overall I don't think we're very consistent about this choice. We can address this later with a cleanup patch (though I don't really have strong feelings about doing this) but I think for now this is fine. Thanks for the careful read!

PR #211, opened 2015-12-17
help message for sclang

regehr:


rsas:
We don't want instruction synthesis help here, right? Other than that, LGTM.

regehr:
I'll merge this and you can add that somewhere else when you get time?

rsas:
Too late to answer :-)

PR #213, opened 2016-01-06
bugfix and cleanup suggested by John for printing var instruction with nzp option

jubitaneja:
There is a bug in function to print the non-zero, non-negative and power of two options printing in Inst.cpp. Invalid test cases are added to expose the bug.

### DIFF
--- a/test/Tool/more-knownbits17.opt
+++ b/test/Tool/more-knownbits17.opt
@@ -1,0 +1,12 @@
+; REQUIRES: solver, solver-model
+
+; RUN: %souper-check %solver -infer-rhs -print-replacement %s > %t1 2>&1
+; RUN: FileCheck %s < %t1
+
+; CHECK: %0:i16 = var (np)
+; CHECK-NEXT: %1:i1 = slt 0:i16, %0
+; CHECK-NEXT: cand %1 1:i1
+
+%0:i16 = var (np)
+%1:i1 = slt 0:i16, %0
+infer %1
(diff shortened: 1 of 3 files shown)


regehr:
Don't just check that the result is invalid, you need a full parse/print loop. At the end you check that the parsed/printed code is the same as at the start.

test/Tool/more-knownbits17.opt
```diff
@@ -0,0 +1,12 @@
+; REQUIRES: solver, solver-model
+
+; RUN: %souper-check %solver -infer-rhs -print-replacement %s > %t1 2>&1
+; RUN: FileCheck %s < %t1
+
+; CHECK: %0:i16 = var (np)
+; HECK-NEXT: %1:i1 = slt 0:i16, %0
```
regehr: HECK-NEXT ??

jubitaneja: Extremely sorry about the worst typo I have ever made. I have fixed and
pushed the updates. I ensure to be really careful next time.
ᐧ

On Sat, Jan 16, 2016 at 4:15 PM, John Regehr <notifications@github.com>
wrote:

> In test/Tool/more-knownbits17.opt
> <https://github.com/google/souper/pull/213#discussion_r49938653>:
>
> > @@ -0,0 +1,12 @@
> > +; REQUIRES: solver, solver-model
> > +
> > +; RUN: %souper-check %solver -infer-rhs -print-replacement %s > %t1 2>&1
> > +; RUN: FileCheck %s < %t1
> > +
> > +; CHECK: %0:i16 = var (np)
> > +; HECK-NEXT: %1:i1 = slt 0:i16, %0
>
> HECK-NEXT ??
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/souper/pull/213/files#r49938653>.
>



-- 
Thanks and Regards,
Jubi Taneja

regehr:
LGTM

PR #214, opened 2016-01-23
Added support for another DFA on number of signbits

jubitaneja:


lib/Extractor/Candidates.cpp
```diff
@@ -158,6 +158,7 @@ Inst *ExprBuilder::makeArrayRead(Value *V) {
   unsigned Width = DL.getTypeSizeInBits(V->getType());
   APInt KnownZero(Width, 0, false), KnownOne(Width, 0, false);
   bool NonZero = false, NonNegative = false, PowOfTwo = false;
+  unsigned NumSignBits = 0;
```
regehr: Why initialize NumSignBits to 0 here and 1 above?

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -169,6 +172,23 @@ ref<Expr> ExprBuilder::makeSizedArrayRead(unsigned Width, StringRef Name,
                                                                                   klee::ConstantExpr::create(i, Width))));
       PowerTwoBitsMap[Origin] = PowerExpr;
     }
+    if (Origin->NumSignBits > 1) {
+      ref<Expr> MSB = ExtractExpr::create(Var, Width-1, Expr::Bool);
+      ref<Expr> SignBitsExpr = klee::ConstantExpr::create(1, Expr::Bool);
+      for (int I=Width-1; I >= (int) (Width-Origin->NumSignBits); --I)
+        SignBitsExpr = AndExpr::create(SignBitsExpr, EqExpr::create(MSB, ExtractExpr::create(Var, I, Expr::Bool)));
+      if (Origin->NumSignBits != Width) {
```
regehr: Why the special case? What do you think should happen if NumSignBits == Width?

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -169,6 +172,23 @@ ref<Expr> ExprBuilder::makeSizedArrayRead(unsigned Width, StringRef Name,
...
+         int J = (int) Width - 1- Origin->NumSignBits;
+         ref<Expr> Expr1 = ExtractExpr::create(Var, J, Expr::Bool);
+         ref<Expr> Expr2 = ExtractExpr::create(Var, J+1, Expr::Bool);
+         SignBitsExpr = AndExpr::create(SignBitsExpr, EqExpr::create(AndExpr::create(Expr1, Expr2),
+                                        klee::ConstantExpr::create(0, Expr::Bool)));
+      }
+      SignBitsExpr = AndExpr::create(SignBitsExpr,
+                                     Expr::createImplies(EqExpr::create(MSB, klee::ConstantExpr::create(0, Expr::Bool)),
+                                                         EqExpr::create(ctlz(Var), klee::ConstantExpr::create(Origin->NumSignBits, Width))));
```
regehr: This all seems a bit complicated. Wouldn't it be easier to simply assert that the N sign bits are either all 1 or all 0? Then you don't need any for loops.

lib/Extractor/Candidates.cpp
jubitaneja: Yes, here it should be initialized to 1 because by default the return value of computeNumSignBits is 1.

lib/Extractor/KLEEBuilder.cpp
jubitaneja: These extra cases to handle the number of sign bits are added to find more optimizations. The case of (NumSignBits != width) here makes sure that if we have 'N' sign bits, then (N+!)th bit from MSB will be opposite to the sign bit value.
For the cases when, (NumSignbits == Width), such cases don't need to be applied and that's why I didn't add anything.

lib/Extractor/KLEEBuilder.cpp
jubitaneja: Yes, the idea of asserting all bits are either 0 or 1 is much more efficient than the for loop. I think if we add extra cases to teach souper about number of sign bits. Like, I have added the cases to handle the variables which are both power of two and have N sign bits. Should I keep this as a simple expression or can I add more cases to make it more better in terms of finding more optimizations?

test/Tool/signbits1-invalid.opt
```diff
@@ -0,0 +1,11 @@
+; REQUIRES: solver
+
+; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
+; RUN: FileCheck %s < %t
+
+; CHECK: Invalid
+%0:i32 = var
+%1:i32 = var
```
regehr: this test would be a bit better if you defined each var as (s=1)

test/Tool/signbits10.opt
```diff
@@ -0,0 +1,11 @@
+; REQUIRES: solver
+
+; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
+; RUN: FileCheck %s < %t
+
+; CHECK: LGTM
+%0:i16 = var (p) (s=2)
```
regehr: I think I understand this one but can you please add a comment explaining it?

test/Tool/signbits11.opt
```diff
@@ -0,0 +1,11 @@
+; REQUIRES: solver
+
+; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
+; RUN: FileCheck %s < %t
+
+; CHECK: LGTM
+%0:i16 = var (p) (s=10)
```
regehr: here too

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -169,6 +172,21 @@ ref<Expr> ExprBuilder::makeSizedArrayRead(unsigned Width, StringRef Name,
...
+      ref<Expr> TestOnes = SubExpr::create(ShlExpr::create(klee::ConstantExpr::create(1, Width),
+                                           klee::ConstantExpr::create(Origin->NumSignBits, Width)),
+                                           klee::ConstantExpr::create(1, Width));
+      SignBitsExpr = AndExpr::create(SignBitsExpr,
+                                     OrExpr::create(EqExpr::create(Res, TestOnes),
+                                                    EqExpr::create(Res, klee::ConstantExpr::create(0, Width))));
+      ref<Expr> MSB = ExtractExpr::create(Var, Width-1, Expr::Bool);
+      SignBitsExpr = AndExpr::create(SignBitsExpr,
+                                     Expr::createImplies(EqExpr::create(MSB, klee::ConstantExpr::create(0, Expr::Bool)),
+                                                         EqExpr::create(ctlz(Var), klee::ConstantExpr::create(Origin->NumSignBits, Width))));
```
regehr: I'm not sure this is correct -- it looks like you're asserting the the number of sign bits is equal to the value. This is too strong. The number of sign bits is a conservative estimate: it has to be at least that much, but might be more. You can's say that it's exactly a certain value.

I think there's an easier way to do what you are trying to do here: just assert that the highest N bits of the value are either all 0 or all 1.  

Or better yet, shift the value right (arithmetically) by width-#signbits and then assert that the resulting value is either 0 or all ones.

jubitaneja: I compared for "N" higher bits only and thus logically shifted them. Arithmetic shift would lead to comparison with all bits. I have updated it to arithmetic shift though.
The second condition to build sign bits expression is to make sure that the leading number of zeros should be equal to number of sign bits for positive variables.
If I don't teach Souper about this condition, then we can't validate test cases #6, #8, #10, #11. I just tried to make souper more stronger herethan LLVM.

test/Tool/signbits1-invalid.opt
jubitaneja: Each variable by default has s=1. I only print number of sign bits if s>1

test/Tool/signbits10.opt
jubitaneja: Added.

test/Tool/signbits1-invalid.opt
regehr: but the parser accepts s=1, right? this will make it clear to people that this is the negative test case corresponding to the s=2 case, which is positive

lib/Extractor/KLEEBuilder.cpp
regehr: Your goal here is to have the same semantics as LLVM. If you implement a stronger semantics then we will derive optimizations that fail to hold in LLVM and then Souper's results are no longer useful.

You're correct that some of your test cases go away when you implement the LLVM semantics, you'll have to get rid of them

regehr: I'm pretty sure that arithmetic shift right followed by assertion that the result == 0 or result == -1 is the simplest and best implementation for this

jubitaneja: Updated. Removed the leading zeros assertions.

test/Tool/signbits1-invalid.opt
jubitaneja: I have updated the parser to accept s=1. Earlier I didn't allow it to accept s=0, 1.

lib/Extractor/Candidates.cpp
```diff
@@ -158,16 +158,18 @@ Inst *ExprBuilder::makeArrayRead(Value *V) {
   unsigned Width = DL.getTypeSizeInBits(V->getType());
   APInt KnownZero(Width, 0, false), KnownOne(Width, 0, false);
   bool NonZero = false, NonNegative = false, PowOfTwo = false;
+  unsigned NumSignBits = 1;
   if (HarvestKnownBits)
```
regehr: plz change this to HarvestDataflow or similar, since it is doing more than known bits

lib/Parser/Parser.cpp
```diff
@@ -226,15 +251,23 @@ Token Lexer::getNextToken(std::string &ErrStr) {
       ErrStr = "invalid more knownbits string";
       return Token{Token::Error, Begin, 0, APInt()};
     }
+    if (Begin != NumBegin && NumSignBitsFlag && *Begin != ')') {
+      ErrStr = "invalid number of sign bits string";
+      return Token{Token::Error, Begin, 0, APInt()};
+    }
     if (Begin == NumBegin) {
-      ErrStr = "invalid, expected [0|1|x]+ or [n|z|p]";
+      ErrStr = "invalid, expected [0|1|x]+ or [n|z|p] or [s=][0-9]+";
```
regehr: this isn't right: [s=][0-9]+

this is a regex that would match, for example, '=' or 's' or 's34'

regehr:
also you need to refresh to latest souper before this can land

test/Tool/signbits6.opt
```diff
@@ -0,0 +1,10 @@
+; REQUIRES: solver
+
+; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
+; RUN: FileCheck %s < %t
+
+; CHECK: LGTM
+%0:i32 = var (p) (s=31)
+%1:i32 = cttz %0
+%2:i1 = eq %1, 0:i32
+cand %2 1:i1
```
regehr: I don't understand this one. If there are 31 sign bits, then the lowest bit should be able to be either 0 or 1, right?

jubitaneja: In my understanding, here the lowest significant bit has to be 1. I can still confirm if the 32-bit number with value 0, has 32 sign bits or 31.

regehr: ah, you are right, I was thinking non-negative, not strictly positive, thanks

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -177,6 +179,14 @@ ref<Expr> ExprBuilder::makeSizedArrayRead(unsigned Width, StringRef Name,
                                                                                   klee::ConstantExpr::create(i, Width))));
       PowerTwoBitsMap[Origin] = PowerExpr;
     }
+    if (Origin->NumSignBits > 1) {
+      ref<Expr> Res = AShrExpr::create(Var, klee::ConstantExpr::create(Width-Origin->NumSignBits, Width));
+      ref<Expr> TestOnes = AShrExpr::create(ShlExpr::create(klee::ConstantExpr::create(1, Width),
+                                                            klee::ConstantExpr::create(Width-1, Width)),
```
regehr: put spaces around binary operators please

jubitaneja: Fixed. Thanks for pointing out this one.

PR #223, opened 2016-07-18
Add an option to exclude the Clang tool

rgov:
This adds the option `-DBUILD_CLANG_TOOL:BOOL=NO` which allows one to disable all of the Clang-related build targets.


rgov:
Signed.

rgov:
@regehr What do you think?

regehr:
LGTM, thanks.

PR #225, opened 2016-07-19
SQLite support

rgov:
**This is an incomplete implementation.**

I'd like feedback on this sketch of how SQLite support could be added before I continue with it. I wanted to make it so that Redis and SQLite are interchangeable at compile-time, and either can be selected at runtime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,6 +275,12 @@ target_link_libraries(extractor_tests souperExtractor ${GTEST_LIBS})
 target_link_libraries(inst_tests souperInst ${GTEST_LIBS})
 target_link_libraries(parser_tests souperParser ${GTEST_LIBS})
 
+SET(BUILD_CLANG_TOOL 1 CACHE BOOL "Build the Souper Clang tool")
+if (NOT BUILD_CLANG_TOOL)
+  message(STATUS "Excluding Clang tool from build")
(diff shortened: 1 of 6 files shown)


regehr:
I don't see any showstoppers here.

Let's try to keep the conditional compilation to a bare minimum.

If I understand correctly, every compiler invocation is going to have to open up a (potentially large) SQLite file. Is this going to perform acceptably?

rgov:
> Is this going to perform acceptably?

It's hard to say without implementing it. SQLite performance is generally pretty good, but the advantage Redis has is that the client doesn't need to wait for Redis to store the data. SQLite allows multiple concurrent readers, which works well if you have a high cache hit rate, but only a single writer at a time.

After running Souper across a large project, Redis grew about 140k keys, most or all of which have only one field (the result). In total that's 65 megabytes of data, which Redis appears to compress.

I made a naïve SQLite schema and dumped Redis's keys into it. The file I produced was 442% larger, without using compression or any optimizations. I haven't benchmarked the lookup speed, but this size is definitely something I'd want to bring down.

From this one example project, only 600 queries had a non-empty "result" field, accounting for only 8 KB of data. So an obvious improvement to space, albeit one that isn't solved by using SQLite, would be to store failed queries in a bloom filter. This would shrink the whole dataset down to only a few MB, although you'd lose the ability to inspect the findings.

rgov:
This is not intended to be merged in at this point, so I'm closing. Thanks for the feedback.

PR #229, opened 2016-10-08
allow kleebuilder to fail 

regehr:
instead of exploding and using all RAM + a testcase from building LLVM that explodes and uses all RAM without this patch

this code isn't too pretty, it takes a code path that used to always succeed and allows it to fail. but I believe it is the right answer here, we've already tried several mitigations for explosion in kleebuilder and this solves all future problems of that sort.


regehr:
anyone have time to look this over? I'd like to move on with some other changes as we move towards the PLDI deadline

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -843,11 +850,14 @@ void ExprBuilder::getUBPhiPaths(Inst *I, PhiPath *Current,
     Current->BlockConstraints[I->B] = 0;
     // Continue recursively
     for (unsigned J = 0; J < Ops.size(); ++J)
-      getUBPhiPaths(Ops[J], Tmp[J], Paths, CachedPhis);
+      if (!getUBPhiPaths(Ops[J], Tmp[J], Paths, CachedPhis, Depth + 1))
+	return false;
```
chenyang78: Two spaces?

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -843,11 +850,14 @@ void ExprBuilder::getUBPhiPaths(Inst *I, PhiPath *Current,
...
   } else {
     for (unsigned J = 0; J < Ops.size(); ++J)
-      getUBPhiPaths(Ops[J], Current, Paths, CachedPhis);
+      if (!getUBPhiPaths(Ops[J], Current, Paths, CachedPhis, Depth + 1))
+	return false;
```
chenyang78: Same space issue?

lib/Extractor/Solver.cpp
```diff
@@ -87,6 +89,8 @@ class BaseSolver : public Solver {
       Inst *I = IC.createVar(LHS->Width, "constant");
       InstMapping Mapping(LHS, I);
       std::string Query = BuildQuery(BPCs, PCs, Mapping, &ModelInsts, /*Negate=*/true);
+      if (Query.empty())
+	return std::make_error_code(std::errc::value_too_large);
```
chenyang78: Two spaces

lib/Extractor/Solver.cpp
```diff
@@ -149,10 +157,11 @@ class BaseSolver : public Solver {
       }
       return EC;
     } else {
+      std::string Query = BuildQuery(BPCs, PCs, Mapping, 0);
+      if (Query.empty())
+	return std::make_error_code(std::errc::value_too_large);
```
chenyang78: same issue

lib/Infer/InstSynthesis.cpp
```diff
@@ -214,6 +216,8 @@ std::error_code InstSynthesis::synthesize(SMTLIBSolver *SMTSolver,
       // Don't use original BPCs/PCs, they are useless
       std::string QueryStr = BuildQuery({}, LoopPCs, Mapping,
                                         &ModelInsts, /*Negate=*/true);
+      if (QueryStr.empty())
+	return std::make_error_code(std::errc::value_too_large);
```
chenyang78: Ditto

lib/Infer/InstSynthesis.cpp
```diff
@@ -268,6 +272,8 @@ std::error_code InstSynthesis::synthesize(SMTLIBSolver *SMTSolver,
       ModelVals.clear();
       InstMapping CandMapping(LHS, Cand);
       QueryStr = BuildQuery(BPCs, PCs, CandMapping, &ModelInsts, /*Negate=*/false);
+      if (QueryStr.empty())
+	return std::make_error_code(std::errc::value_too_large);
```
chenyang78: Same

lib/Extractor/KLEEBuilder.cpp
regehr: Thanks... . I think emacs on my laptop has something configured badly...

regehr:
Argh some tab chars sneaked in-- should be fixed now.

PR #231, opened 2016-10-28
Stop harvesting blockpc's if predecessors of phi node is another phi …

jubitaneja:
…node
To find if a predecessor of phi node is another phi node, we simply look for the single predecessor and keep harvesting blockpc's for that. Otherwise, for predecessor block which has multiple predecessors (i.e. itself a phi node) is skipped and we don't harvest blockpc's for this predecessor.

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -602,10 +602,12 @@ void ExprBuilder::addPathConditions(BlockPCs &BPCs,
 
     VisitedBlocks.insert(BI.B);
     for (unsigned i = 0; i < BI.Preds.size(); ++i) {
-      std::vector<InstMapping> PCs;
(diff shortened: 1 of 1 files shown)


regehr:
I don't totally understand what is going on in this part of Souper but it looks like this patch causes us to harvest fewer pcs as well as fewer blockpcs, is that correct?  Does that sounds right @chenyang78 ?

chenyang78:
Argh, you are right, John. The code has some issue. We should only skip adding PCs to BlockPCs, but we still need to invoke ```addPathConditions``` even if ```BI.Preds[i]``` has single predecessor.

chenyang78:
Hmm, thinking about it more. @regehr, John, do you have an example which shows the patch caused fewer ```PCs``` being harvested? Thanks.

regehr:
I don't have an example, Yang! Sorry if this was a false alarm, the structure of the patch just seemed worrisome. I'll spend a bit of time looking for an example.

Issue #55, opened 2014-09-17
select is wrong when there is undefined behavior


regehr:
I'll just paste in a mail that I sent to the llvmdev list. The solution is for Raimondas to adapt his phi+UB patch to do something analogous for select, I think.

In the section about poison values, the LLVM language reference manual says:

  "Values other than phi nodes depend on their operands."

This implies that a select instruction's output can be poisoned by its not-selected argument value. If select were poisoned only by its selected argument, we would expect this fact to be mentioned specifically, as it is for phi.

Next I'll show how poisoning by the not-selected value can introduce a problem. Take this C program:

```
int printf(const char *, ...);
int foo(int x0) {
  int x3 = x0 >> 27;
  int x4 = x3 - 27;
  int x2 = x4 ? x4 : (1 >> x4);
  int x1 = x2 != 0;
  return x1;
}
int arg;
int main (void) {
  int x1 = foo(arg);
  printf("%x\n", x1);
  return 0;
}
```

This program has no UB and it should print "1'.

Clang at -O2 turns foo() into this:

```
define i32 @foo(i32 %x0) #0 {
entry:
  %shr = ashr i32 %x0, 27
  %sub = add nsw i32 %shr, -27
  %tobool = icmp ne i32 %sub, 0
  %shr1 = lshr i32 1, %sub
  %cond = select i1 %tobool, i32 %sub, i32 %shr1
  %cmp = icmp ne i32 %cond, 0
  %conv = zext i1 %cmp to i32
  ret i32 %conv
}
```

Although the translation is very straightforward, there's a problem: %shr1 is a poison value and although it is not selected, it propagates through the rest of the function and poisons the return value.

Obviously things are not going well if the source program contains no UB and the emitted code returns a poison value. Either there has been a miscompilation or else the LLVM documentation should be changed to make it clear that the select instruction only has a value dependence on the value that is selected.

The immediate problem here is that both Souper and Alive are happy for foo() to return 0 instead of 1. Souper actually does this optimization -- legally, of course, according to the reference manual.

PR #232, opened 2016-11-01
Apply UB constraints only on the relevant Select branches

chenyang78:
This patch is based Raimondas's patch for addressing the issue #55.

https://github.com/rsas/souper/commit/66934ec740cb7f00d05362ed1b716b7330f6e136

One difference from the original patch is the part where we have to resolve
the issue caused by our cache mechanism, which was introduced after
the original patch.

### DIFF
--- a/lib/Extractor/KLEEBuilder.cpp
+++ b/lib/Extractor/KLEEBuilder.cpp
@@ -52,12 +52,13 @@ namespace {
 
 const unsigned MAX_PHI_DEPTH = 25;
 
-typedef std::unordered_map<Inst *, std::vector<ref<Expr>>> PhiMap;
+typedef std::unordered_map<Inst *, std::vector<ref<Expr>>> UBPathInstMap;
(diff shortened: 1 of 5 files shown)


chenyang78:
Not sure what's going on, but I signed CLA a long time ago...

regehr:
Ugh. Could it be the thing where you didn't set your email in git on the machine you're committing from?

chenyang78:
I use the same machine as before. The config file also looks good to me:

[user]
        name = Yang Chen
        email = chenyang@cs.utah.edu

Also, I can see my email in the commit log:

$ git log
commit ddd9d1224300aedde820158eda028f08dde3cfb8
Author: Yang Chen <chenyang@cs.utah.edu>
Date:   Sat Oct 29 23:41:42 2016 -0700

regehr:
Sounds like we can just merge-- if they want to verify they'll see that you've signed!

chenyang78:
OK.

regehr:
thanks guys!

PR #234, opened 2016-11-08
overflow instructions need to be type checked for operands width

jubitaneja:


lib/Parser/Parser.cpp
```diff
@@ -565,18 +565,14 @@ bool Parser::typeCheckInst(Inst::Kind IK, unsigned &Width,
...
-  // that makes the overall width equals to (32 + 1 = 33). Likewise, for
-  // 64-bit operands, overall width will be (64 + 1 = 65).
+  // NOTE: We don't 'typeCheckOpsMatchingWidths' for ExtractValue
+  // instruction because its a tuple of {aggregate, index}. The first
+  // element, aggregate is result of overflow instruction. The
+  // aggregate is of 33 and 65 bits for 32 and 64 bit operands of
+  // overflow instruction respectively. The second element of
+  // ExtractValue instruction is an index value. We don't type check
+  // the operands width as the two elements vary in width.
   switch (IK) {
```
regehr: there's a better way to implement a switch with just one case

jubitaneja: Oops! I forgot to do this cleanup. Fixed now.

PR #235, opened 2016-11-09
Fixed issues for blockpcs

chenyang78:
This patch fixed following issues for blockpcs:

* When we harvested blockpcs for an instruction, we didn't handle
cases where a direct incoming branch is of a br/switch instruction.
This oversight caused us to miss optimization opportunities. This
patch fixed this issue.

* included Jubi's fix to blockpc soundness issue.

We stop harvesting blockpc when encountering a second phi node.

* don't collect UB instructions for a blockpc without predictions

Previously, we rely on the nullability of UBInstPrecondition to
separate UB instructions from PCs and BlockPCs. However, in cases
where the relevant Phi nodes are not part of the Souper IR,
we can have a null precondition for a BlockPC. Consequently,
we would introduce UB into souper. Note that we create klee exprs
for an UBInst only if it's recorded in the UBExprMap.

This patch fixed this issue by explictly check if we currently
process a PC or BlockPC.

* don't crash on revisiting phis and blocks

Because we treat select instructions as phis, it's incorrect
to assume that we would never re-visit phis or blocks during
processing blockpcs. Just skip those cases.

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -100,6 +100,7 @@ struct ExprBuilder {
   Inst *buildConstant(Constant *c);
   Inst *buildGEP(Inst *Ptr, gep_type_iterator begin, gep_type_iterator end);
   Inst *build(Value *V);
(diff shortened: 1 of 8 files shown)


regehr:
LGTM

PR #236, opened 2016-11-09
give souper an easy way to avoid triggering degeneracies (mostly in K…

regehr:
…lee) due

to large expressions


chenyang78:
LGTM. Thanks.

PR #238, opened 2016-12-13
add nop inference, which guesses that the synthesized quantity is

regehr:
equivalent to one of its inputs

souper can already do this using synthesis but it is hard: souper must
synthesize the entire RHS from scratch; just guessing that the output
is equal to an input is far easier and faster

in the future we might also guess that the output is equivalent to
some intermediate value on the LHS, this is trivial

### DIFF
--- a/lib/Extractor/Solver.cpp
+++ b/lib/Extractor/Solver.cpp
@@ -42,6 +42,9 @@ namespace {
 static cl::opt<bool> NoInfer("souper-no-infer",
     cl::desc("Populate the external cache, but don't infer replacements"),
     cl::init(false));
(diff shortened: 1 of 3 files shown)


lib/Extractor/Solver.cpp
```diff
@@ -124,6 +137,26 @@ class BaseSolver : public Solver {
       }
     }
 
+    if (InferNop) {
+      std::vector<Inst *>Guesses;
+      std::set<Inst *>Visited;
```
rsas: space after `>`

rsas:
Besides the two comments, LGTM

PR #240, opened 2016-12-15
sclang support for nop synthesis

regehr:


regehr:
I think this is trivial enough to merge w/o review

PR #241, opened 2016-12-15
in some cases Souper gets asked to infer a value that is trivially

regehr:
the same as an input-- don't do nop inference or those cases


regehr:
Trivial patch doesn't need review

PR #247, opened 2017-03-06
teach souper about data flow fact isKnownNegative

jubitaneja:
This is represented as: var (-)


regehr:
Jubi I think we want the flags to be symmetrical. That is, we don't want Negative and NonNegative. Rather, we want NonNegative and NonPositive.

regehr:
Then NonNegative + NonZero = Positive and NonPositive + NonZero = Negative

jubitaneja:
In my knowledge, there is no data flow function in LLVM to determine non-positive so far. If you want we can implement our own function, may be?

jubitaneja:
I agree it will be good to have symmetrical flags actually. Since I didn't find non-positive, I thought it will be good enough to add an extra flag.

regehr:
I see, I thought I saw a non-positive call. Let me think about this a bit more.

jubitaneja:
Sure.

lib/Parser/Parser.cpp
```diff
@@ -937,6 +938,12 @@ bool Parser::parseLine(std::string &ErrStr) {
 		  return false;
 		}
                 PowOfTwo = true;
+	      } else if (CurTok.PatternString[i] == '-') {
+		if (Negative) {
+		  ErrStr = makeErrStr(TP, "repeated '-' flag");
+		  return false;
+		}
+                Negative = true;
```
regehr: indentation looks a bit messed up here

jubitaneja: Fixed. The tabs were sitting in Parser.cpp in previous commits too. Hope it looks fine now.

PR #249, opened 2017-08-25
Renaming all data flow facts

jubitaneja:
var (nonNegative) (negative) (nonZero) (knownBits=01x) (powerOfTwo) (signBits=2)


regehr:
the lexing code seems too long and complicated. you don't need a new boolean flag for each token: just match the strings and return.

also, code in the lexer should not look across multiple tokens, so don't look for the "(" before starting to parse these things. the parser should look for "(" then a dataflow token (or tokens), then ")"

lib/Parser/Parser.cpp
```diff
@@ -163,13 +168,75 @@ Token Lexer::getNextToken(std::string &ErrStr) {
...
     const char *TokenBegin = Begin;
     do {
       ++Begin;
     } while (Begin != End && ((*Begin >= 'a' && *Begin <= 'z') ||
-             (*Begin == '.')));
-    return Token{Token::Ident, TokenBegin, size_t(Begin - TokenBegin), APInt()};
+             (*Begin == '.') || (*Begin >= 'A' && *Begin <= 'Z')));
+    std::string DataFlowFact = StringRef(TokenBegin, Begin - TokenBegin);
+    const char *PatternBegin;
+    Token T;
```
regehr: ok this code is better now, but don't declare a Token here and fill it in later, rather use the in-place constructor notation everywhere, as in return Token{.....}

lib/Parser/Parser.cpp
```diff
@@ -163,13 +168,75 @@ Token Lexer::getNextToken(std::string &ErrStr) {
...
+      (*Begin >= 'A' && *Begin <= 'Z')) {
     const char *TokenBegin = Begin;
     do {
       ++Begin;
     } while (Begin != End && ((*Begin >= 'a' && *Begin <= 'z') ||
-             (*Begin == '.')));
-    return Token{Token::Ident, TokenBegin, size_t(Begin - TokenBegin), APInt()};
+             (*Begin == '.') || (*Begin >= 'A' && *Begin <= 'Z')));
+    std::string DataFlowFact = StringRef(TokenBegin, Begin - TokenBegin);
+    const char *PatternBegin;
```
regehr: declare this lower down, at each spot where it is initialized

lib/Parser/Parser.cpp
```diff
@@ -163,13 +168,75 @@ Token Lexer::getNextToken(std::string &ErrStr) {
...
+        ++Begin;
+      if (Begin == PatternBegin) {
+        ErrStr = "expected [0|1|x]+ for knownBits";
+        return Token{Token::Error, Begin, 0, APInt()};
+      }
+      T.K = Token::KnownBits;
+      T.Pos = TokenBegin;
+      T.PatternString = StringRef(PatternBegin, Begin - PatternBegin);
+      return T;
+    } else if (DataFlowFact == "signBits") {
```
regehr: I think you should parse signBits are three tokens, not one

lib/Parser/Parser.cpp
```diff
@@ -40,8 +40,13 @@ struct Token {
...
     UntypedInt,
     KnownBits,
-    MoreKnownBits,
     NumSignBits,
+    PowerOfTwo,
+    NonNegative,
+    Negative,
+    NonZero,
+    OpenParen,
+    CloseParen,
```
regehr: you should eliminate all of these kinds of tokens. rather, they can lex as Ident.

regehr: I mean eliminate: 
￼      KnownBits,
￼ -    MoreKnownBits,
￼      NumSignBits,
￼ +    PowerOfTwo,
￼ +    NonNegative,
￼ +    Negative,
￼ +    NonZero,

lib/Parser/Parser.cpp
```diff
@@ -163,13 +168,75 @@ Token Lexer::getNextToken(std::string &ErrStr) {
...
     do {
       ++Begin;
     } while (Begin != End && ((*Begin >= 'a' && *Begin <= 'z') ||
-             (*Begin == '.')));
-    return Token{Token::Ident, TokenBegin, size_t(Begin - TokenBegin), APInt()};
+             (*Begin == '.') || (*Begin >= 'A' && *Begin <= 'Z')));
+    std::string DataFlowFact = StringRef(TokenBegin, Begin - TokenBegin);
+    const char *PatternBegin;
+    Token T;
+    if (DataFlowFact == "knownBits") {
```
regehr: it seems like there should also be a way to properly lex (knownBits=01x) as five tokens, without needing to backtrack.

What would happen if we modified UntypedInt so that it could also contain "x" characters?

regehr: or just leave this as a disgusting hack for now but make the other changes and I can try to fix it later

regehr:
LGTM

PR #250, opened 2017-10-24
updated power of two klee query with Nuno's suggestion

jubitaneja:


jubitaneja:
@nunoplopes Thanks for your review comment. 1 << bitwidth is zero but I missed out the case 1 << (bitwidth-1) that gives INT_MIN. Yes, the encoding suggested by you is much simpler.

regehr:
Good idea, thanks Nuno.  I feel bad about missing this bug in earlier code review.

nunoplopes:
(actually my comment was bogus; this patch seems correct. I fall fool of the off-by-one mistake.  The encoding I suggested should work better, though).

regehr:
Yes please adapt your pull req to use Nuno's suggestion Jubi.

jubitaneja:
Updated. Thanks.

test/Tool/more-knownbits20.opt
```diff
@@ -3,7 +3,7 @@
 ; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
```
regehr: just remove this test case, I'd say

jubitaneja: Done.

test/Tool/power2.opt
```diff
@@ -0,0 +1,55 @@
+; REQUIRES: solver
```
regehr: this one seems a bit brute-force, why not just assert that the ctpop == 1?

jubitaneja: I wrote this test case to make sure we have all power of two numbers (including INT_MIN) this time. ctpop anyways returned 1 when I didn't include 1 << (bitwidth-1) in my KLEE query earlier. I can add ctpop==1 test case too. Should I get rid of this long brute force test?

regehr: also this test passes with current Souper. so it doesn't test your new code.

jubitaneja: Yes, you're right. I have checked the leading zeros on the power of two now. This confirms the bug is fixed as the invalid and LGTM version of same test case just differ on no leading zeros for INT_MIN case.

regehr:
LGTM

PR #253, opened 2017-11-03
Blockpc fix from Yang

regehr:


rsas:
LGTM

chenyang78:
Thanks for taking care of it! @regehr @rsas

PR #254, opened 2017-11-04
stop souper from deleting instructions (which might have side effects…

regehr:
… or be terminators)

this leaves a dead instruction for LLVM to remove, if it can


regehr:
Whoops, forgot to fix a test failure.

PR #255, opened 2017-11-04
John's patch to stop souper from deleting instructions (which might h…

jubitaneja:
…ave side effects

or be terminators)

### DIFF
--- a/test/Pass/single-pass.ll
+++ b/test/Pass/single-pass.ll
@@ -6,8 +6,8 @@
 define i32 @foo(i32 %x) {
 entry:
   %add = add nsw i32 %x, 1
-  ; CHECK-NOT: icmp
   %cmp = icmp sgt i32 %add, %x
+  ; CHECK: %conv = zext i1 true to i32
   %conv = zext i1 %cmp to i32
   ret i32 %conv
 }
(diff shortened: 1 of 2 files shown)


test/Pass/single-pass.ll
```diff
@@ -6,7 +6,7 @@
 define i32 @foo(i32 %x) {
 entry:
   %add = add nsw i32 %x, 1
-  ; CHECK-NOT: icmp
+  ; CHECK: icmp
```
regehr: with this change you're looking for an icmp but an icmp already existed in the test case, so the test becomes useless. the point of this test is to look for an optimization happening. it's no longer shown by eliminating the icmp, you need to look at something different.

jubitaneja: I have fixed it. It tests the current functionality that instruction is not deleted but the uses are replaced by value, so we don't look for the uses of instruction now.

regehr: I think it'd be better to check for the whole line we want:
%conv = zext i1 true to i32
(or whatever it really is)

test/Pass/single-pass.ll
```diff
@@ -6,8 +6,9 @@
...
```
regehr: also delete this CHECK

jubitaneja:
Fixed.

regehr:
Looks good to me -- do you think this makes sense, Raimondas or Yang?

rsas:
LGTM, but better @chenyang78 should look into this as well.

chenyang78:
Yes, LGTM, too. Thanks.

PR #256, opened 2017-11-05
Keep counterexample wirings

rsas:
This fix cuts the test time of HackersDelight synthesis tests from 17 to 9 seconds on my iMac.


regehr:
LGTM

regehr:
This is great-- synthesis speedups will make a big difference in some upcoming Souper runs that I plan to do!

PR #257, opened 2017-11-17
integer synthesis on by default, and tweak constant guessing strategy

regehr:
this patch does a few easy things:

- turn on integer synthesis by default for things bigger than i1 (not sure why we didn't do this long ago)
- fix resulting testcase breakage by turning off iN synthesis in a number of places
- tweak the integer guessing strategy: oftentimes, mutliple integers are possible and the choice is up to the solver, with this patch we guess 0, 1, -1 before letting the solver pick; the rationale is that these constants should be easier for the backend to cook up


rsas:
LGTM

PR #258, opened 2017-11-22
a broken test case revealed a bug in souper's test harness where it

regehr:
wasn't checking that an APInt in test metadata had a width matching
the APInt computed by Souper

this patch fixes both problems


regehr:
merging as obviously ok....

PR #260, opened 2017-11-23
build LLVM with assertions

regehr:


regehr:
merging w/o review as this is obviously desirable

PR #262, opened 2017-11-26
this patch makes ASan+UBsan happy with Souper

regehr:


lib/Extractor/KLEEBuilder.cpp
```diff
@@ -351,7 +351,7 @@ ref<Expr> ExprBuilder::buildAssoc(
 
 ref<Expr> ExprBuilder::countOnes(ref<Expr> L) {
    Expr::Width Width = L->getWidth();
-   ref<Expr> Count =  klee::ConstantExpr::create(0, Width);
+   ref<Expr> Count =  klee::ConstantExpr::alloc(llvm::APInt(Width, 0));
```
jeroenk: What's the problem with `create`? As far as I can tell that does an `alloc`.

regehr: when called with Width==256 (in one of our tests) there's an UB in Klee that this works around
probably need to audit all calls to ConstExpr::create but for now this gives a clean build

regehr: I could submit a PR for Klee but we're using such an old, hacked up version that I have no idea if the problem's still there

jeroenk: The ConstantExpr code doesn't seem to have changed. It's in `include/klee/Expr.h`. In any case, I don't think this should stop this from going in.

It's not clear to me what the UB is, by the way.

jeroenk:
LGTM.

regehr:
Thanks!
I'll send you a stack trace in email

PR #267, opened 2017-11-30
a huge number of block predecessors was getting turned into ~0

regehr:
and then triggering a tombstone value in LLVM's DenseMap.

a sensible way to fix seems to be to put an arbitrary (but enormous)
limit on the number of predecessors


chenyang78:
LGTM. Thanks.

PR #269, opened 2017-12-01
dataflow parsing was doing OOB shifts of APInts sometimes, to fix

regehr:
these we do the length test before the operations that do shifts

this changes some error messages but not in a bad way, the tests are
legitimately wrong in multiple ways, so this should be OK


jubitaneja:
LGTM. Thanks!

PR #270, opened 2017-12-01
overflow intrinsics are not allowed to be inferred, only

jubitaneja:
the overflow bit or operand's bitwidth result can be returned

souper-check returned protocol error earlier when overflow intrinsic was inferred.


regehr:
LGTM

PR #271, opened 2017-12-01
absurd argument values were causing APInt to assert out, and

regehr:
there's a better way to compare APInts anyway


jubitaneja:
LGTM. Thanks!

PR #272, opened 2017-12-04
souper's parser has some basic forward type inference where widths

regehr:
don't need to always be specified

some error checking specific to ctpop/cttz/ctlz/bswap was breaking
type inference for those instructions since the checking was done too
early, this PR fixes that


regehr:
oops, working in a boring meeting and merged w/o thinking
but I think this trivial patch is pretty obviously ok

PR #273, opened 2017-12-04
overflow intrinc can only be an argument of extractvalue instruction

jubitaneja:


regehr:
wouldn't it be easier to do everything in one place at Parser.cpp:507?

jubitaneja:
Oops! I need to handle this for phi instruction as well. Updating my patch now.

lib/Parser/Parser.cpp
```diff
@@ -1132,6 +1159,23 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+          case Inst::SAddWithOverflow:
+          case Inst::UAddWithOverflow:
+          case Inst::SSubWithOverflow:
+          case Inst::USubWithOverflow:
+          case Inst::SMulWithOverflow:
+          case Inst::UMulWithOverflow:
+            OverflowFlag = true;
+            break;
+          default:
+            OverflowFlag = false;
```
regehr: this is already false so I don't think you need to set it

jubitaneja:
Updated. Thanks.

regehr:
this is too many repetitions of the same case statement, the overflow check needs to be factored out into a separate function

regehr:
also if there's only one case you don't really want a switch

jubitaneja:
Sure, I will fix the code. Thanks.

lib/Parser/Parser.cpp
```diff
@@ -386,6 +387,14 @@ Inst *Parser::parseInst(std::string &ErrStr) {
   }
 }
 
+bool Parser::isOverflow(Inst::Kind IK) {
+  if (IK == Inst::SAddWithOverflow || IK == Inst::UAddWithOverflow ||
```
regehr: you don't need an "if" in this kind of function, just return the result of the or-expression

lib/Parser/Parser.cpp
```diff
@@ -576,10 +592,19 @@ bool Parser::typeCheckInst(Inst::Kind IK, unsigned &Width,
...
   // ExtractValue instruction is an index value. We don't type check
   // the operands width as the two elements vary in width.
-  if (IK != Inst::ExtractValue)
+  if (IK != Inst::ExtractValue) {
     if (!typeCheckOpsMatchingWidths(OpsMatchingWidths, ErrStr))
       return false;
 
+    for (auto Op : Ops) {
+      if (isOverflow(Op->K)) {
+        ErrStr = std::string("overflow intrinsic cannot be an operand of ") + Inst::getKindName(IK) +
```
regehr: I don't think both of these explicit casts to std::string are needed

jubitaneja: It does give build error if I don't typecast explicitly. For the other one, which doesn't combine multiple strings, I don't need typecast. I have fixed that.

regehr:
LGTM

PR #274, opened 2017-12-06
propagate LLVM_PROFILE_FILE environment variable into tests

regehr:


regehr:
merging as trivial and obvious

PR #276, opened 2017-12-21
argh, DemandedBits needs a default value or else it doesn't work to

regehr:
create custom Souper LHSs from scratch, as we want to do in the
Solver, this was causing ugly wrong behaviors

I think this fixes an issue Raimondas and I were having


jubitaneja:
LGTM. Thanks.

Issue #252, opened 2017-10-26
Don't synthesize a component with all constant inputs


rsas:
The synthesis of the following snippet fails at the moment:

```
%0 = block 3
%1:i32 = var
%2:i32 = and 1:i32, %1
%3:i1 = eq 0:i32, %2
blockpc %0 0 %3 1:i1
blockpc %0 1 %3 0:i1
%4:i32 = and 2:i32, %1
%5:i1 = eq 0:i32, %4
blockpc %0 1 %5 1:i1
blockpc %0 2 %3 0:i1
blockpc %0 2 %5 0:i1
%6:i32 = or 1:i32, %1
%7:i32 = and 4294967292:i32, %1
%8:i32 = or 2:i32, %7
%9:i32 = phi %0, %6, %8, 0:i32
%10:i32 = and 3:i32, %9
infer %10

;%11:i32 = add %1, 1:i32
;%12 = and %11, 3:i32
;result %12
```

```
souper-check $SOUPER_SOLVER -infer-rhs -souper-infer-inst -souper-synthesis-comps=const,const,and,add -souper-synthesis-debug-level=3 add_final.opt
```

I see a number of candidates similar to this:

```
candidate:
%26:i32 = and 0:i32, 4013932537:i32
result %26
```

This is not ok, because no component should have inputs consisting only of constants.

rsas:
Done.

PR #277, opened 2017-12-29
No candidates shall contain all operands as constants (refs #252)

rsas:


lib/Infer/InstSynthesis.cpp
```diff
@@ -1244,6 +1293,16 @@ bool InstSynthesis::isWiringInvalid(const LocVar &Left, const LocVar &Right) {
           InvalidWirings.count(std::make_pair(Right, Left)));
 }
 
+bool InstSynthesis::isInputConst(const LocVar &Loc) {
+  if (Loc.first != 0)
+    return false;
+
+  if (!LocInstMap.count(getLocVarStr(Loc, CONST_PREFIX)))
```
regehr: I think this statement would be better written as "return LocInstMap.count(getLocVarStr(Loc, CONST_PREFIX) != 0;"

regehr:
besides the one comment, LGTM, please squash and merge

lib/Infer/InstSynthesis.cpp
rsas: Sure. The statement is ugly anyway, the data structures in synthesis "evolved" over time and are far away from being well-designed.

PR #280, opened 2018-01-03
use travis for testing but do it on a pre-built docker image so we

regehr:
don't have to compile LLVM every time


rsas:
LGTM, I assume you tested it. Please merge.

rsas:
Squash before, of course.

PR #283, opened 2018-01-08
run CI tests under UBSan

regehr:
this has the happy side effect of having CI make sure that souper builds and works using both LLVM and GCC


regehr:
merging as clearly useful

PR #284, opened 2018-01-10
I've been writing quite a bit of souper lately and so have been seeing

regehr:
a lot of error messages and this one can be improved a bit by
mentioning the problematic token instead of just saying its location


jubitaneja:
LGTM. Thanks.

PR #285, opened 2018-01-12
new mode for souper-check that takes a complete replacement, forgets

regehr:
the RHS, asks souper to synthesize a RHS, and then compares the cost
of the old and new RHSs

to be used in combination with lots of souper optimizations ported
over from Alive

this isn't necessarily quite ready to go (needs some tests for example) but wanted to bring it to Raimondas's attention

### DIFF
--- a/tools/souper-check.cpp
+++ b/tools/souper-check.cpp
@@ -39,6 +39,10 @@ static cl::opt<bool> InferRHS("infer-rhs",
     cl::desc("Try to infer a RHS for a Souper LHS (default=false)"),
     cl::init(false));
 
+static cl::opt<bool> ReInferRHS("reinfer-rhs",
+    cl::desc("Try to infer a new RHS and compare its cost with the existing RHS (default=false)"),
+    cl::init(false));
+
 static cl::opt<bool> ParseOnly("parse-only",
     cl::desc("Only parse the replacement, don't call isValid() (default=false)"),
     cl::init(false));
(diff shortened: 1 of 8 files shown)


rsas:
I'm trying to understand how does this differ from the cost checks during the synthesis. At the moment, I'm discarding candidates that have higher cost during synthesis except if you pass the ``-souper-synthesis-ignore-cost`` flag. @regehr can you explain how these two should coexist together?

regehr:
yeah, as you noticed, I moved the cost out of Infer.cpp so that it can be used more broadly. there's work to be done on the cost model but for now the simple one is pretty ok.

tools/souper-check.cpp
```diff
@@ -70,15 +74,34 @@ int SolveInst(const MemoryBufferRef &MB, Solver *S) {
...
       }
       if (Rep.Mapping.RHS) {
-        llvm::outs() << "; RHS inferred successfully\n";
+        ++Success;
+        if (ReInferRHS) {
+          int NewCost = cost(Rep.Mapping.RHS);
+          if (NewCost <= OldCost)
+            llvm::outs() << "; RHS inferred successfully, no cost regression (Old= " <<
+              OldCost << ", New= " << NewCost << ")\n";
+          else
```
rsas: We will need to enable ``-souper-synthesis-ignore-cost`` per default, otherwise this branch will be never executed.

regehr: I see what you mean.
I need to look into how the cost function plays into synthesis, will report back.

regehr:
wait a minute, both of those branches ("no cost regression" and "cost regressed") can fire without -souper-synthesis-ignore-cost.

OldCost refers to the optimization that Alive came up with and we erase it before running synthesis. So whatever it cost, good or bad, has no effect on synthesis.

rsas:
Probably I overlookeed something, I’ll check.

Raimondas

rsas:
Without -souper-synthesis-ignore-cost enabled, synthesis will return failure if the cost(candidate) >= cost(LHS). For example, assume your LHS is of cost 5 and your current RHS (aka OldRHS) is 2. You start reinference with -souper-synthesis-ignore-cost=True and assume synthesis fails to find the small RHS, but it succeeds finding the original LHS. In this case it would be a successful synthesis but with regression of 3 and we want log that. With -souper-synthesis-ignore-cost disabled (default), the synthesis would fail to return the original LHS and you log a synthesis failure instead of cost regression. Does it make sense?

regehr:
OK, that makes sense. I agree we'll most likely always use reinfer mode along with ignoring the cost function.

regehr:
OK, I've added test cases, this is ready for review

rsas:
LGTM, please merge

PR #286, opened 2018-01-12
add a lot of tests translated from Alive; they're good souper tests

regehr:
but really we want to use them to test synthesis


rsas:
LGTM, please merge

PR #287, opened 2018-01-13
this version of nop synthesis is still slow, but it's not that slow

regehr:
and it should find a nop most of the time when one exists; the
strategy is to choose up to 20 candidate nops using BFS starting from
the bottom of the LHS

### DIFF
--- a/include/souper/Inst/Inst.h
+++ b/include/souper/Inst/Inst.h
@@ -105,6 +105,7 @@ struct Inst : llvm::FoldingSetNode {
   unsigned Number;
   unsigned Width;
   Block *B;
+  bool Available = true;
   llvm::APInt Val;
   std::string Name;
   std::vector<Inst *> Ops;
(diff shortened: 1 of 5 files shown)


rsas:
What's the state of this pull request? Have you run more experiments with the BFS selection strategy?

regehr:
I haven't!  will do that.

But in my opinion this PR isn't a throwaway. Rather, it complements the fast nop synthesis that is developing elsewhere. By limiting the number of candidates it'll stop query size from getting out of control. So I think it's something that we want.

regehr:
Raimondas, is it OK to merge this if I promise to (later) investigate what is the proper setting for MAX_NOPS?

rsas:
Sure, I have no objections.

regehr:
Ok, thanks!

I can merge this with your nop branch or you can. There'll definitely be (I hope easy) conflicts.

PR #288, opened 2018-01-16
this change makes the souper parser do stricter checking of constants:

regehr:
we used to just silently truncate whatever value was parsed to the
specified bitwidth, but this masked bugs. now we ensure that the
specified constant con be losslessly converted into the desired
bitwidth.

a slight wrinkle in the plan is that we have examples where "lossless"
depends on signed conversion, and on unsigned conversion, and I don't
know a good way to automatically disambiguate. so the code here checks
if either one works. this doesn't seem satisfying but I don't see a
better answer.


rsas:
LGTM, please merge.

PR #290, opened 2018-01-24
compile with -Werror on the CI machine

regehr:


regehr:
trivial change that I'm merging w/o review, it simply turns on -Werror on travis, which should never cause problems since the compilers there are under our control (currently clang++ 5.0 and whatever g++ comes with Ubuntu xenial).

PR #291, opened 2018-01-29
rest of alive tests

regehr:
This is the rest of ~400 instcombine transformations that Nuno translated into Alive. I've translated them into Souper. They contain some nice UB on RHS, so will be good tests for us when we get that all working properly. But the main purpose of these is to serve as synthesis benchmarks, they fit nicely with souper-check in "reinfer" mode.


rsas:
LGTM

PR #292, opened 2018-01-29
this ignore-solver-errors thing was a bad hack, let's get rid of it

regehr:
There was this ignore-solver-errors feature that basically just made sclang continue when the solver hit value_too_large errors (which are benign). But it was annoying to remember to set it and also it made sclang ignore more serious errors. I no longer think this feature was a good idea. Let's explicitly ignore solver timeouts and value_too_large in the pass and make sure other errors get propagated out.


rsas:
LGTM, please merge.

PR #294, opened 2018-02-01
Commit for abs support

zhengyangl:
This commit enables souper to optimize abs() function call. Some details:
* A simple test case is added. This test will not call InstCombine, the result is just generated by souper.
* Fixed the missing default label of switch.

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -37,7 +37,6 @@
 #include <unordered_set>
 #include <tuple>
 #include "llvm/Analysis/ValueTracking.h"
-#include "llvm/Analysis/TargetLibraryInfo.h"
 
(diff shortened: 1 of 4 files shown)


lib/Extractor/Candidates.cpp
```diff
@@ -37,6 +37,7 @@
 #include <unordered_set>
 #include <tuple>
 #include "llvm/Analysis/ValueTracking.h"
+#include "llvm/Analysis/TargetLibraryInfo.h"
```
chenyang78: Random comment - usually we try to keep alphabetical order for the including header files.

zhengyangl: Thanks for the tip! This include is removed, since TargetLibraryInfo.h is already included in Candidates.h :-)

regehr:
LGTM, thanks!

PR #296, opened 2018-02-02
Fixed a bug caused by NULL returned from CallInst::getCalledFunction()

zhengyangl:
According to the source [1], the CallInst::getCalledFunction() returns NULL pointer if callee is not a llvm::Function. This patches fixed the bug where getLibFunc() tries to get the information from a NULL pointer.

[1] http://llvm.org/doxygen/Instructions_8h_source.html#l01873

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -489,18 +489,20 @@ Inst *ExprBuilder::build(Value *V) {
           return IC.getInst(Inst::UMulWithOverflow, L->Width+1, {Mul, Overflow});
         }
       }
-    } else if (Call->getCalledFunction() &&
-               TLI->getLibFunc(*Call->getCalledFunction(), Func) && TLI->has(Func)) {
-      switch (Func) {
-        case LibFunc_abs: {
-          Inst *A = get(Call->getOperand(0));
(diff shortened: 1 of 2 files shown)


lib/Extractor/Candidates.cpp
```diff
@@ -489,7 +489,8 @@ Inst *ExprBuilder::build(Value *V) {
           return IC.getInst(Inst::UMulWithOverflow, L->Width+1, {Mul, Overflow});
         }
       }
-    } else if (TLI->getLibFunc(*Call->getCalledFunction(), Func) && TLI->has(Func)) {
+    } else if (Call->getCalledFunction() &&
```
regehr: do you want to put the pointer into a variable install of calling getCalledFunction() twice?

zhengyangl: I tried to do initialization and boolean calculation in a single if statement to stop bringing new blocks, but it is not possible. Anyway, a new patch is submitted.

regehr:
LGTM, thanks.

regehr:
Zhengyang-- can you prepare another pull req that changes the "clang++" command in this test (and there is one other test that needs this change) to use a substitution to make sure the correct clang is called? We do not want to rely on the proper version of clang being in the user's path already.

The substitutions are in test/lit.cfg

PR #297, opened 2018-02-02
Add a new substitution rule for handling clang++ tests

zhengyangl:
I've tested and this rule works correctly. The other clang related (sclang) tests also do the substitute correctly (calls the executables from third_party/llvm).


regehr:
Great, please also modify the tests to call %clang instead of clang.

regehr:
LGTM, thanks!  Will merge once tests pass.

PR #298, opened 2018-02-03
I noticed that our testing of the opt pass and sclang is quite

regehr:
weak. These tests start to address that. The trick is to write code
that is robustly optimized by souper and robustly not-optimized by
LLVM. Hopefully these accomplisht that.


regehr:
any comments on this? it's pretty basic stuff, just trying to make sure our pass gets stress-tested a bit

rsas:
LGTM, please merge

PR #299, opened 2018-02-13
Nop Pass Support

zhengyangl:


zhengyangl:
Will upload various test cases in a new commit soon.

regehr:
Ok, great.
Can you add a couple of test cases for the pass? These would be .ll files that contain nops, to make sure that the optimization works?

zhengyangl:
@regehr Now I am testing the robustness of PR by compiling SPEC CINT17. I will send the commit with test cases afterward.

regehr:
One test case should correspond to a failed dominator check.

regehr:
Great!
If it is going too slowly, you can reduce the value of MaxNops in Solver.cpp, for example to 10 instead of 20.

test/Pass/nop-domination-check.c
```diff
@@ -0,0 +1,33 @@
+// REQUIRES: solver
+
+// RUN: env SOUPER_NO_EXTERNAL_CACHE=1 SOUPER_INFER_NOP=1 SOUPER_SOLVER=%solver %sclang -O2 -c -o - %s
+
```
regehr: do you want some kind of CHECK or CHECK-NOT directive here?

zhengyangl: This test will crash (lit will report the crash) if domination check is disabled. I guess we don't need CHECK or CHECK-NOT.

zhengyangl: Or I can add a CHECK-NOT for crash dump if you want.

zhengyangl: @regehr Fixed

regehr:
maybe just add a comment that the test is intended to crash if it fails so no check lines are needed

zhengyangl:
Error message by nop-domination-check.c when domination check is disabled  :

```
FAIL: Souper :: Pass/nop-domination-check.c (183 of 315)
******************** TEST 'Souper :: Pass/nop-domination-check.c' FAILED ********************
Script:
--
env SOUPER_NO_EXTERNAL_CACHE=1 SOUPER_INFER_NOP=1 SOUPER_SOLVER=-z3-path=/usr/bin/z3 /home/liuz/workspace/souper/build/sclang -O2 -c -o - /home/liuz/workspace/source/souper/test/Pass/nop-domination-check.c 2>&1 | /home/liuz/workspace/source/souper/third_party/llvm/Release/bin/FileCheck /home/liuz/workspace/source/souper/test/Pass/nop-domination-check.c
--
Exit Code: 1

Command Output (stderr):
--
<stdin>:35:1: error: CHECK-NOT: string occurred!
Instruction does not dominate all uses!
^
/home/liuz/workspace/source/souper/test/Pass/nop-domination-check.c:8:15: note: CHECK-NOT: pattern specified here
// CHECK-NOT: Instruction does not dominate all uses!
              ^

--

********************
```

regehr:
LGTM.

PR #300, opened 2018-02-14
faster nop synthesis

regehr:
Ok, this PR takes some code from Raimondas to do nop synthesis faster and does some refactoring to try to make it generally useful for new synthesis code.
The new nop algorithm is:
1. do a big query to see if any value works as a nop
2. if yes, check each value to find which one it is that works
Feedback appreciated.


rsas:
LGTM, please review and merge my demanded bits pull request first.

PR #302, opened 2018-02-15
Stop building UB Path conditions for processed blocks

chenyang78:
Because we treat a select instruction the same way as we do for
a phi instruction, it's possible to revisit a block from
a select instruction. Consequently, our previous assumption
did hold.

This patch fixed the issue 289 and issue 301 by simply returning true
upon re-visiting a block.


chenyang78:
I signed it!

regehr:
LGTM, thanks!

PR #305, opened 2018-02-24
Pass Support for Inst Synthesis

zhengyangl:
This is just a working version (passes all the test cases), more test cases will be added tomorrow.


test/Pass/syn-inst-add-ext-use.ll
```diff
@@ -0,0 +1,27 @@
...
+
+; RUN: llvm-as -o %t %s
+; RUN: opt -load %pass -souper %solver -souper-infer-inst -souper-synthesis-comps=add,const -S -o - %s | FileCheck %s
+
+; Investigation and improvement are needed. Current instruction synthesis can
+; not optimization instruction with outside uses smartly. This case shows this
+; issue: %a, %b and %c all have external uses, and they should not be swiped
+; out, thus synthesis a new instruction "add %x, 4" only for replacing %d is not
+; necessary. Now souper does this redundant optimization. 
+
```
zhengyangl: This describes the issue for not able to handle insts with external uses.

test/Pass/syn-inst-add.ll
```diff
@@ -0,0 +1,15 @@
...
+; RUN: llvm-as -o %t %s
+; RUN: opt -load %pass -souper %solver -souper-infer-inst -souper-synthesis-comps=add,const -S -o - %s | FileCheck %s
+
+
+define i32 @foo(i32 %x) {
+entry:
+  %a = add i32 %x, 1
+  %b = add i32 %a, 1
+  %c = add i32 %b, 1
+  ;CHECK: add i32 %x, 4
```
regehr: maybe also add a CHECK-NOT for each of the add-1 lines

lib/Pass/Pass.cpp
```diff
@@ -279,12 +343,14 @@ struct SouperPass : public ModulePass {
       }
       if (!Cand.Mapping.RHS)
         continue;
-      // TODO: add non-const instruction support
+
       Instruction *I = Cand.Origin.getInstruction();
       Instruction *ReplacedInst = Cand.Origin.getInstruction();
+      IRBuilder<> Builder(ReplacedInst->getParent());
+      Builder.SetInsertPoint(ReplacedInst);
```
jeroenk: Why not use ```IRBuilder<> Builder(ReplacedInst);```?

jeroenk: Also. Why is the builder created here and not in the getValue function?

zhengyangl: Thanks for this review!

zhengyangl: Reason for IR builder created outside getValue() function: getValue() is a recursive function, we don't want to create an IR builder in each recursive call.

zhengyangl:
Currently the annotation in these test cases does not correctly handle the checking on commutative operators (say mul i64 %x, 24). Need to find a solution for this issue.

zhengyangl:
When dealing with binop instruction generation in getValue(), we can do a little bit canonicalization (Swap constants to the second operands if the operator is commutative).

regehr:
Great. Yes, we should emit canonical code whenever possible!

regehr:
LGTM, thanks!

PR #306, opened 2018-02-28
stop making nop candidates when we reach a phi, in order to stop

regehr:
miscompilations where Souper is insufficiently respectful of the
differences between phi arguments that look the same, causing
miscompilations in the Souper Pass when multiple optimizations are
performed following a single extraction of LHSs.

this is a conservative fix that causes us to miss some optimizations,
we should figure out a better way, but for now this eliminates the
miscompilations.


rsas:
LGTM

PR #313, opened 2018-03-08
xfail tests for UB

regehr:
OK, I don't want to actually merge this right now, I just want it sitting here where people can see it.

@rsas and @zhengyangl, please issue a pull req to my branch if you find an interesting test case where souper's UB model prevents it from doing this right thing. We'll use these to help us get UB right.


rsas:
I recall having a couple of those, I'll try to find them again.

PR #314, opened 2018-03-08
Add % before llvm-as, llvm-dis, opt, FileCheck command calls in test suite.

zhengyangl:
Before this commit, these commands can be substituted with a full path to the executables in third_party/llvm/Release/bin already. This commit only makes it clear: we're not using the executables from $PATH.


regehr:
LGTM, thanks.

PR #317, opened 2018-03-18
The question is: how do we (in the pass) remember that mapping from

regehr:
Souper instructions to LLVM instructions?

From early in, souper has maintained an Origin pointer as part of the
mapping. In other words, a mapping from LHS or RHS remembers which
LLVM instruction is getting replaced. This worked only as long as we
were synthesizing constant, it does not support more interesting
replacements.

Later, I added an Origins multimap to the expression builder context
object which tracked the LLVM origin of every Souper instruction in
that context; this suffices to support replacements based on full
synthesis. However, just lately it became a pain because we've started
(for multiple reasons) making copies of souper instructions, which
required updating the Origins multimap, resulting in overhead when
many copies were made.

So this PR stores origins in the Inst class, avoiding the need to
update an external map when we copy instructions. I was hoping this
would support removing the Origin part of the Mapping, but I've left
that intact since it remembers the single origin for that mapping,
whereas the instruction may have multiple origins.

Finally, the Pass has a separate map to track origins for instructions
that have already been replaced in this run of the pass -- we want to
avoid adding uses of those. The new test case for this PR triggers
that behavior: Souper performs a nop and then a second nop wants to
refer to the first one, and this gets forwarded to the RHS of the
first replacement.

### DIFF
--- a/include/souper/Inst/Inst.h
+++ b/include/souper/Inst/Inst.h
@@ -18,6 +18,7 @@
 #include "llvm/ADT/APInt.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/FoldingSet.h"
+#include "llvm/IR/Value.h"
 #include <map>
 #include <memory>
 #include <set>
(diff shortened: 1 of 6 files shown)


regehr:
can someone take a look at this? @zhengyangl? I'd like to merge this before it goes stale

zhengyangl:
Sure @regehr

include/souper/Inst/Inst.h
```diff
@@ -137,6 +140,7 @@ struct Inst : llvm::FoldingSetNode {
   bool Negative;
   unsigned NumSignBits;
   llvm::APInt DemandedBits;
+
```
zhengyangl: Unnecessary new line.

regehr:
thanks!

PR #321, opened 2018-03-19
Handling values with external uses

zhengyangl:
I have just upload a commit for external use tagging support for Souper candidate extractor. This is the very first stage of this Pull Request. Here are the details:

I've added a new std::set<Inst *> field called DepsWithExternalUses to Inst. This set stores any ancestor Insts with external uses for that Inst.

I've also added a markExternalUses function. The function takes an Inst and fills the DepsWithExternalUses for that Inst. The algorithm works as follows: First count the number of uses in Souper IR for each ancestor Insts. Then compare the number with the use number in original LLVM IR. If they're different,  we can conclude that the value in LLVM must have external uses, and we insert the corresponding Inst to DepsWithExternalUses.

I've also added a printInstWithExtUsesAnnotations function. This function is an alternative Inst printer and it adds the external tagging to its output. This function will be merged to printInst function later.

Given a input program:
```llvm
define i32 @foo(i32 %x) {
entry:
  %a = add i32 %x, 1
  store i32 %a, i32* @amem, align 4
  %b = add i32 %a, 1
  store i32 %b, i32* @bmem, align 4
  %c = mul i32 %b, 2
  store i32 %c, i32* @cmem, align 4
  %d = add i32 %c, 1
  ret i32 %d
}
```
The printInstWithExtUsesAnnotations will print the following Souper Inst for %d,
```
%0:i32 = var
%1:i32 = add 1:i32, %0 (hasExternalUses)
%2:i32 = add 1:i32, %1 (hasExternalUses)
%3:i32 = mul 2:i32, %2 (hasExternalUses)
%4:i32 = add 1:i32, %3
```

### DIFF
--- a/include/souper/Inst/Inst.h
+++ b/include/souper/Inst/Inst.h
@@ -23,6 +23,7 @@
 #include <memory>
 #include <set>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 namespace souper {
@@ -112,6 +113,7 @@ struct Inst : llvm::FoldingSetNode {
   bool Available = true;
   llvm::APInt Val;
   std::string Name;
+  std::unordered_set<Inst *> DepsWithExternalUses;
   std::vector<Inst *> Ops;
   mutable std::vector<Inst *> OrderedOps;
   std::vector<llvm::Value *> Origins;
@@ -169,6 +171,7 @@ class ReplacementContext {
   llvm::DenseMap<Block *, std::string> BlockNames;
   std::map<std::string, Inst *> NameToInst;
   std::map<std::string, Block *> NameToBlock;
+  std::string printInstImpl(Inst *I, llvm::raw_ostream &Out, bool printNames, Inst *OrigI);
 
 public:
   void printPCs(const std::vector<InstMapping> &PCs,
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -105,6 +105,7 @@ struct ExprBuilder {
                          std::unordered_set<Block *> &VisitedBlocks,
                          BasicBlock *BB);
   Inst *get(Value *V);
+  void markExternalUses(Inst *I);
 };
 
 }
@@ -214,6 +215,31 @@ Inst *ExprBuilder::buildGEP(Inst *Ptr, gep_type_iterator begin,
     }
   }
   return Ptr;
+}
+
+void ExprBuilder::markExternalUses (Inst *I) {
+  std::map<Inst *, unsigned> UsesCount;
+  std::vector<Inst *> Stack;
+  Stack.push_back(I);
+  while(!Stack.empty()) {
+    Inst* T = Stack.back();
+    Stack.pop_back();
+    if (I != T) {
+      if (UsesCount.find(T) == UsesCount.end())
+        UsesCount[T] = 1;
+      else
+        UsesCount[T]++;
+    }
+    for (auto Op: T->Ops) {
+      if (Op->K != Inst::Const && Op->K != Inst::Var
+          && Op->K != Inst::UntypedConst && Op->K != Inst::Phi)
+        Stack.push_back(Op);
+    }
+  }
+  for (auto U : UsesCount)
+    for (auto R : EBC.InstMap)
+      if (R.second == U.first && R.first->getNumUses() != U.second)
+        I->DepsWithExternalUses.insert(U.first);
 }
 
 Inst *ExprBuilder::build(Value *V) {
@@ -738,7 +764,9 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
     std::unique_ptr<BlockCandidateSet> BCS(new BlockCandidateSet);
     for (auto &I : BB) {
       if (I.getType()->isIntegerTy()) {
-        BCS->Replacements.emplace_back(&I, InstMapping(EB.get(&I), 0));
+        Inst *In = EB.get(&I);
+        EB.markExternalUses(In);
+        BCS->Replacements.emplace_back(&I, InstMapping(In, 0));
         assert(EB.get(&I)->hasOrigin(&I));
       }
     }
--- a/lib/Inst/Inst.cpp
+++ b/lib/Inst/Inst.cpp
@@ -92,6 +92,12 @@ const std::vector<Inst *> &Inst::orderedOps() const {
 
 std::string ReplacementContext::printInst(Inst *I, llvm::raw_ostream &Out,
                                           bool printNames) {
+  return printInstImpl(I, Out, printNames, I);
+}
+
+std::string ReplacementContext::printInstImpl(Inst *I, llvm::raw_ostream &Out,
+                                              bool printNames, Inst *OrigI) {
+
   std::string Str;
   llvm::raw_string_ostream SS(Str);
 
@@ -131,7 +137,7 @@ std::string ReplacementContext::printInst(Inst *I, llvm::raw_ostream &Out,
       OpsSS << ", ";
     switch (I->K) {
       default:
-        OpsSS << printInst(Ops[Idx], Out, printNames);
+        OpsSS << printInstImpl(Ops[Idx], Out, printNames, OrigI);
         break;
       case Inst::SAddWithOverflow:
       case Inst::UAddWithOverflow:
@@ -139,7 +145,7 @@ std::string ReplacementContext::printInst(Inst *I, llvm::raw_ostream &Out,
       case Inst::USubWithOverflow:
       case Inst::SMulWithOverflow:
       case Inst::UMulWithOverflow:
-        OpsSS << printInst(I->Ops[1]->Ops[Idx], Out, printNames);
+        OpsSS << printInstImpl(I->Ops[1]->Ops[Idx], Out, printNames, OrigI);
         break;
     }
   }
@@ -177,6 +183,10 @@ std::string ReplacementContext::printInst(Inst *I, llvm::raw_ostream &Out,
           Out << " (signBits=" << I->NumSignBits << ")";
       }
       Out << OpsSS.str();
+
+      if (OrigI->DepsWithExternalUses.find(I) != OrigI->DepsWithExternalUses.end())
+        Out << " (hasExternalUses)";
+
       if (printNames && !I->Name.empty())
         Out << " ; " << I->Name;
       Out << '\n';
--- a/lib/Parser/Parser.cpp
+++ b/lib/Parser/Parser.cpp
@@ -23,6 +23,7 @@
 #include "souper/Inst/Inst.h"
 
 #include <string>
+#include <unordered_set>
 
 using namespace llvm;
 using namespace souper;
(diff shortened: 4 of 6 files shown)


zhengyangl:
To mark a value as a value with external uses, just add the string "(hasExternalUses)" at the end of the line.

'''
%0:i32 = var (hasExternalUses)
%1:i32 = add 0:i32, %0 (hasExternalUses)
%2:i32 = add 0:i32, %1 (hasExternalUses)
%3:i32 = add 0:i32, %2 (hasExternalUses)
%4:i32 = add 0:i32, %3
'''

regehr:
please take a look at the failing test case flagged by Travis

zhengyangl:
@regehr sure.

lib/Infer/InstSynthesis.cpp
```diff
@@ -1469,8 +1469,9 @@ void findCands(Inst *Root, std::vector<Inst *> &Guesses, InstContext &IC,
     ++Benefit;
     if (Visited.insert(I).second) {
       if (I->K != Inst::Phi) {
-        for (auto Op : I->Ops)
+        for (auto Op : I->Ops) {
```
regehr: I think we should choose one of using parens for single-line loops/conditionals or not using them, and try to be consistent, but either way I don't think you should change this in existing code except in a cleanup patch

lib/Inst/Inst.cpp
```diff
@@ -173,6 +176,11 @@ std::string ReplacementContext::printInst(Inst *I, llvm::raw_ostream &Out,
           Out << " (signBits=" << I->NumSignBits << ")";
       }
       Out << OpsSS.str();
+
+      if (OrigI && OrigI->DepsWithExternalUses.find(I) != OrigI->DepsWithExternalUses.end()) {
```
regehr: let's go without the curly braces in cases like this

lib/Parser/Parser.cpp
```diff
@@ -1137,6 +1145,11 @@ bool Parser::parseLine(std::string &ErrStr) {
         }
         Inst *I = IC.createVar(InstWidth, InstName, Zero, One, NonZero,
                                NonNegative, PowOfTwo, Negative, SignBits);
+        if (hasExternalUses)
+          ExternalUsesSet.insert(I);
+        for (auto EU: ExternalUsesSet) {
```
regehr: also here

lib/Parser/Parser.cpp
```diff
@@ -1220,7 +1253,11 @@ bool Parser::parseLine(std::string &ErrStr) {
             break;
         }
       }
-
+      if (hasExternalUses)
+        ExternalUsesSet.insert(I);
+      for (auto EU: ExternalUsesSet) {
```
regehr: also here

lib/Infer/InstSynthesis.cpp
zhengyangl: Sure.

regehr:
it doesn't look like you have any test cases for printing the external use info. best would be to test the full parse-print loop

test/Pass/nop-ext-use1.ll
```diff
@@ -0,0 +1,16 @@
...
+define i32 @func(i32 %x) {
+entry:
+  %a = add i32 %x, 100
+  %b = add i32 %a, 0
+  %c = add i32 %a, 100 ; External Use
+  %d = add i32 %b, 100
+  ; CHECK-NOT: ret i32 %b
+  ; CHECK: ret i32 %d
+  ret i32 %d
+}
```
regehr: I don't understand this test case, what is it trying to test?

include/souper/Inst/Inst.h
```diff
@@ -172,7 +173,7 @@ class ReplacementContext {
                 llvm::raw_ostream &Out, bool printNames);
   void printBlockPCs(const BlockPCs &BPCs,
                      llvm::raw_ostream &Out, bool printNames);
-  std::string printInst(Inst *I, llvm::raw_ostream &Out, bool printNames);
+  std::string printInst(Inst *I, llvm::raw_ostream &Out, bool printNames, Inst * OrigI = nullptr);
```
regehr: remove space after *

lib/Inst/Inst.cpp
```diff
@@ -173,6 +176,10 @@ std::string ReplacementContext::printInst(Inst *I, llvm::raw_ostream &Out,
           Out << " (signBits=" << I->NumSignBits << ")";
       }
       Out << OpsSS.str();
+
+      if (OrigI && OrigI->DepsWithExternalUses.find(I) != OrigI->DepsWithExternalUses.end())
+        Out << " (hasExternalUses)" ;
```
regehr: no space before the semicolon, please

lib/Parser/Parser.cpp
```diff
@@ -306,6 +307,8 @@ struct Parser {
   // set of blockpc(s) related to B. The map is used for error- and
   // type-checking.
   std::map<Block *, unsigned> BlockPCIdxMap;
+
```
regehr: do you want to add the blank line here?

regehr:
OK looks good except it needs a few tweaks and also some more test cases

zhengyangl:
The markExternalUses() function in extractor part still lacks of support for Insts with pc and blockpcs.

Eg. the following Souper IR:
```
%0:i1 = var ; p
%1:i1 = var ; q
%2:i1 = and %0, %1
pc %2 1:i1
%3:i1 = var ; r
%4:i1 = and %1, %3
pc %4 0:i1
pc %3 1:i1
%5:i1 = eq 1:i1, %0
cand %5 1:i1
```

The markExternalUses() function now only looks the dependencies of a value and %5 only depends on %0, thus the functions think the use count of %0 is 1, however, %0 is also used in the add Inst %2. %0 is marked as hasExternalUses in this case, which is wrong given a llvm source like 
```llvm
define void @f(i1 %p, i1 %q, i1 %r, i1 %s) {
entry:
  %pq = and i1 %p, %q
  br i1 %pq, label %bb1, label %u

bb1:
  %qr = and i1 %q, %r
  br i1 %qr, label %u, label %bb2

bb2:
  br i1 %r, label %bb3, label %u

bb3:
  br i1 %s, label %bb4, label %u

bb4:
  %cmp1 = icmp eq i1 %p, true
  %cmp2 = icmp eq i1 %s, true
  ret void

u:
  unreachable
}

```

regehr:
we definitely don't care about external uses on instructions that are only used by pc and blockpc. the pc and blockpc only add constraints for synthesis, they are never going to be synthesized

zhengyangl:
John: I've uploaded the rebased commit, please have a look. Also this commit contains some new test cases.

lib/Extractor/Candidates.cpp
```diff
@@ -215,6 +216,31 @@ Inst *ExprBuilder::buildGEP(Inst *Ptr, gep_type_iterator begin,
...
+  while(!Stack.empty()) {
+    Inst* T = Stack.back();
+    Stack.pop_back();
+    if (I != T) {
+      if (UsesCount.find(T) == UsesCount.end())
+        UsesCount[T] = 1;
+      else
+        UsesCount[T]++;
+    }
+    for (auto i: T->Ops) {
```
regehr: I think it's best to avoid using both "i" and "I" in the same function, and also please use initial capital letters for local vars, and also we usually put spaces on both sides of the ":" in this context

lib/Extractor/Candidates.cpp
```diff
@@ -215,6 +216,31 @@ Inst *ExprBuilder::buildGEP(Inst *Ptr, gep_type_iterator begin,
...
+      if (i->K != Inst::Const && i->K != Inst::Var
+          && i->K != Inst::UntypedConst && i->K != Inst::Phi)
+        Stack.push_back(i);
+    }
+  }
+  for (auto i : UsesCount)
```
regehr: also this i

lib/Inst/Inst.cpp
```diff
@@ -87,7 +87,10 @@ const std::vector<Inst *> &Inst::orderedOps() const {
 }
 
 std::string ReplacementContext::printInst(Inst *I, llvm::raw_ostream &Out,
-                                          bool printNames) {
+                                          bool printNames, Inst *OrigI) {
+  if (OrigI == nullptr)
+    return printInst(I, Out, printNames, I);
```
regehr: this seems a little weird...  OrigI has a default value of null, but if this function is called with OrigI == null then it calls itself again using a different value for OrigI? can we think of a better way to do this same thing?

also I'm coming to the opinion that we should mostly avoid default values for parameters. how messy would your patch become if you did this differently?

lib/Inst/Inst.cpp
```diff
@@ -173,6 +176,10 @@ std::string ReplacementContext::printInst(Inst *I, llvm::raw_ostream &Out,
           Out << " (signBits=" << I->NumSignBits << ")";
       }
       Out << OpsSS.str();
+
+      if (OrigI && OrigI->DepsWithExternalUses.find(I) != OrigI->DepsWithExternalUses.end())
```
regehr: so, for example, here you check OrigI against nullptr but didn't you already filter out the case where OrigI == nullptr? again, I think there must be a nicer way to say all this

lib/Inst/Inst.cpp
zhengyangl: The OrigI parameter is designed for holding a "global state" for the recursive function.

One stupid way of getting rid of the default parameter values is to specify the exact OrigI in each function call. This is a bad option since this brings in a redundant argument to every call of the printInst.

Another way is to create a helper function for printInst which does not has OrigI as parameter. Each call to the helper function will be redirected to the printInst, and pass I as OrigI to the printInst.

Finally, the recursive function can always be rewritten into a non-recursive one by adding a stack. 

John: Which way do you prefer?

lib/Parser/Parser.cpp
```diff
@@ -1175,7 +1179,26 @@ bool Parser::parseLine(std::string &ErrStr) {
 
         Ops.push_back(I);
 
-        if (CurTok.K != Token::Comma) break;
+        if (CurTok.K != Token::Comma) {
+          if (CurTok.K == Token::OpenParen) {
+            if (!consumeToken(ErrStr))
+              return false;
```
regehr: is this path tested?

zhengyangl:
@regehr Sure, will fix them soon.

lib/Parser/Parser.cpp
```diff
@@ -306,6 +307,7 @@ struct Parser {
   // set of blockpc(s) related to B. The map is used for error- and
   // type-checking.
   std::map<Block *, unsigned> BlockPCIdxMap;
+  std::unordered_set<Inst*> ExternalUsesSet;
```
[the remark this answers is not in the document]
zhengyangl: Oops.

regehr:
ping

regehr:
can you squash this, Zhengyang, and then I'll  merge?

zhengyangl:
@regehr squashed. Let's see how CI says.

regehr:
thanks, I'll merge once it passes

PR #324, opened 2018-04-23
Make KLEEBuilder a pure SMT-LIBv2 backend

rsas:
Until now, `KLEEBuilder` was creating constraints (e.g. UB) and generating SMT-LIBv2 queries at the same time. As a result, it prevented us to experiment with different SMT-LIBv2 backends or APIs (e.g. Z3) directly. Moreover, the `ExprBuilder` class was stateful preventing us to separate LHS and RHS properly.

This patch introduces a base class `ExprBuilder` that creates all constraints using `Inst`. `KLEEBuilder` is a derived class now that implements the `BuildQuery` function taking the `Cand` `Inst` returned from `GetCandidateExprForReplacement` and calling `get(Cand)` to generate the KLEE's `Expr` that is in turn printed to SMT-LIBv2.

All demanded bit, UB, and any other constraints were rewritten using `Inst`s. In addition, the code was refactored to separate LHS and RHS. `KLEEBuilder.cpp` is the only KLEE dependency enabling us to experiment with other SMT-LIBv2 back-ends or solver APIs.

As suggested by @regehr, UB exploitation is enabled by default. To this end, I removed two test cases. Other than that, the previous functionality was preserved (I tried).

@chenyang78 @jubitaneja @regehr @zhengyangl I'd appreciate any type of feedback, please start testing if you have time! Apologies for the size of the patch, but there was some work to do.


regehr:
Yaaayyyy!!!
Will look soon.

chenyang78:
Cool! I will look through it this week.

include/souper/Extractor/ExprBuilder.h
```diff
@@ -0,0 +1,114 @@
+// Copyright 2014 The Souper Authors. All rights reserved.
```
chenyang78: nit - since ExprBuilder.h is a newly-added file, we might want to make the copyright year start from 2018.

lib/Extractor/ExprBuilder.cpp
```diff
@@ -0,0 +1,945 @@
...
+    // possible that I->B has been added already.
+    if (Current->BlockConstraints.count(I->B))
+      return true;
+    std::vector<UBPath *> Tmp = { Current };
+    // Create copies of the current path
+    for (unsigned J = 1; J < Ops.size(); ++J) {
+      UBPath *New = new UBPath;
+      *New = *Current;
+      New->BlockConstraints[I->B] = J;
+      Paths.push_back(std::move(std::unique_ptr<UBPath>(New)));
```
chenyang78: ```std::vector``` is move-aware. We can simply use the follow form:

    Paths.push_back(std::unique_ptr<UBPath>(New));

I notice there are several other similar places where we can make the simplification.

test/Tool/too-many-phis.opt
```diff
@@ -1,1881 +0,0 @@
...
-%1866:i64 = lshr %1864, 47:i64
-%1867:i64 = xor %1865, %1866
-%1868:i64 = mul 11376068507788127593:i64, %1867
-%1869:i64 = lshr %1868, 47:i64
-%1870:i64 = xor %1868, %1869
-%1871:i64 = mul 11376068507788127593:i64, %1870
-%1872:i64 = xor %1855, %1871
-%1873:i64 = phi %0, %1854, %1872
-%1874:i64 = var
-cand %1873 %1874
```
chenyang78: Out of curiosity, what's the reason for removing this test?

rsas: The test was checking for "Value too large", but now it's not too large anymore returning "Invalid". Thus, it defeats the purpose of the test, no?

rsas:
Thanks for the initial comments @chenyang78, I updated the branch accordingly.

rsas:
Thanks, maybe you find a more elegant way to instantiate the derived class with the ``InstContext``. The rest is a pure rewrite.

chenyang78:
I will try :) I think things will become more concrete when we add another ```ExprBuilder``` such as ```Z3Builder```.

rsas:
I plan to implement the `Z3Builder` next. It will generate `SMT-LIBv2` similar to `KLEEBuilder`. Subsequently, we can learn from that effort to switch to a library based API of `Z3`.

zhengyangl:
This commit looks good to me! Thanks.

rsas:
@chenyang78 did you have chance to look into this? ;-)

@regehr could you please run a couple of longer (constant) synthesis tests (e.g., what you did for wasm) using this branch? I want to get more confidence.

If possible, it would be good to get this merged soon to move on with other efforts.

rsas:
Thank you for your time @chenyang78!

I'll merge this once @regehr runs it on one of his test setups.

regehr:
It's annoying that travis is timing out. I'll turn off some of our tests or something in the trunk.

regehr:
I get all tests passing in a Release build of Souper, but am seeing some failures in Debug mode:
https://gist.github.com/regehr/267b617546590dad6dbb9da3c95a8dda

regehr:
Ok, the debug-mode tests pass under Valgrind as well, if given huge stacks, so this really isn't any kind of logic error.

regehr:
LGTM as long as we know how to fix the stack usage issues later.

regehr:
yayyy!!

PR #327, opened 2018-05-11
WIP: Use LHS components as synthesis components

rsas:
This WIP PR adds a parameter to instruction synthesis corresponding to a vector of instructions from the LHS that can be used as components in synthesis. The instructions are initialized as ordinary library components and are reconstructed if a candidate is found.

The first test case ``test/Infer/four-adds.opt`` shows that the synthesis engine can reconstruct the original LHS out of two sub-components (two ``add``s). Feel free to add more reasonable tests.

Usage: LHS components are always initialized. You might want to specify additional components/constants manually using ``-souper-synthesis-comps=...``. Otherwise, the whole component library will be initialized as well (slow).

You will probably encounter issues, please provide me test cases, thanks!


regehr:
why does stupid googlebot keep complaining

regehr:
this is awesome! I have a long drive home tomorrow and will try this out / write tests when I get there

rsas:
@regehr The googlebot complaint is valid, I initially started with a merge from your branch, then squashed leaving you as a committer. I change it to me, looks good now.

PR #328, opened 2018-05-14
turn off valgrind in CI testing

regehr:
argh, we're bumping into Travis's 50 minute limit, this makes testing quite a bit faster


rsas:
LGTM, please merge.

PR #329, opened 2018-05-17
this PR makes two minor changes:

regehr:
first, we had two functions called getInstCopy but they didn't do the
same thing, so I renamed one of them to be replaceVars, which seems
reasonable, but Raimondas should double check that this is OK

second, remove an unintentional exponential time in the other
getInstCopy, ugh


rsas:
Merging to put my fix in a minute

PR #331, opened 2018-05-19
new test cases intended to stress-test the new asymmetical UB model in

regehr:
souper, which is mostly correct, thanks to Raimondas's recent work.
however, there are a few xfails related to select that we'll need to
fix before we can turn on Souper for real.


regehr:
@rsas I think we should merge this. all of the UB stress-tests that I added are passed by your new code except for 4 of them that involve select, as we've already discussed.

regehr:
ok I'm  just gonna merge this-- no real code changes are included

PR #332, opened 2018-06-03
Build script improvements

pranavk:
In the earlier setup, the ./build_deps.sh was cloning and building LLVM in third_party/ directory unconditionally. This can be {time,space}-consuming when you already have a LLVM build in your machine. Let's Introduce some flags to the script to mention the custom LLVM build. This solves the problem for me (and hopefully will help others as well).

Also, no need to run cmake again in the build script, if it was already run before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 find_program(LLVM_CONFIG_EXECUTABLE HINTS ${CMAKE_SOURCE_DIR}/third_party/llvm/${CMAKE_BUILD_TYPE}/bin
+                                    HINTS ${CMAKE_LLVM_INSTALL_DIR}/bin
                                     NAMES llvm-config DOC "llvm-config executable")
 if(LLVM_CONFIG_EXECUTABLE STREQUAL "LLVM_CONFIG_EXECUTABLE-NOTFOUND")
   message(FATAL_ERROR "llvm-config could not be found!")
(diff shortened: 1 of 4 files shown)


regehr:
there are a couple potential problems with this. one is that any given version of souper is likely to work only with one version of LLVM, so any code referring to an external version of LLVM requires a version check, it should exit with an error if the version isn't the right one. the other problem is that we actually patch LLVM to disable some optimizations that conflict with souper, and of course the external version isn't going to have those patches. so I think that for now, we just want to keep building a custom LLVM.

pranavk:
The patch shouldn't prevent the possibility of building a separate LLVM as well. If you give no arguments to the script, it will be just as it used to be.

My only intention was to share LLVM builds somehow or have flexibility to put LLVM build on different, bigger partition.

> one is that any given version of souper is likely to work only with one version of LLVM, 

Hmm, I had LLVM built from trunk on my local already and if I am not wrong, souper is working smoothly with it :) (apart from 6.0 tag that I saw in the build script)

> the other problem is that we actually patch LLVM to disable some optimizations that conflict with souper, and of course the external version isn't going to have those patches. 

Yeah, I saw the disabling of select -> logic optimizations in the build script. The build script took care of it for me and patched my existing LLVM before doing an incremental build which was cheap. I can revert the patch on the same build tree and do an incremental build (cheap) before using it for other purposes. So I end up saving some time and space :)

regehr:
> Hmm, I had LLVM built from trunk on my local already and if I am not wrong, souper is working smoothly with it :) 

Yes but that is not the expected case. We want to force the specific version.

> The build script took care of it for me and patched my existing LLVM before doing an incremental build which was cheap.

We very definitely do not want to randomly patch an LLVM that is in someone's filesystem, that is very unfriendly behavior.

PR #337, opened 2018-08-29
Pass tests to verify that synthesis is working

manasij7479:
Test pull request.
If everything works, will add more tests.


manasij7479:
I signed it!

regehr:
these look good.
in general a brief comment to explain the test case is useful, but these probably are so simple that they don't need it
please add a few components to the list that you give synthesis, otherwise you are making things too easy for it.

regehr:
LGTM

PR #344, opened 2018-09-16
Instructions with external uses cost 0

manasij7479:
This is a simple fix to incentivize Souper to not remove instructions with external uses.
This is accomplished by making those instructions in the LHS cost 0. The rationale behind this is that the other 'users' of those instructions will generate it anyway.


regehr:
intuitively, the cost of the LHS should not be changed by external uses. these should only reduce the benefit in transforming a particular LHS to a RHS.

clearly this patch does not implement what I said above. if it produces equivalent behavior, then I think an argument to this effect should be provided.

also, what if some part of souper actually wants to know how much a LHS costs and now you've made souper tell lies by leaving out the cost of anything with external uses? have you made absolutely sure that nowhere in souper does this?

manasij7479:
I think it produces equivalent behavior, might need some help coming up with an argument though.

Would a middle ground solution (extra flag for cost function to control this) work?  
  
Note that 0 is only a provisional value which happens to work here, we should be abke to come up with with a more sophisticated calculation for sequences of optimizations which also remove the extra uses.

regehr:
I think it's worth doing the up-front work to do this right, the zero cost seems like a hack that might backfire

regehr:
can you come up with a better solution?

lib/Inst/Inst.cpp
```diff
@@ -667,18 +667,27 @@ int Inst::getCost(Inst::Kind K) {
...
 
-int souper::cost(Inst *I) {
+int souper::cost(Inst *I, bool IgnoreDepsWithExternalUses) {
   std::set<Inst *> Visited;
-  return costHelper(I, Visited);
+  return costHelper(I, I, Visited, IgnoreDepsWithExternalUses);
+}
+
+int souper::benefit(Inst *LHS, Inst *RHS) {
+  return cost(LHS, /*IgnoreDepsWithExternalUses*/ true) - cost(RHS);
```
regehr: let's go with the exact LLVM idiom here:

/*IgnoreDepsWithExternalUses=*/true

regehr:
looks good except for one comment, please squash

regehr:
did you make sure that no other calls to cost() need to be turned into calls to benefit()?

manasij7479:
(afaik) There are no other explicit calls to cost which are relevant here.

PR #346, opened 2018-09-19
update to LLVM 7, also update some random components of our Docker

regehr:
setup, this is for CI and should not affect any non-souper-developers

this contains an old LLVM patch for de-fanging InstCombine, Zhengyang
will patch it up soon


regehr:
merging as obvious

PR #347, opened 2018-09-19
Pulls the instcombine patch out of build-deps.sh

zhengyangl:


regehr:
ok, this patch is updated for LLVM 7, right?
do you want to add the option to turn off instcombine or is that a different pull request?

zhengyangl:
@regehr Yes that will be in a different PR

regehr:
thanks!

PR #349, opened 2018-09-20
Added commandline option for turning off instruction combiner

zhengyangl:
With this patch one may call llvm opt with "-disable-instruction-combiner", to disable the instruction combiner during the optimization. For clang/clang++, call with '-m -disable-instruction-combiner'.

Also the sclang is updated. Setting environment variable "LLVM_DISABLE_INSTRUCTION_COMBINER" will also turn off the instcombine.

### DIFF
--- a/build_deps.sh
+++ b/build_deps.sh
@@ -40,6 +40,9 @@ svn co https://llvm.org/svn/llvm-project/cfe/${llvm_branch} $llvmdir/tools/clang
 svn co https://llvm.org/svn/llvm-project/compiler-rt/${llvm_branch} $llvmdir/projects/compiler-rt
 # Disable the broken select -> logic optimizations
 patch $llvmdir/lib/Transforms/InstCombine/InstCombineSelect.cpp < patches/disable-instcombine-select-to-logic.patch
+# Apply instcombine switch patch
+patch -d $llvmdir -p0 -i $(pwd)/patches/enable-instcombine-switch.patch
+
 mkdir -p $llvm_builddir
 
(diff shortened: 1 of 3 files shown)


build_deps.sh
```diff
@@ -14,10 +14,21 @@
...
+        --enable-instcombine-switch)  build_llvm_with_instcombine_switch=true;;
+        *)  echo "Usage: ./build_deps.sh [--enable-instcombine-switch]" && exit 0;;
+    esac
+    shift
+done
+
+#if [ -d "third_party" ]; then
+#  echo "Directory third_party exists, remove this directory before running build_deps.sh."
+#  exit 1;
+#fi
```
regehr: you left this check of the third_party directory commented out!

but really, I think it adds complexity to make this patch apply conditionally, I would recommend that we apply it unconditionally

patches/enable-instcombine-switch.patch
```diff
@@ -0,0 +1,367 @@
...
++    addInstructionCombiningPass(PM);
++  }
+   addExtensionsToPM(EP_Peephole, PM);
+ 
+   // Inline small functions
+@@ -825,7 +840,8 @@
+   PM.add(createArgumentPromotionPass());
+ 
+   // The IPO passes may leave cruft around.  Clean up after them.
+-  addInstructionCombiningPass(PM);
```
regehr: this approach is fine if you want to take it, but you can have the same effect with less code if you just (conditionally) return early from the instcombine code. in that case the pass runs but it doesn't do anything.

zhengyangl:
Squashed, please review and merge.

PR #352, opened 2018-09-27
Bootstrap LLVM with Ubuntu pre-build clang, solve the go cache issue …

zhengyangl:
…and use bootstrapped clang only for Souper test builds.


zhengyangl:
I am still testing this PR.

PR #353, opened 2018-09-27
build libcxx and libcxxabi by default in build_deps.sh

zhengyangl:
Souper build crashes when building with the third_party LLVM on Mac OS X. A LLVM build with libcxx and libcxxabi solves the crash. This PR forces building libcxx.


regehr:
thanks!

PR #354, opened 2018-09-28
IR generation support  for select instruction

zhengyangl:


lib/Pass/Pass.cpp
```diff
@@ -311,6 +311,20 @@ struct SouperPass : public ModulePass {
         break;
       }
     }
+    case 3:{
+      Value *V1 = getValue(I->Ops[1], ReplacedInst, EBC, DT,
+                           ReplacedValues, Builder, M);
+      Value *V2 = getValue(I->Ops[2], ReplacedInst, EBC, DT,
+                           ReplacedValues, Builder, M);
```
regehr: did we call getValue on V0 somewhere earlier?

PR #355, opened 2018-09-28
Crashes with report when call getValue() on operands returns null

zhengyangl:


regehr:
I think it might be better to add a single check around like 421 of this file

zhengyangl:
@regehr John, are you saying putting a fatal check for NewVal in line 421, instead of just reporting and continue?

regehr:
I'm just saying that there's already code in the pass (in runOnFunction) to check for failure of getValue() and print something, if we report the fatal error there, it's only one place to change in the code instead of multiple places like this patch currently has

regehr:
can you replace this patch with a very small one that hooks into the place where getValue() is originally called, rather than changing getValue() itself?

regehr:
hey @manasij7479 coud you look into the failing tests here and see if they stem from pass bugs?

PR #357, opened 2018-09-29
just some random cleanup, no bug fixes or anything

regehr:
@zhengyangl please look this over

this does not fix the bug I mentioned earlier, but also the bug isn't in findCands, can you look into it?


regehr:
thanks

PR #360, opened 2018-10-02
Generate new input guesses for each constant synthesis try

zhengyangl:
This patch reduces the number of constant synthesis tries required for the following program from 175 to 82

```
%in:i16 = var
%x = add %in, 25245
%a = lshr %x, 1
%b = shl %a, 2
%c = lshr %b, 1
infer %c
```


regehr:
LGTM thanks!

PR #361, opened 2018-10-03
Towards Handling Big LHS

zhengyang92:
Always feed getGuesses() with Vars in LHS: the original findCands() for generating synthesis candidates are called with MaxCand = 15, this may miss the input variables declared at the beginning of a large program. This patch fixes this.

Add a switch for big query: In some cases the huge big query may abort the Souper

Enable the ctpop test case.

### DIFF
--- a/lib/Infer/ExhaustiveSynthesis.cpp
+++ b/lib/Infer/ExhaustiveSynthesis.cpp
@@ -377,11 +377,9 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
                                 const std::vector<InstMapping> &PCs,
                                 Inst *LHS, Inst *&RHS,
                                 InstContext &IC, unsigned Timeout) {
-  std::vector<Inst *> Vars;
-  findVars(LHS, Vars);
 
-  std::vector<Inst *> Inputs(Vars);
-  findCands(LHS, Inputs, /*WidthMustMatch=*/false, /*FilterVars=*/true, MaxLHSCands);
+  std::vector<Inst *> Inputs;
+  findCands(LHS, Inputs, /*WidthMustMatch=*/false, /*FilterVars=*/false, MaxLHSCands);
 
   if (DebugLevel > 1)
     llvm::errs() << "got " << Inputs.size() << " candidates from LHS\n";
@@ -452,6 +450,10 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
   // find the valid one
   int Unsat = 0;
   int GuessIndex = -1;
+
+  std::vector<Inst *> Vars;
+  findVars(LHS, Vars);
+
   for (auto I : Guesses) {
     GuessIndex++;
     if (DebugLevel > 2) {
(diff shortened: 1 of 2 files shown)


zhengyangl:
Solving the conflicts.

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -374,8 +377,12 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
                                 const std::vector<InstMapping> &PCs,
                                 Inst *LHS, Inst *&RHS,
                                 InstContext &IC, unsigned Timeout) {
-  std::vector<Inst *> Inputs;
-  findCands(LHS, Inputs, /*WidthMustMatch=*/false, /*FilterVars=*/false, MaxLHSCands);
+  std::vector<Inst *> Vars;
+  findVars(LHS, Vars);
+
```
regehr: we should talk about this part. why do you think it's a good idea? do you have data supporting that it does a better job?

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -32,6 +32,9 @@ namespace {
     "The larger the number is, the more fine-grained debug "
     "information will be printed"),
     cl::init(0));
+  static cl::opt<bool> NoBigQuery("souper-exhaustive-synthesis-no-big-query",
```
regehr: let's get some data about this also -- maybe run synthesis with and without it and see how it changes the CPU time used

lib/Infer/ExhaustiveSynthesis.cpp
zhengyangl: I did this because the current findCands() stops when it reaches MaxLHSCands(which is set to 15 by default). However, this might miss the input variable declared at the top of program. Due to this limitation, the synthesizer cannot gives correct ctpop result since the input %x is not harvested as a component of synthesis.

Exhaustive synthesis has a call findVars() after doing the big query, hoisting out this call does not have much impact to the compilation performance, compared to the potential benefit of getting better synthesis results. 

This approach is stupid I admit, it basically runs the traversal on the LHS twice. I can write a better version of the the findCands which runs once.

zhengyangl: This code basically guarantees all the program inputs are listed as a components in the synthesis.

zhengyangl: Okay this makes sense. But instead of stopping at a hard bounded limit when doing traversal, can we traverse all the tree and hard bounds the result by the descend order of the "benefit" of each candidates? There is already a "benefit" tagging algorithm there implemented in findVars.

zhengyangl: Currently the deeper the traverse goes, the higher benefit the candidate gets. We may further customize this benefit function to fit our needs. Say some algorithm can tell some kind of candidates are more likely to be used as a component in synthesis, then we can increase the "benefit" of this kind of candidates.

regehr:
I think you should separate the bigquery part of this PR from the changes to how LHSs are chosen

lib/Infer/ExhaustiveSynthesis.cpp
zhengyangl: Sure I will split this pr and collect data

PR #362, opened 2018-10-03
Minor fixes

zhengyangl:
Put isCmp() and isShift() in Inst.cpp
Do not reinitialize Vars in different guesses.

### DIFF
--- a/lib/Infer/ExhaustiveSynthesis.cpp
+++ b/lib/Infer/ExhaustiveSynthesis.cpp
@@ -68,17 +68,6 @@ namespace {
 // experiment with synthesizing at reduced bitwidth, then expanding the result
 // aggressively avoid calling into the solver
 
-// TODO move this to Inst
-const bool isCmp(Inst::Kind K) {
-  return K == Inst::Eq || K == Inst::Ne || K == Inst::Ult ||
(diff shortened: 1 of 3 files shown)


lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -445,6 +434,10 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
   // find the valid one
   int Unsat = 0;
   int GuessIndex = -1;
+
+  std::vector<Inst *> Vars;
+  findVars(LHS, Vars);
```
regehr: what's the reason for moving this code earlier?

zhengyangl:
This code is inside the again: label and evaluated on each constant synthesis

lib/Infer/ExhaustiveSynthesis.cpp
zhengyangl: This code is inside the again: label and evaluated on each constant guess try. I am only moving this before the again: label

regehr: aha, great catch, thanks!

regehr:
LGTM

PR #367, opened 2018-10-10
Multiple RHS Support

zhengyangl:
This patch changes the type of the RHS argument of ExhaustiveSynthesis::synthesize() and Solver::infer() from a Inst pointer to a vector of Inst pointers.

Exhaustive Synthesis now takes a constant "MaxRHS". This constant is set to 1 for performance. Exhaustive Synthesis now generates at most MaxRHS number of RHSs. The original Souper test suite on my machine runs 282.54s while with MaxRHS set to 3, the test suite runs 454.83s.

Call sites of infer() function takes the result vector of RHS and extract the first RHS, this preserves the behaviors of these call sites.

### DIFF
--- a/lib/Extractor/Solver.cpp
+++ b/lib/Extractor/Solver.cpp
@@ -75,7 +75,8 @@ public:
 
   std::error_code infer(const BlockPCs &BPCs,
                         const std::vector<InstMapping> &PCs,
-                        Inst *LHS, Inst *&RHS, InstContext &IC) override {
+                        Inst *LHS, std::vector<Inst *> &RHSs,
+                        InstContext &IC) override {
     std::error_code EC;
 
     /*
@@ -103,7 +104,7 @@ public:
         if (EC)
           return EC;
         if (!IsSat) {
-          RHS = I;
+          RHSs.emplace_back(I);
           return EC;
         }
       }
@@ -142,7 +143,7 @@ public:
         if (EC)
           return EC;
         if (!IsSat) {
-          RHS = Const;
+          RHSs.emplace_back(Const);
           return EC;
         }
       }
@@ -188,7 +189,7 @@ public:
           if (EC)
             return EC;
           if (!SmallQueryIsSat) {
-            RHS = I;
+            RHSs.emplace_back(I);
             break;
           }
         }
@@ -214,18 +215,21 @@ public:
     if(SMTSolver->supportsModels()) {
       if (EnableExhaustiveSynthesis) {
         ExhaustiveSynthesis ES;
-        EC = ES.synthesize(SMTSolver.get(), BPCs, PCs, LHS, RHS, IC, Timeout);
-        if (EC || RHS)
+        EC = ES.synthesize(SMTSolver.get(), BPCs, PCs, LHS, RHSs,
+                           IC, Timeout);
+        if (EC || !RHSs.empty())
           return EC;
       } else if (InferInsts) {
         InstSynthesis IS;
+        Inst *RHS;
         EC = IS.synthesize(SMTSolver.get(), BPCs, PCs, LHS, RHS, IC, Timeout);
+        RHSs.emplace_back(RHS);
         if (EC || RHS)
           return EC;
       }
     }
 
-    RHS = 0;
+    RHSs.clear();
     return EC;
   }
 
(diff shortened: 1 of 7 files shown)


zhengyangl:
Note: no support for multiple RHS dump in souper-check and redis cache in this patch.

lib/Extractor/Solver.cpp
```diff
@@ -103,7 +104,7 @@ class BaseSolver : public Solver {
         if (EC)
           return EC;
         if (!IsSat) {
-          RHS = I;
+          RHSs.emplace_back(I);
```
regehr: do you want to return here, or try to find more RHSs?

lib/Extractor/Solver.cpp
```diff
@@ -142,7 +143,7 @@ class BaseSolver : public Solver {
         if (EC)
           return EC;
         if (!IsSat) {
-          RHS = Const;
+          RHSs.emplace_back(Const);
```
regehr: same

lib/Extractor/Solver.cpp
```diff
@@ -214,18 +215,21 @@ class BaseSolver : public Solver {
     if(SMTSolver->supportsModels()) {
       if (EnableExhaustiveSynthesis) {
         ExhaustiveSynthesis ES;
-        EC = ES.synthesize(SMTSolver.get(), BPCs, PCs, LHS, RHS, IC, Timeout);
-        if (EC || RHS)
+        EC = ES.synthesize(SMTSolver.get(), BPCs, PCs, LHS, RHSs,
```
regehr: did you intend to change this line? was it too long?

zhengyangl: RHS is changed to RHSs, and it is too long so I wrapped.

lib/Extractor/Solver.cpp
```diff
@@ -281,16 +285,17 @@ class MemCachingSolver : public Solver {
...
     const auto &ent = InferCache.find(Repl);
     if (ent == InferCache.end()) {
       ++MemMissesInfer;
-      std::error_code EC = UnderlyingSolver->infer(BPCs, PCs, LHS, RHS, IC);
+      std::error_code EC = UnderlyingSolver->infer(BPCs, PCs, LHS, RHSs, IC);
       std::string RHSStr;
-      if (!EC && RHS) {
-        RHSStr = GetReplacementRHSString(RHS, Context);
+      if (!EC && !RHSs.empty()) {
+        RHSStr = GetReplacementRHSString(RHSs.front(), Context);
```
regehr: seems weird to take the first element here, is this a placeholder for something different?

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -438,6 +439,7 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
   std::vector<Inst *> Vars;
   findVars(LHS, Vars);
 
+  unsigned ValidRHS = 0;
```
regehr: why do you want a separate variable for this? doesn't it work to inspect the length of the vector?

lib/Extractor/Solver.cpp
zhengyangl: These are both constant synthesis, why we wants to collect multiple RHSs if a constant fit the LHS

lib/Extractor/Solver.cpp
zhengyangl: This PR does not support cache dumping yet, so only print the first RHS in the list. I'll add support for cache dumping soon.

lib/Extractor/Solver.cpp
regehr: well, it depends on why we want multiple RHSs. for manasij's purposes a constant is always good, but for building a cost model I'd like as many alternatives as possible

lib/Extractor/Solver.cpp
regehr: I'd say don't do the caching part of this yet, until we're sure we want it

lib/Extractor/Solver.cpp
zhengyangl: ping me if you need this.

lib/Extractor/Solver.cpp
zhengyangl: Agreed.

manasij7479: Sounds good, should be possible to prototype without caching.

PR #369, opened 2018-10-11
Shortcut first query if there is no constants

zhengyangl:


regehr:
can you explain this? I thought we would jump all the way to line 555

zhengyangl:
Yes this does jump all the way to line 555.

zhengyangl:
Besides this shortcut, I also did some reorder to avoid unnecessary initialization

regehr:
thanks will look again after getting kids to bed

regehr:
ok I see now, LGTM

PR #370, opened 2018-10-11
Remove unused assembler step in tests

manasij7479:
Removes the following runline from 119  tests.
; RUN: %llvm-as -o %t %s
%t was never used, opt was being invoked with %s
'make check' with 16 cores takes about 5 seconds less to complete.


manasij7479:
CI tests failed, doesn't seem like it was a good idea.

PR #372, opened 2018-10-13
Multiple Instructions Exhaustive Synthesis

zhengyangl:


lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -32,6 +32,12 @@ namespace {
     "The larger the number is, the more fine-grained debug "
     "information will be printed"),
     cl::init(0));
+  static cl::opt<unsigned> NumInstructions("souper-exhaustive-synthesis-num-instructions",
+    cl::desc("Number of instruction of RHS (default=1). "),
```
regehr: please change to "Maximum number of instructions to synthesize (default=1)."

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -32,6 +32,12 @@ namespace {
...
+    cl::init(1));
+  static cl::opt<bool> DisableBigQuery("souper-exhaustive-synthesis-disable-big-query",
```
regehr: let's flip this around: have a flag (default=false) for enabling big query, not for disabling it

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -130,7 +136,7 @@ void findVars(Inst *Root, std::vector<Inst *> &Vars) {
 void getGuesses(std::vector<Inst *> &Guesses,
                 std::vector<Inst *> &Inputs,
                 int Width, int LHSCost,
-                InstContext &IC) {
+                InstContext &IC, int depth) {
```
regehr: capitalize first letter of variables used as parameters

regehr:
I don't understand how there's an argument for number of instructions but then it seems to be used as a depth, these are different things.
if at all possible, let's actually have a number of instructions for real, not just a depth

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -148,6 +161,8 @@ void getGuesses(std::vector<Inst *> &Guesses,
   if (Width > 1) {
     for (auto K : Unary) {
       for (auto I : Inputs) {
+        int CostK_I = souper::cost(I, true) + Inst::getCost(K);
+        if (CostK_I > LHSCost) continue;
```
regehr: please put continue on a separate line (braces optional)

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -166,12 +181,16 @@ void getGuesses(std::vector<Inst *> &Guesses,
 
   Inputs.push_back(IC.getReserved());
 
-  // Binary and Unary operators
   for (auto K : Kinds) {
     for (auto I = Inputs.begin(); I != Inputs.end(); ++I) {
+      int CostK_I = souper::cost((*I), true) + Inst::getCost(K);
+      if  (CostK_I > LHSCost) continue;
```
regehr: same

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -166,12 +181,16 @@ void getGuesses(std::vector<Inst *> &Guesses,
...
+
       // PRUNE: don't try commutative operators both ways
       auto Start = Inst::isCommutative(K) ? I : Inputs.begin();
       for (auto J = Start; J != Inputs.end(); ++J) {
+        int GuessCost = souper::cost((*J), true) + CostK_I;
+        if  (GuessCost > LHSCost) continue;
```
regehr: same, plus a few more below

regehr:
I still think MaxNumberOfInstructions is wrong! you are using it as a depth,not a max number of instructions. is that correct?

zhengyangl:
@regehr Yes, I mistakenly used it as a depth. You're correct, I'll do a bit refactor and get this fixed. Also I'll modify the code to generate the code from root to leaves (currently this implementation generates leaves first).

regehr:
If you want, just leave it as a depth for now (but rename it).

regehr:
what's status?

zhengyangl:
@regehr I did some refactor. Now the guess generation is updated to generate the root first, instead of the previous implementation which generates the leaves first. I am doing testing and will update this pr soon.

regehr:
ok good, I think we need root-first to support early pruning

regehr:
I just noticed that SOUPER_EXHAUSTIVE_SYNTHESIS isn't in the sclang help message, can you add that as part of this PR?

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -109,69 +115,68 @@ void addGuess(Inst *RHS, int MaxCost, std::vector<Inst *> &Guesses,
...
   };
   if (Width > 1) {
     for (auto K : Unary) {
-      for (auto I : Inputs) {
-        for (auto V : matchWidth(I, Width, IC)) {
-          auto N = IC.getInst(K, Width, { V });
-          addGuess(N, LHSCost, Guesses, TooExpensive);
+      for (auto I : Comps) {
+        if (I->K == Inst::ReservedInst && I->Width == 0) {
+          std::vector<int> UnaryArgWidths = {16,32};
```
regehr: what's going on here? why 16 and 32?

lib/Inst/Inst.cpp
```diff
@@ -712,6 +725,33 @@ int souper::cost(Inst *I, bool IgnoreDepsWithExternalUses) {
   return costHelper(I, I, Visited, IgnoreDepsWithExternalUses);
 }
 
+
+static int countHelper(Inst *I, std::set<Inst *> &Visited) {
+  if (!Visited.insert(I).second)
+    return 0;
+
+  int Count;
+  switch (I->K) {
```
regehr: I think this would work better as an if instead of a switch

lib/Inst/Inst.cpp
```diff
@@ -850,7 +947,8 @@ Inst *souper::getInstCopy(Inst *I, InstContext &IC,
       }
     }
     if (!Copy) {
-      if (CloneVars && I->Name.find("reserved_") == std::string::npos)
+      if (CloneVars &&
+          I->Name.find("reservedconst_") == std::string::npos)
```
regehr: Since this magic string is used in multiple places, please factor it out into a constant string and then refer to that. Same for "reservedinst_"

lib/Inst/Inst.cpp
```diff
@@ -876,6 +974,27 @@ Inst *souper::getInstCopy(Inst *I, InstContext &IC,
...
+    Ops.push_back(NewOp);
+  }
+
+  Inst *Copy = 0;
+  if (I == EmptyInst) {
+    Copy = NewInst;
+  } else if (I->K == Inst::Var) {
+    Copy = I;
+  } else
+    Copy = IC.getInst(I->K, I->Width, Ops);
```
regehr: if the "if" part has curly braces, then use curly braces for all subsequent parts of the conditional

test/Infer/syn-nand.ll
```diff
@@ -0,0 +1,29 @@
+; REQUIRES: solver
+; RUN: %souper-check -infer-rhs -souper-exhaustive-synthesis -souper-exhaustive-synthesis-num-instructions=2 %solver %s > %t1
+; RUN: %FileCheck %s < %t1
+; CHECK: RHS inferred successfully
```
regehr: please check for the actual instructions you want to see, not just for successful synthesis

regehr:
ok generally this looks very good!
please fix up a few comments I just made
also, can you add a couple more test cases? try to stress things like ext/trunc, multiple constants, and anything else that you found hard to get right

lib/Infer/ExhaustiveSynthesis.cpp
zhengyangl: This needs discuss:
For unary operation on a instruction, like 
C = bswap %reservedinst.

The width of reservedinst is not ready yet, I assigned 16 and 32 temporally.

regehr: this is fine for now, but:
- also support 64
- add a comment explaining this situation (and anything else similar in this patch)
- write a test case for this!!!

zhengyangl: Sure

regehr:
it looks like this isn't ready for review yet?

zhengyangl:
No I am still looking into a bug triggered by a ctlz guess.

CMakeLists.txt
```diff
@@ -286,6 +286,7 @@ endif()
 
 set(TEST_SOLVER "" CACHE STRING "Solver command line argument to use in tests, e.g. -stp-path=/path/to/stp")
 set(TEST_SYNTHESIS "" CACHE STRING "Enable additional, computationally intensive synthesis tests")
+set(TEST_LONG_DURATION_SYNTHESIS "" CACHE STRING "Enable long duration (> 10 min) synthesis")
```
regehr: please update to "synthesis tests"

lib/Inst/Inst.cpp
```diff
@@ -821,6 +856,63 @@ void souper::findCands(Inst *Root, std::vector<Inst *> &Guesses,
   }
 }
 
+/* TODO call findCands instead */
```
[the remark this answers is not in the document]
zhengyangl: This function is used to locate all the variables in a souper inst, and it is used in input specialization, to figure out which value needs specialization and to gets all the variables in path condition. It is original located in ExhaustiveSynthesis.cpp, I moved it out.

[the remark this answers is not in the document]
regehr: but I mean why call it instead of findCands, if findCands is the right thing to do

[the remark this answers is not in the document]
zhengyangl: findCands is implemented to find all the synthesis components of a LHS, and findVars finds all Inst::Vars. Though it has almost same code structure, but basically it does different works. Like findCands has MaxCands bound, but for findVars, we want to find all possible Vars in a Inst.

regehr:
LGTM, thanks!

manasij7479:
Awesome!  
Thanks for getting this done.  
I will get started with the pruning work.

PR #375, opened 2018-11-06
Add script to measure infer success rate

manasij7479:


test_infer.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+#Example invocation:
+#./test_infer.sh './build/souper-check --z3-path="/usr/bin/z3" -reinfer-rhs -souper-infer-inst -souper-exhaustive-synthesis' test/Solver/alive
+
+mkdir -p /tmp/results
+rm -rf /tmp/results/*
```
regehr: I don't like the hardcoded directory name and it's super dangerous to rm -rf anything on a machine from a script, don't do that please

manasij7479: Fixed.
Is rm -r okay?

regehr: it would be better to just ask the user for a fresh directory into which to put results, let's try to avoid hard-coded paths like this (or just put results into the current dir maybe)

zhengyangl: Why not use on the mktemp

manasij7479: Good idea, thanks.

regehr:
Looks great, thanks!

PR #377, opened 2018-11-14
Harvesting uses

zhengyangl:
Note: this patch is not fully working for now, the replacement part is not implemented yet.

### DIFF
--- a/include/souper/Extractor/Candidates.h
+++ b/include/souper/Extractor/Candidates.h
@@ -38,11 +38,9 @@ class Value;
 
 namespace souper {
 
-enum class CandidateType { HarvestedFromDef, HarvestedFromUse };
-
 struct CandidateReplacement {
-  CandidateReplacement(llvm::Instruction *Origin, InstMapping Mapping, CandidateType T)
-  : Origin(Origin), Mapping(Mapping), Type(T) {}
+  CandidateReplacement(llvm::Instruction *Origin, InstMapping Mapping)
+  : Origin(Origin), Mapping(Mapping) {}
 
   /// The instruction from which the candidate was derived.
   llvm::Instruction *Origin;
@@ -57,8 +55,6 @@ struct CandidateReplacement {
   /// A BlockPC has the same semantics as a PC, except that the PC only applies
   /// if the given predecessor of the given block is chosen.
   BlockPCs BPCs;
-
-  CandidateType Type;
 
   void printFunction(llvm::raw_ostream &Out) const;
   void printLHS(llvm::raw_ostream &Out, ReplacementContext &Context,
--- a/include/souper/Extractor/Candidates.h
+++ b/include/souper/Extractor/Candidates.h
@@ -38,9 +38,11 @@ class Value;
 
 namespace souper {
 
+enum class CandidateType { HarvestedFromDef, HarvestedFromUse };
+
 struct CandidateReplacement {
-  CandidateReplacement(llvm::Instruction *Origin, InstMapping Mapping)
-  : Origin(Origin), Mapping(Mapping) {}
+  CandidateReplacement(llvm::Instruction *Origin, InstMapping Mapping, CandidateType T)
+  : Origin(Origin), Mapping(Mapping), Type(T) {}
 
   /// The instruction from which the candidate was derived.
   llvm::Instruction *Origin;
@@ -56,6 +58,8 @@ struct CandidateReplacement {
   /// if the given predecessor of the given block is chosen.
   BlockPCs BPCs;
 
+  CandidateType Type;
+
   void printFunction(llvm::raw_ostream &Out) const;
   void printLHS(llvm::raw_ostream &Out, ReplacementContext &Context,
                 bool printNames = false) const;
@@ -67,6 +71,7 @@ struct BlockCandidateSet {
   std::vector<InstMapping> PCs;
   BlockPCs BPCs;
   std::vector<CandidateReplacement> Replacements;
+  llvm::BasicBlock *BB;
 };
 
 struct FunctionCandidateSet {
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -827,12 +827,13 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
       // Harvest Uses (Operands)
       if (HarvestUses) {
         for (auto &Op : I.operands()) {
+          // TODO: support regular values
           if (auto U = dyn_cast<Instruction>(Op)){
             if (U->getType()->isIntegerTy()) {
               Inst *In = EB.get(U);
+              In->HarvestKind = HarvestType::HarvestedFromUse;
               EB.markExternalUses(In);
-              BCS->Replacements.emplace_back(U, InstMapping(In, 0),
-                                             CandidateType::HarvestedFromUse);
+              BCS->Replacements.emplace_back(U, InstMapping(In, 0));
               assert(EB.get(U)->hasOrigin(U));
             }
           }
@@ -851,9 +852,9 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
       } else {
         In = EB.get(&I);
       }
+      In->HarvestKind = HarvestType::HarvestedFromDef;
       EB.markExternalUses(In);
-      BCS->Replacements.emplace_back(&I, InstMapping(In, 0),
-                                     CandidateType::HarvestedFromDef);
+      BCS->Replacements.emplace_back(&I, InstMapping(In, 0));
       assert(EB.get(&I)->hasOrigin(&I));
     }
     if (!BCS->Replacements.empty()) {
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -55,6 +55,10 @@ static llvm::cl::opt<bool> HarvestConstantRange(
     "souper-harvest-const-range",
     llvm::cl::desc("Perform range analysis (default=true)"),
     llvm::cl::init(true));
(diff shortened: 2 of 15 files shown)


regehr:
if you're looking for something easy to work on while home, this might be good?
once it's working I want to run some tests to see how often it succeeds

zhengyangl:
Not ready for review yet.

manasij7479:
@zhengyangl Consider uses inside select and phi instructions too.
For example, in `select %c, %x, %y`  %x and %y can be constrained.   
(Is that correct?)

include/souper/Extractor/Candidates.h
```diff
@@ -38,9 +38,11 @@ class Value;
 
 namespace souper {
 
+enum CandidateType { HarvestedFromUser, HarvestedFromUse };
```
regehr: can you use an enum class here instead of an old-style enum?

lib/Extractor/Candidates.cpp
```diff
@@ -818,7 +822,23 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
 
   for (auto &BB : F) {
     std::unique_ptr<BlockCandidateSet> BCS(new BlockCandidateSet);
+    BCS->BB = &BB;
     for (auto &I : BB) {
+      // Harvest Uses (Operands)
+      if (HarvestUses) {
+        for (auto &Op : I.operands()) {
+          if (auto U = dyn_cast<Instruction>(Op)){
```
regehr: is this test necessary? what about a use that's an argument to a function, don't we want to optimize those too?

include/souper/Extractor/Candidates.h
regehr: I think it may better reflect existing compiler terminology for these to be HarvestedFromDef, HarvestedFromUse

lib/Extractor/Candidates.cpp
```diff
@@ -843,7 +863,7 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
       auto BPCSets = AddBlockPCSets(BCS->BPCs, BPCVars);
 
       for (auto &R : BCS->Replacements) {
-        std::tie(R.BPCs, R.PCs) =
+        std::tie(R.BPCs, R.PCs) =
```
regehr: please avoid adding whitespace

lib/Extractor/Solver.cpp
```diff
@@ -148,6 +148,8 @@ class BaseSolver : public Solver {
       }
     }
 
+    if (T == souper::HarvestedFromUse) return EC;
```
regehr: please put the return on a separate line from the if

lib/Pass/Pass.cpp
```diff
@@ -374,6 +374,7 @@ struct SouperPass : public ModulePass {
     }
 
     for (auto &B : CS.Blocks) {
+      CandidateMap CandMap;
```
regehr: can you explain why this map had to move inside this loop? does it have any other effects that we should think about?

lib/Pass/Pass.cpp
```diff
@@ -385,96 +386,108 @@ struct SouperPass : public ModulePass {
...
+      for (auto &Cand : CandMap) {
+        if (StaticProfile) {
+          std::string Str;
+          llvm::raw_string_ostream Loc(Str);
+          Cand.Origin->getDebugLoc().print(Loc);
+          std::string HField = "sprofile " + Loc.str();
+          ReplacementContext Context;
+          KV->hIncrBy(GetReplacementLHSString(Cand.BPCs, Cand.PCs,
+                                              Cand.Mapping.LHS, Context),
+                      HField, 1);
```
regehr: fix indentation please

lib/Pass/Pass.cpp
```diff
@@ -385,96 +386,108 @@ struct SouperPass : public ModulePass {
...
+            continue;
+          } else {
+            report_fatal_error("Unable to query solver: " + EC.message() + "\n");
+          }
+        }
+        if (!Cand.Mapping.RHS)
+          continue;
 
-      if (ReplaceCount >= FirstReplace && ReplaceCount <= LastReplace) {
+        if (ReplaceCount >= FirstReplace && ReplaceCount <= LastReplace) {
```
regehr: avoid whitespace changes

lib/Pass/Pass.cpp
```diff
@@ -385,96 +386,108 @@ struct SouperPass : public ModulePass {
...
 
-	Instruction *I = Cand.Origin;
-	assert(Cand.Mapping.LHS->hasOrigin(I));
-	IRBuilder<> Builder(I);
+          Instruction *I = Cand.Origin;
+          assert(Cand.Mapping.LHS->hasOrigin(I));
+          IRBuilder<> Builder(I);
```
regehr: avoid whitespace changes

lib/Pass/Pass.cpp
```diff
@@ -385,96 +386,108 @@ struct SouperPass : public ModulePass {
...
+          if (DynamicProfile)
+            dynamicProfile(F, Cand);
+
+          if (Cand.Type == HarvestedFromUser) {
+            I->replaceAllUsesWith(NewVal);
+          } else {
+            llvm::Value::use_iterator UI = I->use_begin(), E = I->use_end();
+            for (; UI != E;) {
+              llvm::Use &U = *UI;
+              ++UI;
```
regehr: this should go in the "increment" part of the for loop construct, right?

lib/Pass/Pass.cpp
```diff
@@ -385,96 +386,108 @@ struct SouperPass : public ModulePass {
...
+          }
 
-	ReplacedValues[Cand.Mapping.LHS] = NewVal;
+          if (DynamicProfile)
+            dynamicProfile(F, Cand);
+
+          if (Cand.Type == HarvestedFromUser) {
+            I->replaceAllUsesWith(NewVal);
+          } else {
+            llvm::Value::use_iterator UI = I->use_begin(), E = I->use_end();
```
regehr: is there not a way to use a ranged for loop here?

lib/Tool/CandidateMapUtils.cpp
```diff
@@ -85,7 +85,7 @@ bool SolveCandidateMap(llvm::raw_ostream &OS, CandidateMap &M,
 
       Inst *RHS = 0;
       if (std::error_code EC =
-              S->infer(Cand.BPCs, Cand.PCs, Cand.Mapping.LHS, RHS, IC)) {
+          S->infer(Cand.BPCs, Cand.PCs, Cand.Mapping.LHS, RHS, IC, HarvestedFromUser)) {
```
regehr: avoid whitespace changes

lib/Tool/CandidateMapUtils.cpp
```diff
@@ -123,7 +123,7 @@ bool CheckCandidateMap(llvm::Module &Mod, CandidateMap &M, Solver *S,
   for (auto &Cand : M) {
     Inst *RHS = 0;
     if (std::error_code EC =
-            S->infer(Cand.BPCs, Cand.PCs, Cand.Mapping.LHS, RHS, IC)) {
```
regehr: avoid whitespace changes

test/Extractor/harvest-uses.ll
```diff
@@ -0,0 +1,31 @@
+; REQUIRES: solver
+
+; RUN: %llvm-as -o %t %s
+; RUN: %opt -load %pass -souper %solver -souper-harvest-uses -S -o - %s | %FileCheck %s
+
+define i1 @fn1(i1) {
+  %x = and i1 1, %0
```
regehr: can we make this optimization work even without this weird extra instruction?

test/Extractor/harvest-uses.ll
```diff
@@ -0,0 +1,31 @@
...
+
+l2:
+  ret i1 %x
+}
+
+
+define i8 @fn2(i8) {
+  %x = icmp slt i8 0, %0
+  %y = sdiv i8 %0, 16
+  ; CHECK-NOT: ashr i8 %0, 4
```
regehr: what is this line trying to do?

zhengyangl: This line is trying to test the use  of %y  in branch %l1  does not call instruction synthesis.

lib/Pass/Pass.cpp
zhengyangl: Correct, I'll get this fixed.

lib/Extractor/Candidates.cpp
zhengyangl: The ExprBuilder only allows Instruction as input for now, I can make the ExprBuilder to allow argument as input and transform it into Souper::Var.

include/souper/Extractor/Candidates.h
zhengyangl: Def is better since a User might not contain a return value for harvesting.

lib/Extractor/Candidates.cpp
zhengyangl: This will be another PR.

lib/Pass/Pass.cpp
zhengyangl: The original program loops the CS.Blocks, push all the replacement candidates in each block into CandMap. After all the candidates are pushed into CandMap, then do the replacements. By doing this these candidates loses the information of their original block, and later in replacement stage it cannot tell which use it should be replaced. 
This patch does the replacement in the block loop. CandMap gets refreshed for each Block, and the corresponding uses can now be replaced correctly.
I don't think there is any other side effect.

zhengyangl: Actually some same replacements might be listed more than once even it is harvested from defs. I'm fixing this.

regehr:
looks great, thanks!

PR #379, opened 2018-12-03
currently, souper attempts to synthesize a better computation for all

regehr:
integer-typed registers, even those that have no uses. previously this
was a bit wasteful, but not problematic. however, since turning
demanded bits on by default, souper started synthesizing and replacing
every non-used return value from a function, which is just silly.

this patch causes souper to only extract registers that have at least
one use. the patch is trivial but it broke a lot of tests, which are
all fixed in obvious ways.

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -767,12 +767,14 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
   for (auto &BB : F) {
     std::unique_ptr<BlockCandidateSet> BCS(new BlockCandidateSet);
     for (auto &I : BB) {
-      if (I.getType()->isIntegerTy()) {
-        Inst *In = EB.get(&I);
-        EB.markExternalUses(In);
-        BCS->Replacements.emplace_back(&I, InstMapping(In, 0));
-        assert(EB.get(&I)->hasOrigin(&I));
-      }
+      if (!I.getType()->isIntegerTy())
+        continue;
+      if (I.hasNUses(0))
+        continue;
+      Inst *In = EB.get(&I);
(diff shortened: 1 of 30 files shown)


lib/Extractor/Candidates.cpp
```diff
@@ -767,12 +767,14 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
...
     for (auto &I : BB) {
-      if (I.getType()->isIntegerTy()) {
-        Inst *In = EB.get(&I);
-        EB.markExternalUses(In);
-        BCS->Replacements.emplace_back(&I, InstMapping(In, 0));
-        assert(EB.get(&I)->hasOrigin(&I));
-      }
+      if (!I.getType()->isIntegerTy())
+        continue;
+      if (I.hasNUses(0))
```
manasij7479: What is the rationale for this?  
It is possible that a new RHS could utilize a value which was not used.  
That is easier than synthesizing it from scratch, especially if constants are involved.

regehr: this patch causes souper to refuse to optimize a value that isn't used anywhere

you're talking about a different issue, which is where some value that is used could perhaps be computed in a way that takes advantage of a value that isn't used, but which happens to be sitting around. making souper use those is a different problem.

and anyway, LLVM aggressively cleans up unused values when possible, so souper seeing them would be a matter of chance and phase ordering.

manasij7479: Okay, maybe it would make sense to harvest a set of 'components' independently from the LHS candidate.   
LGTM

regehr: this might work-- but let's have the conversation elsewhere

PR #380, opened 2018-12-03
upgrading Z3 to latest breaks this test and it's not really obvious

regehr:
how to fix it, so just remove it since it's not that different than a
lot of other synthesis tests that we have sitting around anyway


manasij7479:
Can we keep this around with an XFAIL flag?

regehr:
sure, make the pull req and I'll delete this one

PR #381, opened 2018-12-03
Mark a test XFAIL with new Z3 and make some icmp tests less fragile

manasij7479:


regehr:
ah, good, I was going to do the icmp ones in a separate PR
does this fix all test failures that happen when upgrading to the latest Z3?

manasij7479:
All tests seem to pass with Z3 4.8.3.
Not sure if that's the one you are targeting.

regehr:
LGTM thanks!

PR #388, opened 2018-12-10
demanded bits value should be set for instruction that matches the characteristics of an existing instruction

jubitaneja:


lib/Inst/Inst.cpp
```diff
@@ -654,6 +654,43 @@ Inst *InstContext::getInst(Inst::Kind K, unsigned Width,
   return N;
 }
 
+Inst *InstContext::getInst(Inst::Kind K, unsigned Width,
```
regehr: are you duplicating the whole body of this function? don't do that, just make one version call the other one

lib/Extractor/Candidates.cpp
```diff
@@ -246,35 +251,38 @@ void ExprBuilder::markExternalUses (Inst *I) {
...
       return makeArrayRead(V); // could be a vector operation
 
     Inst *L = get(ICI->getOperand(0)), *R = get(ICI->getOperand(1));
+
+    llvm::APInt DemandedBitsVal = DB->getDemandedBits(ICI);
+
     switch (ICI->getPredicate()) {
       case ICmpInst::ICMP_EQ:
-        return IC.getInst(Inst::Eq, 1, {L, R});
+        return IC.getInst(Inst::Eq, 1, {L, R}, DemandedBitsVal);
```
regehr: are you going to change these back to calling the 3-argument version?

regehr:
let me know when this is ready for more review

PR #392, opened 2018-12-12
Do not reinitialize TriedVars after each constant synthesis try

zhengyangl:
This patch fixes the issue of the bug causing constant synthesis loop with same specialized inputs.


regehr:
LGTM, thanks!

PR #394, opened 2018-12-12
Do not reinitialize badConsts in each constant synthesis try

zhengyangl:
This patch fix the bug causing multiple guesses of same constant.


regehr:
wow that fix was easy :)
should I wait for tests to complete or can this be merged?

zhengyangl:
@regehr This is a rather stupid bug made by me. This PR can be merged directly.

manasij7479:
The test seems to be taking unusually long.  
Can you check if there is an unexpected performance regression?

PR #396, opened 2018-12-25
Support Constant Range from LVI and SCEV

jubitaneja:


lib/Extractor/Candidates.cpp
```diff
@@ -171,7 +181,26 @@ Inst *ExprBuilder::makeArrayRead(Value *V) {
       Negative = isKnownNegative(V, DL);
       NumSignBits = ComputeNumSignBits(V, DL);
     }
-  return IC.createVar(Width, Name, Known.Zero, Known.One, NonZero, NonNegative,
+
+    // constant range
+    ConstantRange Range = llvm::ConstantRange(Width, true);
```
regehr: for boolean function arguments always include a comment like this "ConstantRange(Width, /*isFullSet=*/false);"

lib/Extractor/ExprBuilder.cpp
```diff
@@ -376,6 +376,31 @@ Inst *ExprBuilder::getDemandedBitsCondition(Inst *I) {
                                    LIC->getInst(Inst::Eq, 1, {Res, Zero})});
     Result = LIC->getInst(Inst::And, 1, {Result, SignBits});
   }
+  if (!I->Range.isEmptySet() && !I->Range.isFullSet()) {
+    Inst *Lower = LIC->getConst(I->Range.getLower());
+    Inst *Upper = LIC->getConst(I->Range.getUpper());
+    Inst *NonWrappedRange = LIC->getConst(llvm::APInt(1, 0));
```
regehr: you don't need these two temporaries, just add the appropriate condition directly to Result in the wrapped or unwrapped case

lib/Parser/Parser.cpp
```diff
@@ -43,6 +43,7 @@ struct Token {
     KnownBits,
     OpenParen,
     CloseParen,
+    RangeOpenParen,
```
regehr: this isn't a good name for the character that opens a ConstantRange, which is not a paren

lib/Parser/Parser.cpp
```diff
@@ -408,8 +414,9 @@ bool Parser::typeCheckOpsMatchingWidths(llvm::MutableArrayRef<Inst *> Ops,
                                         std::string &ErrStr) {
   unsigned Width = 0;
   for (auto Op : Ops) {
-    if (Width == 0)
+    if (Width == 0) {
```
regehr: please do not make spurious changes like this

lib/Parser/Parser.cpp
```diff
@@ -1053,8 +1060,11 @@ bool Parser::parseLine(std::string &ErrStr) {
       Block *B = 0;
 
       if (IK == Inst::Var) {
+        // TODO: Maybe, we should initialize lower bound and upper bound to min/max value
+        // of int range w.r.t. width. for now, its initialized to 0.
         llvm::APInt Zero(InstWidth, 0, false), One(InstWidth, 0, false),
-                    ConstOne(InstWidth, 1, false);
+                    ConstOne(InstWidth, 1, false), Lower(InstWidth, 0, false),
+                    Upper(InstWidth, 0, false);
```
regehr: why not just create a constantrange here instead of separate lower and upper variables?

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1132,66 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                  if (CurTok.K != Token::RangeOpenParen) {
+                    ErrStr = makeErrStr(TP, "expected '[' to specify lower bound of range");
+                    return false;
+                  }
+
+                  // look for untypedconst
+                  if (!consumeToken(ErrStr))
+                    return false;
+                  if (CurTok.K != Token::UntypedInt) {
+                    ErrStr = makeErrStr(TP, "expected lower bound of range without width");
```
regehr: don't specify "without width" unless the only way the user can go wrong here is providing an integer value with a width attached

test/Solver/range1.ll
```diff
@@ -0,0 +1,14 @@
+; REQUIRES: solver
```
regehr: in this and other test cases you need a "check" line or else you're not actually checking for anything in particular to happen

jubitaneja:
Thanks for your review comments. I have fixed and updated the PR.

lib/Extractor/ExprBuilder.cpp
```diff
@@ -376,6 +376,22 @@ Inst *ExprBuilder::getDemandedBitsCondition(Inst *I) {
                                    LIC->getInst(Inst::Eq, 1, {Res, Zero})});
     Result = LIC->getInst(Inst::And, 1, {Result, SignBits});
   }
+  if (!I->Range.isEmptySet() && !I->Range.isFullSet()) {
```
regehr: please add a TODO saying that we might want to look into handling the EmptySet case later

lib/Inst/Inst.cpp
```diff
@@ -184,6 +184,9 @@ std::string ReplacementContext::printInstImpl(Inst *I, llvm::raw_ostream &Out,
           Out << " (powerOfTwo)";
         if (I->NumSignBits > 1)
           Out << " (signBits=" << I->NumSignBits << ")";
+        if (!I->Range.isEmptySet() && !I->Range.isFullSet())
```
regehr: I think you should print the empty set case, unless it is somehow impossible

lib/Parser/Parser.cpp
regehr: please do this

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1131,74 @@ bool Parser::parseLine(std::string &ErrStr) {
                   }
                   if (!consumeToken(ErrStr))
                     return false;
+                } else if (CurTok.str() == "range") {
+                  // look for =
```
regehr: comment not needed

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1131,74 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                  if (!consumeToken(ErrStr))
+                    return false;
+                  if (CurTok.K != Token::Eq) {
+                    ErrStr = makeErrStr(TP, "expected '=' for range as 'range='");
+                    return false;
+                  }
+
+                  // look for [ (inclusive paren)
```
regehr: comment not needed

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1131,74 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                  if (!consumeToken(ErrStr))
+                    return false;
+                  if (CurTok.K != Token::OpenBracket) {
+                    ErrStr = makeErrStr(TP, "expected '[' to specify lower bound of range");
+                    return false;
+                  }
+
+                  // look for untypedconst
```
regehr: comment not needed

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1131,74 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                  if (!consumeToken(ErrStr))
+                    return false;
+                  if (CurTok.K != Token::UntypedInt && CurTok.K == Token::Int) {
```
regehr: what purpose does this check serve?

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1131,74 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                    ErrStr = makeErrStr(TP, "expected lower bound of range");
+                    return false;
+                  }
+                  Lower = CurTok.Val;
+
+                  if (Lower.getBitWidth() != InstWidth) {
+                    Lower = Lower.sextOrTrunc(InstWidth);
+                  }
+
+                  // look for comma
```
regehr: remove all of these comments, which are about things that are pretty obvious from looking at the code

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1131,74 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                    return false;
+                  if (CurTok.K != Token::Comma) {
+                    ErrStr = makeErrStr(TP, "expected ',' after lower bound of range");
+                    return false;
+                  }
+
+                  // look for untypedconst
+                  if (!consumeToken(ErrStr))
+                    return false;
+                  if (CurTok.K != Token::UntypedInt && CurTok.K == Token::Int) {
```
regehr: this check also doesn't seem helpful, I think it can be removed

test/Solver/range1.ll
```diff
@@ -0,0 +1,23 @@
...
+  %v1 = load i8, i8* %0, !range !0
+  %v2 = load i8, i8* %1, !range !1
+  ; CHECK: instruction:
+  ; CHECK-NEXT: %out = icmp eq i8 %v1, %v2
+  ; CHECK-NEXT: unexpected simplification:
+  ; CHECK-NEXT: Function: cmp_with_range
+  ; CHECK-NEXT: %0:i8 = var (knownBits=000001xx) (nonNegative) (nonZero) (signBits=5) (range=[5,7)
+  ; CHECK-NEXT: %1:i8 = var (knownBits=0000000x) (nonNegative) (signBits=7) (range=[0,2)
+  ; CHECK-NEXT: %2:i1 = eq %0, %1
+  ; CHECK-NEXT: cand %2 0:i1
```
regehr: to avoid fragile test cases, when possible, you should only check for the important thing that identifies the test case as having been processed successfully by souper
in this case I think it's only the "cand" line that you want to look for, right? this also makes the test cases smaller and easier to read

test/Solver/range1.ll
```diff
@@ -0,0 +1,23 @@
...
+; RUN: %llvm-as -o %t %s
+; RUN: %souper %solver -check %t 2> %t2 || true
+; RUN: %FileCheck %s < %t2
+
+define i1 @cmp_with_range(i8*, i8*) {
+  %v1 = load i8, i8* %0, !range !0
+  %v2 = load i8, i8* %1, !range !1
+  ; CHECK: instruction:
+  ; CHECK-NEXT: %out = icmp eq i8 %v1, %v2
+  ; CHECK-NEXT: unexpected simplification:
```
regehr: also i'm not sure what it means to be checking for "unexpected simplification" -- why check for this? does it mean that you are checking to make sure that souper found an optimization that it was not supposed to find?

test/Solver/range2.ll
```diff
@@ -0,0 +1,25 @@
...
+; RUN: %souper %solver -check %t 2> %t2 || true
+; RUN: %FileCheck %s < %t2
+
+define i1 @cmp_with_range(i8*, i8*) {
+  %v1 = load i8, i8* %0, !range !0
+  %v2 = load i8, i8* %1, !range !1
+  %v3 = shl i8 %v1, %v2
+  ; CHECK: instruction:
+  ; CHECK-NEXT: %out = icmp ule i8 %v3, 32
+  ; CHECK-NEXT: unexpected simplification:
```
regehr: same here and in others

test/Tool/range12.opt
```diff
@@ -0,0 +1,11 @@
+; REQUIRES: solver
+
+; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
+; RUN: %FileCheck %s < %t
+
+; CHECK: LGTM
+%0:i8 = var (knownBits=000000xx) (nonNegative) (nonZero) (signBits=6) (range=[1,3))
```
regehr: can you explain the purpose of providing all of these dataflow facts? are they necessary to make the test case work? if not, you should remove the extra ones as this is confusing and also provides souper with redundant information that makes it hard to figure out what effect the range information is having

test/Tool/range13.opt
```diff
@@ -0,0 +1,10 @@
+; REQUIRES: solver
+
+; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
+; RUN: %FileCheck %s < %t
+
+; CHECK: LGTM
```
regehr: this is a good test, but in some cases (like this one) you should also have a negative test where you make sure that an optimzation is not performed when it isn't justified by the ranges
even better would be if the positive and negative test cases are right "next to" each other -- that is, there's a corner case right near INT_MAX / 2 where overflow changes from impossible to possible. if your tests straddle that edge case, it is a better way to make sure you got everything right

regehr:
I didn't look that carefully, but at a quick glance it didn't look like very many test cases exploited wrapped ranges-- this should be done in at least a few cases. remember, the purpose of testing is to try to break your code, in any edge cases where it might be wrong.

regehr:
also, when it's conveninent, err on the side of more LLVM-level tests instead of souper-level tests since they tend to test more code

regehr:
finally, we'd like a few round-trip tests for ranges (see ParserTests.cpp)
I think you missed these for the other dataflow facts, but don't worry about that now

regehr:
finally I think you're not checking for the case where a range contains a too-large integer, for example [0,700) for an i8. this should error out in the parser.

lib/Parser/Parser.cpp
jubitaneja: I wanted to handle this separately so that if someone writes Souper IR directly and mentions lower/upper bound as '5:i32' (which is Token::Int) should get an informative error message that bounds must be specified without width. I can simplify this error check to this actually:
if (CurTok.K == Token::Int) {...}

lib/Parser/Parser.cpp
jubitaneja: If you think, we don't need this kind of information, I can take it off completely. Otherwise, I was just going to simplify this error check as I mentioned above. IMO, this seems useful. Because, while I was writing the tests manually, I accidentally specified bitwidth against all constants that I used for the lower/upper bound. And, having such error message was helpful to figure out what's wrong in this.

test/Solver/range1.ll
```diff
@@ -0,0 +1,16 @@
+; REQUIRES: solver
+
+; RUN: %llvm-as -o %t %s
+; RUN: %souper %solver -check %t 2> %t2 || true
```
regehr: what is the purpose of the || true in this and other tests?

test/Tool/range6.opt
```diff
@@ -0,0 +1,18 @@
...
+; max = 4096 - (-1024) [upper_1 - lower2]
+
+%0:i32 = var (range=[1024,4097))
+%1:i32 = var (range=[-1024,2049))
+%2:i32 = subnw %0, %1
+%3:i1 = sle -1024, %2
+%4:i1 = slt %2, 5120
+%6:i1 = and %3, %4
+cand %6 1:i1
+
```
regehr: I think you can delete one of range5 and range6 since they seem to test the same thing

test/Tool/range3.opt
```diff
@@ -0,0 +1,12 @@
+; REQUIRES: solver
+
+; RUN: %souper-check %solver -print-counterexample=false  %s > %t 2>&1
+; RUN: %FileCheck %s < %t
+
+; CHECK: LGTM
+%0:i32 = var (range=[2, 5))
+%1:i32 = var (range=[3, 5))
```
regehr: %1 doesn't seem to be used anywhere?

test/Solver/range1.ll
```diff
@@ -0,0 +1,16 @@
...
+
+define i1 @cmp_with_range(i8*, i8*) {
+  %v1 = load i8, i8* %0, !range !0
+  %v2 = load i8, i8* %1, !range !1
+  ; CHECK: cand %2 0:i1
+  %out = icmp eq i8 %v1, %v2
+  ret i1 %out
+}
+
+!0 = !{i8 5, i8 7}
```
regehr: please add at least one LLVM IR test where the range information provided is too weak to make an optimization happen

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1129,102 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                    ErrStr = makeErrStr(TP, "expected lower bound of range without width");
+                    return false;
+                  }
+                  if (CurTok.K != Token::UntypedInt) {
+                    ErrStr = makeErrStr(TP, "expected lower bound of range");
+                    return false;
+                  }
+                  Lower = CurTok.Val;
+
+                  if (Lower.getBitWidth() != InstWidth) {
```
regehr: the code in this block seems too complicated for what you are trying to accomplish

regehr: I suggest writing a little helper function that checks if an APInt fits into a different (smaller) bitwidth and then just calling it to validate both the lower and upper bounds

regehr: in fact it looks like APInt already has the helper functions you need: isintN() and isSignedIntN()

lib/Extractor/ExprBuilder.cpp
```diff
@@ -376,6 +376,23 @@ Inst *ExprBuilder::getDemandedBitsCondition(Inst *I) {
                                    LIC->getInst(Inst::Eq, 1, {Res, Zero})});
     Result = LIC->getInst(Inst::And, 1, {Result, SignBits});
   }
+  // TODO: We might want to look into handling empty set later
+  if (!I->Range.isEmptySet() && !I->Range.isFullSet()) {
+    Inst *Lower = LIC->getConst(I->Range.getLower());
+    Inst *Upper = LIC->getConst(I->Range.getUpper());
+ 
+    // not wrapped set: x >= Lower && x <= Upper
```
regehr: this comment isn't correct, it's x < Upper, right?
anyway, since the code below is clear, you can delete these lines

lib/Extractor/ExprBuilder.cpp
```diff
@@ -376,6 +376,23 @@ Inst *ExprBuilder::getDemandedBitsCondition(Inst *I) {
...
+    // wrapped set: x >= Lower || x <= Upper
+    if (!I->Range.isWrappedSet()) {
+      Result = LIC->getInst(Inst::And, 1, {Result, LIC->getInst(Inst::And, 1,
+                            {LIC->getInst(Inst::Ule, 1, {Lower, I}),
+                             LIC->getInst(Inst::Ult, 1, {I, Upper})})});
+    } else if (I->Range.isWrappedSet()) {
```
regehr: why are you checking !isWrappedSet() and then also isWrappedSet()??

lib/Parser/Parser.cpp
regehr: I think the easy thing to do here is to call both of these functions, since textual Souper IR is allowed to specify numbers either signed or unsigned

jubitaneja: Thanks, this API has saved me from writing that ugly code. I added a check: if (isSignedIntN(Width) || isIntN(Width)) {OK} else {ERROR}. Now, in general, i8 will have numbers between -128 to 255, i.e. -128 to 127 and 0 to 255. (though I am not happy with this kind of check here because signed and unsigned have different range). But, surely if we have 256 for an i8, it will return an error. Since Souper IR cannot explicitly specify signed or unsigned variables, should we make this check complicated to see if one of the values of a range is negative, then it must satisfy isSignedIntN() API?

lib/Extractor/ExprBuilder.cpp
jubitaneja: I did that earlier because I thought of handling isSignedWrappedSet() separately, but it's not handled here and is just an else case now.

test/Solver/range1.ll
jubitaneja: Yes, the last thing to be added is this.

test/Tool/range3.opt
jubitaneja: Thanks for catching this test, I had something else in my notepad and wrote a different test here. I have fixed it.

test/Solver/range1.ll
jubitaneja: If I run this test separately by running all the commands individually, it passes. However, if I put it a test suite without "|| true", it fails. I saw more test files where for the same kind of tests which perform an unexpected simplification, true is added to makes sure that it passes. I did the same here. Honestly, I am not sure why it is done for those tests too. But, I did run it with both debug/release build to find if there is any error in the test, but there is nothing like that because all these tests run successfully when ran individually. Reference: souper/test/Tool/unexpected1.ll

lib/Parser/Parser.cpp
regehr: right, if it's negative use is SignedIntN and otherwise use the unsigned check. this won't be very complicated. you should have test cases making sure we reject -129 and 256, but accept -128 and 255.

test/Solver/range1.ll
regehr: some of the existing tests do this to describe the situation where souper fails (return non-zero to the OS) but the test case passes. but your test cases aren't of that kind, so you don't want this in your tests.

regehr: at least that's what I think is going on, you should make sure.

jubitaneja: Well, I see the exact same behavior here. I get an exit code 1 but the test case passes. So, I added "true", is that fine?

jubitaneja: For reference: 
******************** TEST 'Souper :: Solver/range1.ll' FAILED ********************
Script:
--
: 'RUN: at line 3';   /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/third_party/llvm/Release/bin/llvm-as -o /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/build/test/Solver/Output/range1.ll.tmp /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/test/Solver/range1.ll
: 'RUN: at line 4';   /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/build/souper -z3-path=/uusoc/exports/scratch/jubi/my_repo/tool-installations/install/z3/bin/z3 -check /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/build/test/Solver/Output/range1.ll.tmp 2> /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/build/test/Solver/Output/range1.ll.tmp2
: 'RUN: at line 5';   /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/third_party/llvm/Release/bin/FileCheck /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/test/Solver/range1.ll < /uusoc/exports/scratch/jubi/my_repo/superopt_project/souper/build/test/Solver/Output/range1.ll.tmp2
--
Exit Code: 1

regehr: it doesn't seem right
it's souper returning 1, or filecheck? shouldn't souper-check return 0 if the optimization checks out?

jubitaneja: souper returns 1 here.

regehr: and what is souper's output?

lib/Parser/Parser.cpp
jubitaneja: I have simplified this error check, but I still have two different errors: one to make sure user does not specify width with lower/upper bound, and second to make sure they do specify an untyped bound.

jubitaneja: Please check once, so that we can resolve this comment.

lib/Inst/Inst.cpp
regehr: you need to fix this-- LLVM can give us empty sets and it's wrong to not print that information that souper knows internally

lib/Inst/Inst.cpp
```diff
@@ -574,6 +578,7 @@ Inst *InstContext::createVar(unsigned Width, llvm::StringRef Name,
   I->Number = Number;
   I->Width = Width;
   I->Name = Name;
+  I->Range = Range;
```
regehr: I think you should add some assertions here, making sure that the widths of Range, KnownZeros, and KnownOnes match the width of the var, to make sure we're never getting stuck using the default values for these parameters

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1129,84 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                    ErrStr = makeErrStr(TP, "expected lower bound of range");
+                    return false;
+                  }
+                  Lower = CurTok.Val;
+
+                  if (Lower.isNegative() && !Lower.isSignedIntN(InstWidth)) {
+                    ErrStr = makeErrStr(TP, "Lower bound is out of range");
+                    return false;
+                  }
+                  if (Lower.isSignedIntN(InstWidth) || Lower.isIntN(InstWidth)) {
```
regehr: this is a bit confusing, please separate out the two things you're doing here:
first, the bounds checks
second, trunc / extend if needed

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1129,84 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                    return false;
+                  if (CurTok.K != Token::Comma) {
+                    ErrStr = makeErrStr(TP, "expected ',' after lower bound of range");
+                    return false;
+                  }
+
+                  if (!consumeToken(ErrStr))
+                    return false;
+                  if (CurTok.K == Token::Int) {
+                    ErrStr = makeErrStr(TP, "expected upper bound of range without width");
```
regehr: I still don't see why this error is useful -- if an integer with a width is specified, the next error check will catch this situation and everything is fine. so this is redundant and should be deleted.

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1129,84 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                    return false;
+                  }
+                  Upper = CurTok.Val;
+
+                  if (Upper.isNegative() && !Upper.isSignedIntN(InstWidth)) {
+                    ErrStr = makeErrStr(TP, "Upper bound is out of range");
+                    return false;
+                  }
+                  if (Upper.isSignedIntN(InstWidth) || Upper.isIntN(InstWidth)) {
+                    if (Upper.getBitWidth() != InstWidth)
```
regehr: also move this ext/trunc below the error checks

lib/Parser/Parser.cpp
```diff
@@ -1122,6 +1130,73 @@ bool Parser::parseLine(std::string &ErrStr) {
...
+                  if (!consumeToken(ErrStr))
+                    return false;
+                  if (CurTok.K != Token::UntypedInt) {
+                    ErrStr = makeErrStr(TP, "expected upper bound of range");
+                    return false;
+                  }
+                  Upper = CurTok.Val;
+
+                  if ((Upper.isNegative() && !Upper.isSignedIntN(InstWidth)) ||
+                      (!Upper.isNegative() && !(Upper.isSignedIntN(InstWidth) || Upper.isIntN(InstWidth)))) {
```
regehr: this test can be simplified by factoring out the !isSignedIntN() check -- it has to be done regardless of whether the value is negative

unittests/Parser/ParserTests.cpp
```diff
@@ -176,6 +176,24 @@ TEST(ParserTest, Errors) {
...
+      { "%0:i32 = var (range=2,90))\n",
+        "<input>:1:1: expected '[' to specify lower bound of range" },
+      { "%0:i32 = var (range=[2:i32,90:i32))\n",
+        "<input>:1:1: expected lower bound of range" },
+      { "%0:i32 = var (range=[2,90:i32))\n",
+        "<input>:1:1: expected upper bound of range" },
+      { "%0:i32 = var (range=[2 90))\n",
+        "<input>:1:1: expected ',' after lower bound of range" },
+      { "%0:i32 = var (range=[2,90)\n",
+        "<input>:1:1: expected ')' to complete data flow fact string" },
```
regehr: "data flow fact string" is awkward, how about
expected ')' to complete range

jubitaneja: this check is not just for range, it's outside the big switch statement for each dataflow fact (ref:Parser.cpp). This is in general for any kind of dfa.

lib/Parser/Parser.cpp
```diff
@@ -1136,11 +1211,12 @@ bool Parser::parseLine(std::string &ErrStr) {
               ErrStr = makeErrStr(TP, "expected ')' to complete data flow fact string");
```
jubitaneja: Just for reference. That's what I was mentioning in my previous comment for the parser unit test.

lib/Parser/Parser.cpp
jubitaneja: Fixed. Thanks!

unittests/Parser/ParserTests.cpp
regehr: ok just remove "string" then, it doesn't make sense here

regehr:
I'm getting a lot of failed assertions, did you test a debug build?

lib/Inst/Inst.cpp
```diff
@@ -584,6 +590,13 @@ Inst *InstContext::createVar(unsigned Width, llvm::StringRef Name,
   return I;
 }
 
+Inst *InstContext::createVar(unsigned Width, llvm::StringRef Name) {
+  return createVar(Width, Name, llvm::ConstantRange(Width, /*isFullSet*/ true),
+                   /*KnownZero*/ llvm::APInt(Width, 0), /*KnownOnes*/ llvm::APInt(Width, 0),
+                   /*NonZero*/ false, /*NonNegative*/ false, /*PowerOfTwo*/ false,
+                   /*Negative*/ false, /*SignBits*/ 1);
```
regehr: for all of these please follow the existing idiom, for example:
/*Negative=*/false

PR #399, opened 2018-12-31
Floating point cost values with some heuristics for ExhaustiveSynthesis

manasij7479:
Implements some infrastructure to support richer cost models.

### DIFF
--- a/lib/Infer/ExhaustiveSynthesis.cpp
+++ b/lib/Infer/ExhaustiveSynthesis.cpp
@@ -50,6 +50,8 @@ namespace {
   static cl::opt<bool> EnableBigQuery("souper-exhaustive-synthesis-enable-big-query",
     cl::desc("Enable big query in exhaustive synthesis (default=false)"),
     cl::init(false));
+  static cl::opt<std::string> CostModel("souper-cost-model", cl::desc("Cost Model"),
+    cl::init("default"));
 }
 
 // TODO
@@ -86,6 +88,60 @@ namespace {
 // experiment with synthesizing at reduced bitwidth, then expanding the result
 // aggressively avoid calling into the solver
 
+using CostFunctionType = std::function<float(Inst *, bool)>;
+
+float defaultCostFunction(Inst* I, bool IgnoreDepsWithExternalUses) {
+  return souper::cost(I, IgnoreDepsWithExternalUses);
+}
+
+// TODO : Derive these randomly to maximize infer rate with test-infer.sh
+float getCost(Inst::Kind K) {
+  switch (K) {
+    case souper::Inst::Var:
+    case souper::Inst::Const:
+    case souper::Inst::Phi:
+      return 0;
+    case souper::Inst::BSwap:
+    case souper::Inst::CtPop:
+    case souper::Inst::Cttz:
+    case souper::Inst::Ctlz:
+    case souper::Inst::SDiv:
+    case souper::Inst::UDiv:
+    case souper::Inst::SRem:
+    case souper::Inst::URem:
+      return 5;
+    case souper::Inst::Select:
+      return 3;
+    case souper::Inst::Mul:
+      return 2.5f;
+    case souper::Inst::Add:
+    case souper::Inst::Sub:
+      return 1.5f;
+    default:
+      return 1;
+  }
+}
+
+// TODO: Recognize simple patterns
+static float costHelper(Inst *I, Inst *Root, std::set<Inst *> &Visited,
+                      bool IgnoreDepsWithExternalUses) {
+  if (!Visited.insert(I).second)
+    return 0;
+  if (IgnoreDepsWithExternalUses && I != Root &&
+      Root->DepsWithExternalUses.find(I) != Root->DepsWithExternalUses.end()) {
+    return 0;
+  }
+  float Cost = getCost(I->K);
+  for (auto Op : I->Ops)
+    Cost += costHelper(Op, Root, Visited, IgnoreDepsWithExternalUses);
+  return Cost;
+}
+
+float simpleHeuristicsCostFunction(Inst* I, bool IgnoreDepsWithExternalUses) {
+  std::set<Inst *> Visited;
+  return costHelper(I, I, Visited, IgnoreDepsWithExternalUses);
+}
+
 void hasConstantHelper(Inst *I, std::set<Inst *> &Visited,
                        std::vector<Inst *> &ConstList) {
   // FIXME this only works for one constant and keying by name is bad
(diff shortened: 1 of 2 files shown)


lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -50,6 +50,8 @@ namespace {
   static cl::opt<bool> EnableBigQuery("souper-exhaustive-synthesis-enable-big-query",
     cl::desc("Enable big query in exhaustive synthesis (default=false)"),
     cl::init(false));
+  static cl::opt<std::string> CostModel("souper-cost-model", cl::desc("Cost Model"),
+    cl::init("default"));
```
regehr: in the description you should list the possible choices

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -86,6 +88,60 @@ namespace {
 // experiment with synthesizing at reduced bitwidth, then expanding the result
 // aggressively avoid calling into the solver
 
+using CostFunctionType = std::function<float(Inst *, bool)>;
+
+float defaultCostFunction(Inst* I, bool IgnoreDepsWithExternalUses) {
+  return souper::cost(I, IgnoreDepsWithExternalUses);
```
regehr: might as well move the code over here, instead of calling out to somewhere else

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -86,6 +88,60 @@ namespace {
 // experiment with synthesizing at reduced bitwidth, then expanding the result
 // aggressively avoid calling into the solver
 
+using CostFunctionType = std::function<float(Inst *, bool)>;
```
regehr: let's use double instead of float everywhere

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -86,6 +88,60 @@ namespace {
...
+
+float defaultCostFunction(Inst* I, bool IgnoreDepsWithExternalUses) {
+  return souper::cost(I, IgnoreDepsWithExternalUses);
+}
+
+// TODO : Derive these randomly to maximize infer rate with test-infer.sh
```
regehr: what does "derive these randomly" mean?

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -86,6 +88,60 @@ namespace {
...
+      return 1;
+  }
+}
+
+// TODO: Recognize simple patterns
+static float costHelper(Inst *I, Inst *Root, std::set<Inst *> &Visited,
+                      bool IgnoreDepsWithExternalUses) {
+  if (!Visited.insert(I).second)
+    return 0;
+  if (IgnoreDepsWithExternalUses && I != Root &&
```
regehr: I think you should leave a comment explaining the rationale for counting things with external uses are zero cost

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -119,9 +175,9 @@ std::vector<Inst *> matchWidth(Inst *I, unsigned NewW, InstContext &IC) {
   return { I };
 }
 
-void addGuess(Inst *RHS, int MaxCost, std::vector<Inst *> &Guesses,
-              int &TooExpensive) {
-  if (souper::cost(RHS) < MaxCost)
+void addGuess(Inst *RHS, float MaxCost, std::vector<Inst *> &Guesses,
+              float &TooExpensive, CostFunctionType &Cost) {
```
regehr: rather than passing the cost function around a lot, investigate whether there's a better place to stash this pointer, for example in one of the context objects

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -119,9 +175,9 @@ std::vector<Inst *> matchWidth(Inst *I, unsigned NewW, InstContext &IC) {
...
+  if (Cost(RHS, false) < MaxCost)
```
regehr: when passing a boolean constant, follow the convention of including the name of the flag as a comment

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -445,18 +501,24 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
   if (DebugLevel > 1)
     llvm::errs() << "got " << Inputs.size() << " candidates from LHS\n";
 
+  CostFunctionType Cost;
+  if (CostModel == "default") {
+     Cost = defaultCostFunction;
+  } else {
```
regehr: the else clause needs to be an error

regehr:
I agree that we want pluggable cost functions, but do you need to add a new one right now? it is very similar to the existing one, so maybe the existing one simply wants to be replaced (while also leaving the pluggable infrastructure so that we can plug in more sophisticated stuff later)?

test/Infer/syn-shl-from-add.opt
```diff
@@ -0,0 +1,11 @@
+; REQUIRES: solver, solver-model
+
+; RUN: %souper-check %solver -reinfer-rhs -souper-exhaustive-synthesis -souper-exhaustive-synthesis-num-instructions=1 -souper-cost-model=heuristics %s > %t
```
regehr: this isn't a great cost model test, since many different cost models would allow this optimization to go.
please try to find a place in our unit tests where you can test the cost model directly, including whether it properly deals with the costs of external uses.

lib/Infer/ExhaustiveSynthesis.cpp
manasij7479: Pick a plausible cost range for each instruction and sample frequency.
For all possible cost combinations, run test-infer.sh on the solver/alive directory and pick the combinations with the best infer success rate.

test/Infer/syn-shl-from-add.opt
manasij7479: The default one (without ignoring cost) does not infer this.

lib/Infer/ExhaustiveSynthesis.cpp
manasij7479: > I agree that we want pluggable cost functions, but do you need to add a new one right now? it is very similar to the existing one, so maybe the existing one simply wants to be replaced (while also leaving the pluggable infrastructure so that we can plug in more sophisticated stuff later)?

I think it makes sense to keep the existing one because the CEGIS synthesis needs it.
Also, I want to try adding a few variants and maybe a parametric one that can be controlled from the command line.

manasij7479:
>I agree that we want pluggable cost functions, but do you need to add a new one right now? it is very similar to the existing one, so maybe the existing one simply wants to be replaced (while also leaving the pluggable infrastructure so that we can plug in more sophisticated stuff later)?

I think it makes sense to keep the existing one because the CEGIS synthesis needs it.
Also, I want to try adding a few variants and maybe a parametric one that can be controlled from the command line.

lib/Infer/ExhaustiveSynthesis.cpp
regehr: don't explain it here, explain it in the code

test/Infer/syn-shl-from-add.opt
regehr: it's not a good test case and wants to be improved

regehr:
it's fine to leave the old one if you want, but make the other changes

PR #401, opened 2019-01-03
CMake: explicitly use target_include_directories() to specify include dirs.

LebedevRI:
This is an NFC change for the normal builds.
When souper cmake project is loaded into some IDE (qtcreator in my case),
the IDE's CMake integration it produces a CodeBlocks project, which is
then used by the IDE. That works fine. Nver had any issues with that.

However, there appears to be a problem in this case.
souper is passing include directories (`-I<dir>`) via COMPILE_FLAGS,
and that confuses (blinds?) the CodeBlocks's CMake generator,
these include paths are not listed in the generated project...

That, in turn, results in sub-par IDE experience, since the includes
are failing to be resolved due to missing include paths.

These appear to be all the 'incorrectly' passed include dirs.
I've also took liberty to extend .gitignore file to hide the build
dirs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 souperweb.tar.gz
 third_party
+build
+build-*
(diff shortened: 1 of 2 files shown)


regehr:
LGTM, thanks!

PR #403, opened 2019-01-04
demanded bits wasn't getting added to the folding set ID for instruct…

regehr:
…ions,

leading to miscompilation. this fixes that.


regehr:
nope, default argument doesn't work in this case since we don't know the width

jubitaneja:
LGTM.

PR #404, opened 2019-01-05
CMake: mark 'check' target as using terminal.

LebedevRI:
Else, all the `lit` output is being buffered, and is flushed after
the target finishes 'building'. In particular, that hides the
progress meter.


regehr:
thanks! this had bothered me for a while but I didn't know the fix!

Issue #400, opened 2019-01-02
KLEEBuilder is recursive


LebedevRI:
With any sample larger than just a few instructions, `-souper-infer-inst` now _often_ run out of stack.
[test.ll.txt](https://github.com/google/souper/files/2722002/test.ll.txt)

```
$ /opt/souper/third_party/llvm/Release/bin/opt test.ll.txt -o test.bc
$ redis-cli flushall
$ /opt/souper/souper-build/souper -souper-internal-cache -souper-external-cache -souper-infer-iN -souper-infer-nop -souper-infer-inst -o - test.bc -z3-path=/opt/z3/bin/z3
```
```
$ /opt/souper/souper-build/souper -souper-internal-cache -souper-external-cache -souper-infer-iN -souper-infer-nop -souper-infer-inst -o - test.bc -z3-path=/opt/z3/bin/z3 
; Listing valid replacements.
; Using solver: Z3 + external cache + internal cache
#0 0x0000000000b1f2a4 llvm::sys::PrintStackTrace(llvm::raw_ostream&) /opt/souper/third_party/llvm/Release-build/../lib/Support/Unix/Signals.inc:490:13
#1 0x0000000000b1f2a4 PrintStackTraceSignalHandler(void*) /opt/souper/third_party/llvm/Release-build/../lib/Support/Unix/Signals.inc:553:0
#2 0x0000000000b1d06c llvm::sys::RunSignalHandlers() /opt/souper/third_party/llvm/Release-build/../lib/Support/Signals.cpp:68:18
#3 0x0000000000b1f462 SignalHandler(int) /opt/souper/third_party/llvm/Release-build/../lib/Support/Unix/Signals.inc:353:1
#4 0x00007f3df92ca670 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x12670)
#5 0x0000000000b31ccd (anonymous namespace)::KLEEBuilder::build(souper::Inst*) /opt/souper/souper-build/../lib/Extractor/KLEEBuilder.cpp:121:41
#6 0x0000000000b31bad klee::ref<klee::Expr>::inc() const /opt/souper/souper-build/../third_party/klee/include/klee/util/Ref.h:40:9
#7 0x0000000000b31bad klee::ref<klee::Expr>::operator=(klee::ref<klee::Expr> const&) /opt/souper/souper-build/../third_party/klee/include/klee/util/Ref.h:76:0
[...]

PR #405, opened 2019-01-05
Specify -souper-synthesis-comps= in syn-inst-trunc.ll test

LebedevRI:
While currently that test does not crash (for me at least,
with standard stack size), it does start crashing like in
https://github.com/google/souper/issues/400
with WIP patch for `fsh[lr]` handling.

The rest (or at least some of) of the `syn-inst-*.ll` tests
do specify `-souper-synthesis-comps=` explicitly, so i guess
this one should/can too?

Note that i specify `const`, since `trunc` is implicit,
and can't be specified manually there.

### DIFF
--- a/test/Pass/syn-inst-trunc.ll
+++ b/test/Pass/syn-inst-trunc.ll
@@ -1,7 +1,7 @@
 ; REQUIRES: solver
 
 ; RUN: %llvm-as -o %t %s
-; RUN: %opt -load %pass -souper -dce %solver -souper-infer-inst -S -o - %s | %FileCheck %s
+; RUN: %opt -load %pass -souper -dce %solver -souper-infer-inst -souper-synthesis-comps=add,sub,mul,const -S -o - %s | %FileCheck %s
 
 ; Translated from test/Infer/odd.opt
 


LebedevRI:
Also, it would be really great to have some doc with an overview of the test/ layout.
It isn't really obvious where what should be tested.

test/Pass/syn-inst-trunc.ll
```diff
@@ -1,7 +1,7 @@
 ; REQUIRES: solver
 
 ; RUN: %llvm-as -o %t %s
-; RUN: %opt -load %pass -souper -dce %solver -souper-infer-inst -S -o - %s | %FileCheck %s
+; RUN: %opt -load %pass -souper -dce %solver -souper-infer-inst -souper-synthesis-comps=const -S -o - %s | %FileCheck %s
```
regehr: if you don't specify the components to use, the default is to use them all-- which uses a lot of resources. but if you specify just a constant, synthesis is extremely boring and the test cast isn't very interesting. so maybe add a few more instructions to this list? just random stuff like add,sub,mul is fine.

LebedevRI: Sure, i can do that. I'm just not very sure *what* which test should be testing.
I.e. `test/Pass/`, are we testing here that we can synthesize `trunc`, or that we pick `trunc` and not some other alternative?
Like i said, it would be awesome to have an overview.

regehr:
LGTM, thanks
I agree we should document the testing structure-- I've added it to my todo list

LebedevRI:
> LGTM, thanks
> I agree we should document the testing structure-- I've added it to my todo list

Thank you.
I bundled some docs with https://github.com/google/souper/pull/406

regehr:
great, thanks -- I'll review that one hopefully in the next day or two (I start teaching this week so things are a little busy!)

PR #406, opened 2019-01-06
Support for LLVM IR funnel shift left/right intrinsic

LebedevRI:
LangRef: https://releases.llvm.org/7.0.0/docs/LangRef.html#llvm-fshl-intrinsic

```
The ‘llvm.fshl’/‘llvm.fshr’ family of intrinsic functions performs
a funnel shift left/right: the first two values are concatenated
as { %a : %b } (%a is the most significant bits of the wide value),
the combined value is shifted left/right,
and the most/least significant bits are extracted to produce a result
that is the same size as the original arguments.
<...>
For vector types, the operation occurs for each element of the vector.
The shift argument is treated as an unsigned amount modulo
the element size of the arguments.
```

Observations:
* If first and second arguments are the same, it's just a normal rotate.
* If it's ‘llvm.fshl’, and second argument is `0`, it's just `shl`.
* If it's ‘llvm.fshr’, and first argument is `0`, it's just `lshr`.
* If it's ‘llvm.fshl’, and third argument is `0`, it's just first argument.
* If it's ‘llvm.fshr’, and third argument is `0`, it's just second argument.

Doubts:
* Do we really want to be this aggressive in synthesizing `fshl`/`fshr`?
  Currently, i tuned `Inst::getCost()` to synthesize it in all the cases
  covered in newly-added `test/Pass/` tests.
* I did not check exhaustive synthesis (no tests).
* First real patch to souper, i'm sure i have missed things,
  or failed at test coverage.

Bugs:
* If the shift amount is constant, even though
  `InstSynthesis::createCleanInst()` does reduce the value of that
  constant (makes it modulo bitwidth), it does not want to use this
  replacement, since the cost is identical.

Bonus:
* Added some docs to the `test/*/README.md` with my understanding of
  what test should be put where.

CC @rotateright

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -510,6 +510,14 @@ Inst *ExprBuilder::buildHelper(Value *V) {
           return IC.getInst(Inst::Cttz, L->Width, {L});
         case Intrinsic::ctlz:
           return IC.getInst(Inst::Ctlz, L->Width, {L});
+        case Intrinsic::fshl:
+        case Intrinsic::fshr: {
+          Inst *Low = get(II->getOperand(1));
+          Inst *ShAmt = get(II->getOperand(2));
+          Inst::Kind K =
+              II->getIntrinsicID() == Intrinsic::fshl ? Inst::FShl : Inst::FShr;
+          return IC.getInst(K, L->Width, {/*High*/L, Low, ShAmt});
+        }
         case Intrinsic::sadd_with_overflow: {
           Inst *R = get(II->getOperand(1));
           Inst *Add = IC.getInst(Inst::Add, L->Width, {L, R}, /*Available=*/false);
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -516,7 +516,7 @@ Inst *ExprBuilder::buildHelper(Value *V) {
           Inst *ShAmt = get(II->getOperand(2));
           Inst::Kind K =
               II->getIntrinsicID() == Intrinsic::fshl ? Inst::FShl : Inst::FShr;
-          return IC.getInst(K, L->Width, {/*High*/L, Low, ShAmt});
+          return IC.getInst(K, L->Width, {/*High=*/L, Low, ShAmt});
         }
         case Intrinsic::sadd_with_overflow: {
           Inst *R = get(II->getOperand(1));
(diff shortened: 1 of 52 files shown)


nunoplopes:
Not sure @pranavk wasn't working on it already? @regehr 
If not, I would wait for us to merge his pull request to avoid conflicts (it's on cttz).

LebedevRI:
> Not sure @pranavk wasn't working on it already? @regehr

Hence the question, really.

> If not, I would wait for us to merge his pull request to avoid conflicts (it's on cttz).

That's not really an issue for me, `git rebase` does wonders.

pranavk:
You can go ahead. I haven't yet started working on it.

LebedevRI:
@pranavk,@nunoplopes ok, thank you, i'll take a look.

lib/Extractor/Candidates.cpp
```diff
@@ -510,6 +510,14 @@ Inst *ExprBuilder::buildHelper(Value *V) {
           return IC.getInst(Inst::Cttz, L->Width, {L});
         case Intrinsic::ctlz:
           return IC.getInst(Inst::Ctlz, L->Width, {L});
+        case Intrinsic::fshl:
+        case Intrinsic::fshr: {
+          Inst *Low = get(II->getOperand(1));
+          Inst *ShAmt = get(II->getOperand(2));
+          Inst::Kind K =
+              II->getIntrinsicID() == Intrinsic::fshl ? Inst::FShl : Inst::FShr;
+          return IC.getInst(K, L->Width, {/*High*/L, Low, ShAmt});
```
regehr: please use the ``/*High=*/`` idiom often seen in llvm code

test/Infer/fshl-reduce-const-shamt.ll
```diff
@@ -0,0 +1,20 @@
+; REQUIRES: solver
+
+; RUN: %llvm-as %s -o - | %souper %solver -souper-infer-inst -souper-synthesis-comps=fshl,const | %FileCheck %s
+
+; XFAIL: *
+; FIXME: does not seem to try to reduce the constant.
```
regehr: right, souper currently has no preferences among constants -- in general doing a good job at this seems difficult, though in this particular case we might want to add a specific constraint into constant synthesis that makes sure a funnel shift always gets the reduced version of a constant

regehr:
thanks for this patch, and for doing a careful job with test cases, and for helping out with the test documentation!
I didn't yet read the second 50% of your test cases-- will do that soon. overall the patch looks excellent.

test/Infer/fshl-reduce-const-shamt.ll
LebedevRI: Right. I have looked around in the code, and to me it isn't obvious where that should go, so i think i'll just leave everything as-is for now.

regehr:
LGTM-- thanks again!

LebedevRI:
Thank _you_, this has been interesting!

PR #409, opened 2019-01-09
Alive backend implementation

manasij7479:
Limitations:
* Fails 5 test cases, all except Pass/syn-icmp-variants.ll because of missing instructions. 
* Can not synthesize multiple constants
* Requires changes in Alive ( https://github.com/AliveToolkit/alive2/pull/22 )
* Does not support CEGIS implementation (InstSynthesis.cpp) yet. (should be straightforward to implement)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,21 +69,27 @@ execute_process(
 
 set(CLANG_CXXFLAGS "")
 set(CLANG_INCLUDEDIR "${LLVM_SRC}/tools/clang/include" "${LLVM_BUILD}/tools/clang/include")
+
+set(ALIVE_DIR "${CMAKE_SOURCE_DIR}/third_party/alive2")
+set(ALIVE_LDFLAGS "-L ${ALIVE_DIR}/build -lalive2")
+
 set(CLANG_LIBS "-lclangCodeGen -lclangTooling -lclangRewrite -lclangFrontend -lclangAnalysis -lclangParse -lclangSerialization -lclangSema -lclangEdit -lclangAnalysis -lclangAST -lclangDriver -lclangLex -lclangBasic")
 
 set(GTEST_CXXFLAGS "-DGTEST_HAS_RTTI=0")
 set(GTEST_INCLUDEDIR "${LLVM_SRC}/utils/unittest/googletest/include")
 set(GTEST_LIBS "-lgtest_main -lgtest")
 
-set(LLVM_CXXFLAGS "${LLVM_CXXFLAGS} -std=c++11")
+set(LLVM_CXXFLAGS "${LLVM_CXXFLAGS} -std=c++17")
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-  set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++11")
+  set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++17")
   set(LLVM_CXXFLAGS "${LLVM_CXXFLAGS} -fvisibility-inlines-hidden")
   set(PASS_LDFLAGS "-Wl,-undefined,dynamic_lookup")
 endif()
 
 include_directories(include third_party/klee/include)
+include_directories(include "${ALIVE_DIR}")
+include_directories(include "${ALIVE_DIR}/alive2")
 
 set(KLEE_EXPR_FILES
   third_party/klee/lib/Expr/Constraints.cpp
@@ -116,6 +122,10 @@ find_library(HIREDIS_LIBRARY
   hiredis
   PATHS
   third_party/hiredis/install/lib)
+
+find_library(ALIVE_LIBRARY alive2 PATHS "${ALIVE_DIR}/build")
+
+find_library(Z3_LIBRARY z3)
 
 set(SOUPER_CLANG_TOOL_FILES
   lib/ClangTool/Actions.cpp
@@ -154,6 +164,8 @@ set(SOUPER_INFER_FILES
   include/souper/Infer/InstSynthesis.h
   lib/Infer/ExhaustiveSynthesis.cpp
   include/souper/Infer/ExhaustiveSynthesis.h
+  lib/Infer/AliveDriver.cpp
+  include/souper/Infer/AliveDriver.h
 )
 
 add_library(souperInfer STATIC
@@ -258,6 +270,7 @@ add_executable(parser_tests
   unittests/Parser/ParserTests.cpp
 )
 
+set(LLVM_LDFLAGS "${LLVM_LDFLAGS} ${ALIVE_LDFLAGS}")
 foreach(target souper internal-solver-test lexer-test parser-test souper-check
                souperExtractor souperInfer souperInst souperKVStore souperParser
                souperSMTLIB2 souperTool souperPass souperPassProfileAll kleeExpr)
@@ -276,20 +289,20 @@ endforeach()
 target_link_libraries(kleeExpr ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperClangTool souperExtractor souperTool ${CLANG_LIBS} ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperExtractor souperInfer souperInst kleeExpr)
-target_link_libraries(souperInfer ${LLVM_LIBS} ${LLVM_LDFLAGS})
+target_link_libraries(souperInfer ${LLVM_LIBS} ${LLVM_LDFLAGS} ${Z3_LIBRARY})
 target_link_libraries(souperInst ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperKVStore ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperParser souperInst ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperSMTLIB2 ${LLVM_LIBS} ${LLVM_LDFLAGS})
 target_link_libraries(souperTool souperExtractor souperSMTLIB2)
-target_link_libraries(souperPass ${PASS_LDFLAGS} ${HIREDIS_LIBRARY})
-target_link_libraries(souperPassProfileAll ${PASS_LDFLAGS} ${HIREDIS_LIBRARY})
-target_link_libraries(souper souperExtractor souperKVStore souperParser souperSMTLIB2 souperTool kleeExpr ${HIREDIS_LIBRARY})
+target_link_libraries(souperPass ${PASS_LDFLAGS} ${HIREDIS_LIBRARY} ${ALIVE_LIBRARY} ${Z3_LIBRARY})
+target_link_libraries(souperPassProfileAll ${PASS_LDFLAGS} ${HIREDIS_LIBRARY} ${ALIVE_LIBRARY} ${Z3_LIBRARY})
+target_link_libraries(souper souperExtractor souperKVStore souperParser souperSMTLIB2 souperTool kleeExpr ${HIREDIS_LIBRARY} ${ALIVE_LIBRARY} ${Z3_LIBRARY})
 target_link_libraries(internal-solver-test souperSMTLIB2)
 target_link_libraries(lexer-test souperParser)
 target_link_libraries(parser-test souperParser)
-target_link_libraries(souper-check souperTool souperExtractor souperKVStore souperSMTLIB2 souperParser ${HIREDIS_LIBRARY})
-target_link_libraries(clang-souper souperClangTool souperExtractor souperKVStore souperParser souperSMTLIB2 souperTool kleeExpr ${CLANG_LIBS} ${LLVM_LIBS} ${LLVM_LDFLAGS} ${HIREDIS_LIBRARY})
+target_link_libraries(souper-check souperTool souperExtractor souperKVStore souperSMTLIB2 souperParser ${HIREDIS_LIBRARY} ${ALIVE_LIBRARY} ${Z3_LIBRARY})
+target_link_libraries(clang-souper souperClangTool souperExtractor souperKVStore souperParser souperSMTLIB2 souperTool kleeExpr ${CLANG_LIBS} ${LLVM_LIBS} ${LLVM_LDFLAGS} ${HIREDIS_LIBRARY} ${ALIVE_LIBRARY} ${Z3_LIBRARY})
 target_link_libraries(extractor_tests souperExtractor ${GTEST_LIBS})
 target_link_libraries(inst_tests souperInst ${GTEST_LIBS})
 target_link_libraries(parser_tests souperParser ${GTEST_LIBS})
--- a/build_deps.sh
+++ b/build_deps.sh
@@ -25,6 +25,14 @@ llvm_branch=tags/RELEASE_700/final
 klee_repo=https://github.com/rsas/klee
 klee_branch=pure-bv-qf-llvm-7.0
 
+alivedir=third_party/alive2
+alive_builddir=$alivedir/build
+mkdir -p $alivedir $alive_builddir
+git clone git@github.com:manasij7479/alive2.git $alivedir/alive2
+(cd $alive_builddir && cmake ../alive2 -GNinja)
+ninja -C $alive_builddir
+
+
 llvm_build_type=Release
 if [ -n "$1" ] ; then
   llvm_build_type="$1"
--- a/include/souper/Infer/AliveDriver.h
+++ b/include/souper/Infer/AliveDriver.h
@@ -1,0 +1,62 @@
+// Copyright 2014 The Souper Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SOUPER_UTIL_ALIVE_DRIVER_H
+#define SOUPER_UTIL_ALIVE_DRIVER_H
+
+#include "souper/Inst/Inst.h"
+#include "alive2/ir/function.h"
+#include "alive2/smt/smt.h"
+
+#include <unordered_map>
+#include <optional>
+
+namespace souper {
+
+// TODO: Rename to AliveBuilder if we implement the ExprBuilder API in future
+class AliveDriver {
+  typedef std::unordered_map<const Inst *, IR::Value *> Cache;
+public:
+  AliveDriver(Inst *LHS_, Inst *PreCondition_);
+
+  std::optional<int64_t> synthesizeConstant(souper::Inst *RHS);
+
+  bool verify(Inst *RHS);
+  ~AliveDriver() {
+    for (auto &&p : TypeCache) {
+      delete(p.second);
+    }
+  }
+private:
+  Inst *LHS, *PreCondition;
+
+  Cache LExprCache, RExprCache;
+
+  std::unordered_map<int, IR::Type*> TypeCache;
+
+  IR::Type &getType(int n);
+
+  bool translateRoot(const Inst *I, const Inst *PC, IR::Function &F, Cache &ExprCache);
+  bool translateAndCache(const Inst *I, IR::Function &F, Cache &ExprCache);
+  IR::Function LHSF;
+
+  int InstNumbers;
+  std::unordered_map<const Inst *, std::string> NamesCache;
+
+  smt::smt_initializer smt_init;
+};
+
+}
+
+#endif
--- a/lib/Infer/AliveDriver.cpp
+++ b/lib/Infer/AliveDriver.cpp
@@ -1,0 +1,400 @@
+#include "souper/Infer/AliveDriver.h"
+
+#include "alive2/ir/function.h"
+#include "alive2/ir/instr.h"
+#include "alive2/ir/state.h"
+#include "alive2/smt/ctx.h"
+#include "alive2/smt/expr.h"
+#include "alive2/smt/smt.h"
+#include "alive2/smt/solver.h"
+#include "alive2/tools/transform.h"
+#include "alive2/util/errors.h"
+#include "alive2/util/symexec.h"
+
+#include "llvm/Support/raw_ostream.h"
+
+#include <memory>
+#include <set>
+#include <string_view>
+#include <z3.h>
+
+namespace {
+class FunctionBuilder {
+public:
+  FunctionBuilder(IR::Function &F_) : F(F_) {}
+
+  template <typename A, typename B, typename ...Others>
+  IR::Value *binOp(IR::Type &t, std::string name, A a, B b,Others... others) {
+    return append
+      (std::make_unique<IR::BinOp>
+      (t, std::move(name), *toValue(t, a), *toValue(t, b), others...));
+  }
+
+  template <typename A>
+  IR::Value *conversionOp(IR::Type &t, std::string name, A a,
+                          IR::ConversionOp::Op Op) {
+    return append
+      (std::make_unique<IR::ConversionOp>
+      (t, std::move(name), *toValue(t, a), Op));
+  }
+
+  template <typename A, typename B, typename C>
+  IR::Value *select(IR::Type &t, std::string name, A a, B b, C c) {
+    return append
+      (std::make_unique<IR::Select>
+      (t, std::move(name), *toValue(t, a), *toValue(t, b), *toValue(t, c)));
+  }
+
+  template <typename A, typename B>
+  IR::Value *iCmp(IR::Type &t, std::string name,
+                  IR::ICmp::Cond cond, A a, B b) {
+    return append
+      (std::make_unique<IR::ICmp>
+      (t, std::move(name), cond, *toValue(t, a), *toValue(t, b)));
+  }
+
+  IR::Value *var(IR::Type &t, std::string name) {
+    return toValue(t, name);
+  }
+
+  IR::Value *val(IR::Type &t, int64_t val) {
+    return toValue(t, val);
+  }
+
+  template<typename T>
+  IR::Value *ret(IR::Type &t, T ret) {
+    return append(
+      std::make_unique<IR::Return>(t, *toValue(t, ret)));
+  }
+
+  template <typename T>
+  void assume(T &&V) {
+    auto AI = std::make_unique<IR::Assume>(*std::move(V));
+    F.getBB("").addIntr(std::move(AI));
+  }
+
+  // Unimplemented : Freeze, CopyOp, Unreachable
+
+private:
+  template <typename T>
+  IR::Value *append(T &&p) {
+    auto ptr = p.get();
+    F.getBB("").addIntr(std::move(p));
+    return ptr;
+  }
+
+  IR::Function &F;
+  std::unordered_map<std::string, IR::Value *> identifiers;
+
+  IR::Value *toValue(IR::Type &t, int64_t x) {
+    auto c = std::make_unique<IR::IntConst>(t, x);
+    auto ptr = c.get();
+    F.addConstant(std::move(c));
+    return ptr;
+  }
+
+  IR::Value *toValue(IR::Type &t, std::string x) {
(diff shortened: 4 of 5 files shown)


CMakeLists.txt
```diff
@@ -67,23 +67,32 @@ execute_process(
...
 )
 
+
 set(CLANG_CXXFLAGS "")
 set(CLANG_INCLUDEDIR "${LLVM_SRC}/tools/clang/include" "${LLVM_BUILD}/tools/clang/include")
+
+set(ALIVE_DIR "${CMAKE_SOURCE_DIR}/third_party/alive2")
+set(ALIVE_LDFLAGS "-L ${ALIVE_DIR}/build -lalive2")
+
+set(CLANG_CXXFLAGS "-I\"${LLVM_SRC}/tools/clang/include\" -I\"${LLVM_BUILD}/tools/clang/include\"")
```
LebedevRI: That is already specified by `CLANG_INCLUDEDIR`.

manasij7479: Thanks for pointing it out, I think I messed up rebasing at some point.

manasij7479: Updated

CMakeLists.txt
```diff
@@ -67,23 +67,30 @@ execute_process(
   OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 
+
```
regehr: plz try not to make random whitespace changes

CMakeLists.txt
```diff
@@ -67,23 +67,30 @@ execute_process(
...
+set(ALIVE_LDFLAGS "-L ${ALIVE_DIR}/build -lalive2")
+
 set(CLANG_LIBS "-lclangCodeGen -lclangTooling -lclangRewrite -lclangFrontend -lclangAnalysis -lclangParse -lclangSerialization -lclangSema -lclangEdit -lclangAnalysis -lclangAST -lclangDriver -lclangLex -lclangBasic")
 
 set(GTEST_CXXFLAGS "-DGTEST_HAS_RTTI=0")
 set(GTEST_INCLUDEDIR "${LLVM_SRC}/utils/unittest/googletest/include")
 set(GTEST_LIBS "-lgtest_main -lgtest")
 
-set(LLVM_CXXFLAGS "${LLVM_CXXFLAGS} -std=c++11")
+set(LLVM_CXXFLAGS "${LLVM_CXXFLAGS} -std=c++17")
```
regehr: does this have any global implications that we need to think more carefully about?

build_deps.sh
```diff
@@ -25,6 +25,14 @@ llvm_branch=tags/RELEASE_700/final
 klee_repo=https://github.com/rsas/klee
 klee_branch=pure-bv-qf-llvm-7.0
 
+alivedir=third_party/alive2
+alive_builddir=$alivedir/build
+mkdir -p $alivedir $alive_builddir
+git clone git@github.com:manasij7479/alive2.git $alivedir/alive2
+(cd $alive_builddir && cmake ../alive2 -GNinja)
+ninja -C $alive_builddir
+
```
regehr: what do you think about also building z3 so we can control the version?

include/souper/Infer/AliveDriver.h
```diff
@@ -0,0 +1,62 @@
...
+private:
+  Inst *LHS, *PreCondition;
+
+  Cache LExprCache, RExprCache;
+
+  std::unordered_map<int, IR::Type*> TypeCache;
+
+  IR::Type& getType(int n);
+
+  bool TranslateRoot(Inst *I, Inst *PC, IR::Function& F, Cache &ExprCache);
```
regehr: please follow LLVM coding conventions: function names start with lowercase
https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+  if (!TranslateRoot(LHS, PreCondition, LHSF, LExprCache)) {
+    llvm::errs() << "Failed to translate LHS.\n";
+    assert(false);
+  } else {
+    LHSFExists = true;
+  }
+}
+
+//TODO: Maybe return an APInt
```
regehr: yes, it's definitely not ok to limit constants to 64 bits

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,411 @@
...
+std::pair<int64_t, bool>
```
regehr: I think you should return a std::optional here instead of a pair

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,411 @@
...
+    }
+  } else {
+    assert(false && "Multiple constants unimplemented.");
+    return {-1, false};
+  }
+}
+
+
+bool souper::AliveDriver::verify (Inst *RHS) {
+  if (!LHSFExists) {
```
regehr: we do this both ways, so it you're free to choose either one, but in some cases like this it may make the code more readable to leave out the curly braces

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -690,10 +750,11 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
         llvm::errs() << "refinement check fails for RHS with no constants\n";
       }
     }
-  }
```
regehr: wait a minute, why are you moving this curly brace?

CMakeLists.txt
manasij7479: Everything seems to work so far and there is a good chance LLVM might switch within the next year.

build_deps.sh
manasij7479: Let's wait until we use more Z3 calls directly in souper.

lib/Infer/AliveDriver.cpp
manasij7479: smt::Model::getInt returns an int64, so the limiting is not being done here.  It might still make sense to use APInt here for future proofing.

lib/Infer/AliveDriver.cpp
manasij7479: Done

lib/Infer/AliveDriver.cpp
manasij7479: It seems I can get a Z3 expr from the model and use its string representation to construct an APInt.
Should this go into Alive instead of Souper?  
I want to avoid using Z3 directly as much as possible here.

regehr: I'm afraid consulting the model directly will not help, I looked into it a bit and the 64-bit limit in alive2 is pervasive and there's no point fixing this until the problems are fixed

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -475,6 +494,47 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
...
+    for (auto &&G : Guesses) {
+      auto C = FindConst(G, "reservedconst_0");
+      if (!C) {
+        auto Result = Verifier.verify(G);
+        if (Result) {
+          RHS = G;
+          return EC;
+        }
+      } else {
+      auto Result = Verifier.synthesizeConstant(G, "%reservedconst_0");
```
regehr: wrong indentation

include/souper/Infer/AliveDriver.h
```diff
@@ -0,0 +1,63 @@
...
+namespace souper {
+
+// TODO: Rename to AliveBuilder if we implement the ExprBuilder API in future
+class AliveDriver {
+  typedef std::unordered_map<Inst *, IR::Value *> Cache;
+public:
+  AliveDriver(Inst *LHS_, Inst *PreCondition_);
+
+  std::optional<int64_t> synthesizeConstant(souper::Inst *RHS,
+                                              std::string ConstName);
```
regehr: indentation is funny here, I suspect you have some tabs in your patch, please be certain they are all converted to spaces

include/souper/Infer/AliveDriver.h
```diff
@@ -0,0 +1,63 @@
...
+private:
+  Inst *LHS, *PreCondition;
+
+  Cache LExprCache, RExprCache;
+
+  std::unordered_map<int, IR::Type*> TypeCache;
+
+  IR::Type& getType(int n);
+
+  bool translateRoot(Inst *I, Inst *PC, IR::Function& F, Cache &ExprCache);
```
regehr: there are lots of by-reference parameters here, can any of these be marked as const?

include/souper/Infer/AliveDriver.h
```diff
@@ -0,0 +1,63 @@
...
+  bool translateAndCache(Inst *I, IR::Function& F, Cache &ExprCache);
+  IR::Function LHSF;
+  bool LHSFExists = false;
```
regehr: would it make sense to replace this pair of member variables with an Optional<IR::Function>?

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+namespace {
+class FunctionBuilder {
+public:
+  FunctionBuilder(IR::Function& F_) : F(F_) {}
+
+  template <typename A, typename B, typename ...Others>
+  IR::Value *BinOp(IR::Type &t, std::string name, A a, B b,Others... others) {
+    return Append
+        (std::make_unique<IR::BinOp>
+         (t, std::move(name), *ToValue(t, a), *ToValue(t, b), others...));
```
regehr: more odd indentation-- do a search for tabs in your patch and eliminate all of them

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+  IR::Value *ToValue(IR::Type& t, int64_t x) {
+    auto c = std::make_unique<IR::IntConst>(t, x);
+    auto ptr = c.get();
+    F.addConstant(std::move(c));
+    return ptr;
+  }
+
+  IR::Value *ToValue(IR::Type& t, std::string x) {
+    if (auto It = identifiers.find(x); It != identifiers.end()) {
```
regehr: a semicolon inside an if condition? what's going on here?

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+    return v;
+  }
+};
+
+
+bool StartsWith(const std::string &pre, const std::string &str) {
+  return std::equal(pre.begin(), pre.end(), str.begin());
+}
+
+void GetReservedConsts(souper::Inst *I, std::set<std::string>& Result) {
```
regehr: put the ``&`` next to result I think

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+  tools::Transform& t;
+  IR::Function &LHS;
+  ~ReturnLHSRAII() {
+    LHS = std::move(t.src);
+  }
+};
+}
+
+util::Errors SynthesizeConstantUsingSolver(tools::Transform &t,
+                                std::string ConstName, int64_t &result) {
```
regehr: weird indentation

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+    return ToValue(t, std::string(x));
+  }
+
+  IR::Value *ToValue(IR::Type& t, IR::Value *v) {
+    return v;
+  }
+};
+
+
+bool StartsWith(const std::string &pre, const std::string &str) {
```
regehr: function names start with lowercase please

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+      err(r, true, "Unknown/Invalid Result, investigate.");
+    }
+  }}});
+
+  return errs;
+}
+
+souper::AliveDriver::AliveDriver(Inst *LHS_, Inst *PreCondition_)
+    : LHS(LHS_), PreCondition(PreCondition_) {
+  InstNumbers = 101;
```
regehr: where does 101 come from? needs a comment.

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+  if (!translateRoot(LHS, PreCondition, LHSF, LExprCache)) {
+    llvm::errs() << "Failed to translate LHS.\n";
+    assert(false);
```
regehr: assertions are eliminated in release builds by default, if you want to kill the tool you must use ``report_fatal_error("message");``

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+  } else {
+    LHSFExists = true;
+  }
+}
+
+//TODO: Maybe return an APInt
+std::optional<int64_t>
+souper::AliveDriver::synthesizeConstant(souper::Inst *RHS,
+                                        std::string ConstName) {
+  if (!LHSFExists)  return std::nullopt;
```
regehr: return goes on a new line

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+    t.tgt = std::move(RHSF);
+    tools::TransformVerify tv(t, /*check_each_var=*/false);
+    int64_t result = 0;
+    if (auto errs = SynthesizeConstantUsingSolver(t, ConstName, result)) {
+      return std::nullopt; // TODO: Encode errs into ErrorCode
+    } else {
+      return {result};
+    }
+  } else {
+    assert(false && "Multiple constants unimplemented.");
```
regehr: this should also be a fatal error I think

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+    return std::nullopt;
+  }
+}
+
+
+bool souper::AliveDriver::verify (Inst *RHS) {
+  if (!LHSFExists) return false;
```
regehr: return on a new line

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+    Builder.Assume(ExprCache[PC]);
+  }
+  Builder.Ret(getType(I->Width), ExprCache[I]);
+  F.setType(getType(I->Width));
+  return true;
+}
+
+
+bool souper::AliveDriver::translateAndCache(souper::Inst *I,
+    IR::Function& F, Cache &ExprCache) {
```
regehr: tabs again?

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+  if (ExprCache.find(I) != ExprCache.end()) {
+    return true; // Already translated
+  }
+
+  FunctionBuilder Builder(F);
+  for (auto &&Op : I->Ops) {
+    if(!translateAndCache(Op, F, ExprCache)) {
```
regehr: space after ``if``

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    case souper::Inst::Select: {
+      ExprCache[I] = Builder.Select(t, Name,
+        ExprCache[I->Ops[0]],
+        ExprCache[I->Ops[1]],
+        ExprCache[I->Ops[2]]);
+    }
+
+    #define BINOP(SOUPER, ALIVE) case souper::Inst::SOUPER: {\
```
regehr: would be best to align the backslashes, I forgot if you're an emacs user but emacs has a quick command for this (so no doubt vim also does)

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -434,6 +438,21 @@ APInt getNextInputVal(Inst *Var,
   llvm::report_fatal_error("Model does not contain the guess input variable");
 }
 
+Inst *FindConst
+    (souper::Inst *I, std::string ConstName) {
```
regehr: don't think you want a newline here?

regehr:
I don't think this patch requires very many test cases, but it would be nice to add one or two tests making sure we get "select" semantics right now -- basically checking UB (we may have an xfail test or two already that you can reuse)

lib/Infer/AliveDriver.cpp
manasij7479: C++17 stuff
https://en.cppreference.com/w/cpp/language/if#If_Statements_with_Initializer

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -475,6 +493,47 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
...
+      } else {
+        auto Result = Verifier.synthesizeConstant(G);
+        // TODO(manasij7479): Extend to multiple constants
+        if (Result.has_value()) {
+          if (C) {
+            std::map<Inst *, llvm::APInt> ConstMap =
+              {{C, llvm::APInt(C->Width, Result.value())}};
+            auto GWithC = getInstCopy(G, IC, InstCache, BlockCache, &ConstMap, false);
+            auto ResultC = Verifier.verify(GWithC);
+            if (ResultC) {
```
regehr: there are a few places like here where you create a temp variable and use it in an if-condition, and then it isn't used anywhere else. this would be clearer if you just put the verify call inside the if-condition.

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -475,6 +493,47 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
...
+        }
+      } else {
+        auto Result = Verifier.synthesizeConstant(G);
+        // TODO(manasij7479): Extend to multiple constants
+        if (Result.has_value()) {
+          if (C) {
+            std::map<Inst *, llvm::APInt> ConstMap =
+              {{C, llvm::APInt(C->Width, Result.value())}};
+            auto GWithC = getInstCopy(G, IC, InstCache, BlockCache, &ConstMap, false);
+            auto ResultC = Verifier.verify(GWithC);
```
regehr: can you explain the circumstances in which a constant synthesized using an exists-forall call will fail to verify?

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -475,6 +493,47 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
...
+      Inst *Eq = IC.getInst(Inst::Eq, 1, {PC.LHS, PC.RHS});
+      Ante = IC.getInst(Inst::And, 1, {Ante, Eq});
+    }
+
+    AliveDriver Verifier(LHS, Ante);
+    for (auto &&G : Guesses) {
+      auto C = findConst(G, "reservedconst_0");
+      if (!C) {
+        auto Result = Verifier.verify(G);
+        if (Result) {
```
regehr: eliminate the temp var

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -434,6 +438,20 @@ APInt getNextInputVal(Inst *Var,
   llvm::report_fatal_error("Model does not contain the guess input variable");
 }
 
+Inst *findConst(souper::Inst *I, std::string ConstName) {
```
regehr: I think this function has exponential runtime, for an expression that has much branching and merging. we've seen this many times and it's something that we need to try hard to avoid. the solution (see many places in souper) is to keep a map of visited instructions.

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,402 @@
...
+    }
+  }
+}
+
+IR::Type& souper::AliveDriver::getType(int n) {
+  if (TypeCache.find(n) != TypeCache.end()) {
+    return *TypeCache[n];
+  } else {
+    TypeCache[n] = new IR::IntType("i" + std::to_string(n), n);
+    return *TypeCache[n];
```
regehr: since this statement is duplicated, place it below the if/else

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,402 @@
...
+    Builder.assume(ExprCache[PC]);
+  }
+  Builder.ret(getType(I->Width), ExprCache[I]);
+  F.setType(getType(I->Width));
+  return true;
+}
+
+
+bool souper::AliveDriver::translateAndCache(const souper::Inst *I,
+                                            IR::Function& F,
```
regehr: please associate the & with the variable, not the type

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,402 @@
...
+  if (auto errs = tv.verify()) {
+    return false; // TODO: Encode errs into ErrorCode
+  } else {
+    llvm::errs() << "RHS proved valid.\n";
+    return true;
+  }
+}
+
+bool souper::AliveDriver::translateRoot(const souper::Inst *I, const Inst *PC,
+    IR::Function& F, Cache &ExprCache) {
```
regehr: this needs to line up with the first "const" on the line above

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,402 @@
...
+bool startsWith(const std::string &pre, const std::string &str) {
+  return std::equal(pre.begin(), pre.end(), str.begin());
+}
+
+void getReservedConsts(souper::Inst *I, std::set<std::string> &Result) {
+  if (startsWith("reservedconst_", I->Name)) {
+    Result.insert(I->Name);
+  }
+  for (auto &&Op : I->Ops) {
+        getReservedConsts(Op, Result);
```
regehr: too much indentation

include/souper/Infer/AliveDriver.h
```diff
@@ -0,0 +1,62 @@
...
+private:
+  Inst *LHS, *PreCondition;
+
+  Cache LExprCache, RExprCache;
+
+  std::unordered_map<int, IR::Type*> TypeCache;
+
+  IR::Type& getType(int n);
+
+  bool translateRoot(const Inst *I, const Inst *PC, IR::Function& F, Cache &ExprCache);
```
regehr: associate & with F, not the type

lib/Infer/ExhaustiveSynthesis.cpp
manasij7479: The UB checks are not bundled with the constant synthesis query.

lib/Infer/ExhaustiveSynthesis.cpp
manasij7479: That would be true if it was accumulating all consts.
This one returns the first one it finds (as we only support synthesizing single constants for now).

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,400 @@
...
+    t.tgt = std::move(RHSF);
+    tools::TransformVerify tv(t, /*check_each_var=*/false);
+    int64_t result = 0;
+    if (auto errs = synthesizeConstantUsingSolver(t, result)) {
+      return std::nullopt; // TODO: Encode errs into ErrorCode
+    } else {
+      return {result};
+    }
+  } else {
+    llvm::errs() << "Multiple constants unimplemented.";
```
manasij7479: This can not be a fatal error as we still want those RHS guesses to be generated for when we support multiple constants.

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,401 @@
...
+  auto &bp = tgt_state.returnVal();
+  auto &a = ap.first;
+  auto &b = bp.first;
+
+  auto qvars = src_state.getQuantVars();
+  qvars.insert(ap.second.begin(), ap.second.end());
+
+  auto err = [&](const smt::Result &r, bool print_var, const char *msg) {
+    tools::error(errs, src_state, tgt_state, r, print_var, nullptr, a, b, msg,
+          false);
```
regehr: this should be even with "errs"

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,401 @@
...
+  };
+
+  std::set<smt::expr> vars;
```
regehr: please follow the LLVM coding style for local variable names https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

lib/Infer/AliveDriver.cpp
```diff
@@ -0,0 +1,401 @@
...
+      auto app = val.first.value.isApp();
+      assert(app);
+      consts.insert(Z3_get_app_arg(smt::ctx(), app, 1));
+    }
+  }
+
+  auto SimpleConstExistsCheck = smt::expr::mkForAll(vars, a.value == b.value);
+
+  smt::Solver::check({{preprocess(t, qvars, ap.second,
+                std::move(SimpleConstExistsCheck)),
```
regehr: indentation

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -434,6 +438,41 @@ APInt getNextInputVal(Inst *Var,
...
 }
 
+Inst *findConst(souper::Inst *I, std::string ConstName,
+                std::set<const Inst *> &Visited) {
+  if (I->Name == ConstName) {
+    return I;
+  } else {
+    for (auto &&Op : I->Ops) {
+      if (Visited.find(Op) == Visited.end()) {
+        auto Ret = findConst(Op, ConstName, Visited);
```
regehr:
```
if (findConst(Op, ConstName, Visited))
  return true;
```

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -475,6 +514,51 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
...
+        if (Verifier.verify(G)) {
+          RHS = G;
+          return EC;
+        }
+      } else {
+        auto Result = Verifier.synthesizeConstant(G);
+        // TODO(manasij7479): Extend to multiple constants
+        // TODO: Counterexample guided loop or UB constraints in query
+        if (Result.has_value()) {
+          if (C) {
```
regehr: execution would not reach this line unless C were true, so this test needs to go away

lib/Infer/ExhaustiveSynthesis.cpp
manasij7479: It needs to return the constant it found recursively, can't shorten to something like that.

regehr: I'm saying get rid of the useless temporary called "Ret"

manasij7479: Ret has 2 uses, in the if condition and then in the return. Can't get rid of it.

regehr: the code you have looks like this:

```
auto Ret = findConst(Op, ConstName, Visited);
if (Ret) {
  return Ret;
}
```

I am asking you to remove the useless temporary, and rewrite the code like this:
```
if (findConst(Op, ConstName, Visited)) {
  return true;
}
```
I can't think of any way to say this more clearly.

manasij7479: The return type is an Inst *, not a bool here.

regehr: ah, carry on then

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -475,6 +513,50 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
...
+      llvm::report_fatal_error("LHS exceeds 64 bits");
+
+    Inst *Ante = IC.getConst(APInt(1, true));
+    for (auto PC : PCs ) {
+      Inst *Eq = IC.getInst(Inst::Eq, 1, {PC.LHS, PC.RHS});
+      Ante = IC.getInst(Inst::And, 1, {Ante, Eq});
+    }
+
+    AliveDriver Verifier(LHS, Ante);
+    for (auto &&G : Guesses) {
```
regehr: this can wait until after merging but please work on a PR that adds debug prints in this loop that are equivalent (in terms of information content and in what gets printed at what debug level) to the existing non-alive guessing loop

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -434,6 +438,40 @@ APInt getNextInputVal(Inst *Var,
...
+    return true;
+  } else {
+    for (auto &&Op : I->Ops) {
+      if (Visited.find(Op) == Visited.end()) {
+        if (exceeds64Bits(Op, Visited)) {
+          return true;
+        }
+      }
+    }
+    Visited.insert(I);
```
regehr: seems inefficient to insert I even if it was already there, is that the behavior you intended here?

regehr: or nevermind, that's not really what it's doing.... I guess this is ok

regehr:
LGTM, thanks for your patience!

PR #411, opened 2019-01-14
Build Release version of alive2

manasij7479:


build_deps.sh
```diff
@@ -29,7 +29,7 @@ alivedir=third_party/alive2
 alive_builddir=$alivedir/build
 mkdir -p $alivedir $alive_builddir
 git clone git@github.com:manasij7479/alive2.git $alivedir/alive2
-(cd $alive_builddir && cmake ../alive2 -GNinja)
+(cd $alive_builddir && cmake ../alive2 -DCMAKE_BUILD_TYPE=$llvm_build_type -GNinja)
```
regehr: this won't work-- lllvm_build_type is set below here

regehr: move this code block below that one

manasij7479:
Fixed

PR #415, opened 2019-01-15
Docker fixes

regehr:
to match recent alive2 work


manasij7479:
Still seems to be failing, what's going on?

LebedevRI:
(You know https://hub.docker.com can build these images with a single button, right?)

regehr:
usually we only have to do this every six months

PR #416, opened 2019-01-15
Synthesize multiple constants

manasij7479:
No tests yet because it is still too slow to synthesize 2 instructions using this backend.


regehr:
LGTM, thanks!

PR #417, opened 2019-01-18
Makefile version of build_deps.sh

manasij7479:


Makefile
```diff
@@ -0,0 +1,117 @@
...
+deps : alive2 llvm hiredis klee
+
+llvm : third_party/llvm llvm-build
+# 	we want these but they don't get installed by default
+	cp $(llvm_builddir)/bin/llvm-lit $(llvm_installdir)/bin/
+	cp $(llvm_builddir)/bin/FileCheck $(llvm_installdir)/bin/
+	cp $(llvm_builddir)/lib/libgtest_main.a $(llvm_installdir)/lib/
+	cp $(llvm_builddir)/lib/libgtest.a $(llvm_installdir)/lib/
+
+llvm-build:
```
pranavk: This probably needs to be .PHONY to keep `llvm-build` target always out-of-date.

Makefile
```diff
@@ -0,0 +1,117 @@
...
+	mkdir -p $(llvm_builddir)
+	
+	if [ -n "`which ninja`" ]; then\
+		$(NINJA) -C $(llvm_builddir);\
+		$(NINJA) -C $(llvm_builddir) install;\
+	else\
+		make -C $(llvm_builddir) -j4;\
+		make -C $(llvm_builddir) -j4 install;\
+	fi
+alive2 : third_party/alive2
```
pranavk: Same .PHONY thing as above. There are other targets in this file as well which are not marked as .PHONY. I guess the problem will start appearing when there are files in the filesystem with exact same name as these targets. `make` won't build these targets because it will think that these targets are already built (even though you want them to be built unconditionally everytime).

Makefile
```diff
@@ -0,0 +1,117 @@
...
+llvm_installdir=$(llvmdir)/$(llvm_build_type)
+llvm_builddir=$(llvmdir)/${llvm_build_type}-build
+cmake_flags=-DCMAKE_INSTALL_PREFIX="../../../$(llvm_installdir)" -DLLVM_ENABLE_ASSERTIONS=On -DLLVM_TARGETS_TO_BUILD=host -DCMAKE_BUILD_TYPE="$(llvm_build_type)"
+Build=Release
+# $(Build) will not be set here if assigned through the command line
+# TODO: Make the targets build-type specific, currently you have to nuke third_party for changing build type
+NINJA=ninja
+# Override NINJA from the command line with something like NINJA="ninja -j1" if you are running out of memory
+
+deps : alive2 llvm hiredis klee
```
pranavk: nitpick: let's be consistent and not have any `space` before the colon.

regehr:
I'm still not sure what problem this change is intended to solve

regehr:
also, if we're going to do this, I'd like the full set of dependencies encoded in the makefile
for example, if we upgrade souper and it requires a different version of LLVM, then that should trigger a rebuild of LLVM

regehr:
and then this all needs to be tested

manasij7479:
> I'm still not sure what problem this change is intended to solve

My reason: If llvm build is interrupted for any reason, it is annoying to nuke third_party and start from scratch.   This makefile cleanly splits up cloning+cmake and the build steps, so that the second one can be resumed without redoing the first.    
    
> full set of dependencies encoded in the makefile  
    
Will do.  
Maybe put the LLVM version/url information in a file and use that as a dependency?   
Is there a better solution? (cc: @pranavk )

Makefile
```diff
@@ -0,0 +1,117 @@
...
+
+llvm : third_party/llvm llvm-build
```
pranavk: we shouldn't rely on the order of dependencies here. if you want llvm-build to be executed after third_party/llvm, then i think better to make third_party/llvm a dependency of llvm-build. Parallel make can mess things up otherwise. See: https://stackoverflow.com/questions/9159960/order-of-processing-components-in-makefile/9160175#9160175

pranavk:
> Will do.
> Maybe put the LLVM version/url information in a file and use that as a dependency?
> Is there a better solution? (cc: @pranavk )

Using the llvm version info as name of the file (eg: llvm70)  and `touch` it in llvm directory sounds like should work. When souper depends on a new llvm version, it will try to find a `llvm80` in third_party/llvm/ which it won't find in which case we will trigger the llvm-build.

regehr:
sounds good-- I want you to test this though!

manasij7479:
@LebedevRI Do you have any suggestions for improving this Makefile?

Makefile
```diff
@@ -0,0 +1,117 @@
...
+	cp $(llvm_builddir)/lib/libgtest_main.a $(llvm_installdir)/lib/
+	cp $(llvm_builddir)/lib/libgtest.a $(llvm_installdir)/lib/
+
+llvm-build:
+	@echo "Building LLVM"
+
+	mkdir -p $(llvm_builddir)
+	
+	if [ -n "`which ninja`" ]; then\
+		$(NINJA) -C $(llvm_builddir);\
```
LebedevRI: Do you need something like `set -e`, so we abort on the first failure?

pranavk:
Just had a OOM while building debug build of souper (LLVM was being built). This PR would help greatly because I had to hack the current build_deps.sh script so that it doesn't start compiling LLVM from scratch.

regehr:
@pranavk what would happen if I was in the middle of cloning LLVM and my connection got interrupted-- would this makefile gracefully recover from that or would it simply see that the directory exists and move forward, failing in a perhaps-confusing way?

pranavk:
i am not sure about this makefile; i don't think so. i guess manasij is working on it and will update this PR soon; putting those .PHONY attributes to make sure that we unconditionally process some targets should fix the issue. i'll have a look and test it after he updates this PR.

manasij7479:
Wouldn't unconditionally processing the target mean LLVM is cloned even if
it is already cloned?
I can imagine another hacky solution: create a dummy file iff git clone
exists successfully. Deleting that file would be a way to force a refectch.

pranavk:
how about an 'if' condition in the target that checks if .git/ directory exists. If (exists), git pull, otherwise clone? Or something along those lines.

regehr:
it's clear that all of these issues can be solved
the question is whether the solutions are worth the complexity and whether you guys can spend your time a better way....

regehr:
I'm just saying that the thing being solved here isn't something that has caused me pain over multiple years of being a souper developer, and it's very likely to not cause you guys much pain either, once you get your machine issues straightened out

regehr:
and also there are plenty of corner cases where you can easily make things worse, whereas "nuke it all and rebuild" is very easy to think about

PR #419, opened 2019-01-22
Do not list x.with.overflow as synthesis components in exhaustive syn…

zhengyangl:
…thesis and nop synthesis


regehr:
please remove the FilterOverflowInsts flag (always filter these) and add a test

regehr:
LGTM, thanks!

PR #420, opened 2019-01-22
Build z3 into souper

regehr:
Not sure this is ready for prime time but @rsas here's a quick patch for building z3 into souper that makes Alive and everything else happy.

### DIFF
--- a/build_deps.sh
+++ b/build_deps.sh
@@ -25,6 +25,8 @@ llvm_branch=tags/RELEASE_700/final
 klee_repo=https://github.com/rsas/klee
 klee_branch=pure-bv-qf-llvm-7.0
 alive_repo=https://github.com/manasij7479/alive2.git
+z3_repo=https://github.com/Z3Prover/z3.git
+z3_branch=z3-4.8.4
 
 llvm_build_type=Release
 if [ -n "$1" ] ; then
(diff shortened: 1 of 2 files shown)


regehr:
I don't know cmake very well but this feels like overkill

manasij7479:
I think this works:
find_library(Z3_LIBRARY z3 PATHS "${z3_installdir}")

regehr:
I think it has to be you (as the maintainer of our alive2 branch) that does this, right?

the horrible hack in question (hard-coded .so) is in a shell script, not a cmake file, so I can't solve the problem using cmake code

manasij7479:
Ok, will do.
LGTM.

build_deps.sh
```diff
@@ -25,39 +25,52 @@ llvm_branch=tags/RELEASE_700/final
...
 
 if [ -n "`which ninja`" ] ; then
-  (cd $alive_builddir && cmake ../alive2 -DCMAKE_BUILD_TYPE=$llvm_build_type -GNinja)
+  (cd $alive_builddir && cmake ../alive2 -DZ3_LIBRARIES=$z3_installdir/lib/libz3.so -DZ3_INCLUDE_DIR=$z3_installdir/include -DCMAKE_BUILD_TYPE=$llvm_build_type -GNinja)
   ninja -C $alive_builddir
 else
-  (cd $alive_builddir && cmake ../alive2 -DCMAKE_BUILD_TYPE=$llvm_build_type)
-  make -C $alive_builddir -j4
+  (cd $alive_builddir && cmake ../alive2 -DZ3_LIBRARIES=$z3_installdir/lib/libz3.so -DZ3_INCLUDE_DIR=$z3_installdir/include -DCMAKE_BUILD_TYPE=$llvm_build_type)
+  make -C $alive_builddir -j8
```
LebedevRI: Hm, why is this using make instead of ninja, and thus makes a guess at the appropriate thread count?
This looks like a bug.

build_deps.sh
```diff
@@ -25,39 +25,52 @@ llvm_branch=tags/RELEASE_700/final
...
 fi
 
+z3_srcdir=$(pwd)/third_party/z3
+z3_installdir=$(pwd)/third_party/z3-install
+git clone $z3_repo $z3_srcdir
+mkdir -p $z3_installdir
+
+(cd $z3_srcdir && git checkout $z3_branch &&
+python scripts/mk_make.py --noomp --prefix=$z3_installdir &&
+cd build && make -j8 install)
```
LebedevRI: z3's cmake build system worked fine for me
(and thus, the same comment re ninja vs make)

build_deps.sh
```diff
@@ -25,39 +25,52 @@ llvm_branch=tags/RELEASE_700/final
 klee_repo=https://github.com/rsas/klee
 klee_branch=pure-bv-qf-llvm-7.0
 alive_repo=https://github.com/manasij7479/alive2.git
+z3_repo=https://github.com/Z3Prover/z3.git
+z3_branch=z3-4.8.4
```
pranavk: i guess we should change the `Dockerfile` as well to reflect that 4.8.4 is needed. Current Dockerfile builds 4.7.1 which alive2 doesn't like.

pranavk: never mind, i was looking at older version. you already fixed it

regehr:
ok I think I'm in favor of merging this PR after Manasij maybe improves Alive2's z3 detection or else figures out how to avoid the gross hack here.

also it may well be time to require ninja. earlier in souper's history this was a bad idea since OS package managers either lacked it or shipped broken old versions.

rsas:
Still does does not work, at least for me:

```
...
cp libz3.dylib python
Z3 was successfully installed.
Cloning into 'third_party/alive2/alive2'...
remote: Enumerating objects: 6, done.
remote: Counting objects: 100% (6/6), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 825 (delta 0), reused 2 (delta 0), pack-reused 819
Receiving objects: 100% (825/825), 212.16 KiB | 848.00 KiB/s, done.
Resolving deltas: 100% (548/548), done.
-- The C compiler identification is AppleClang 10.0.0.10001145
-- The CXX compiler identification is AppleClang 10.0.0.10001145
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Build type: Release
-- Z3: /Users/rsas/work/souper-rsas/third_party/z3-install/include /Users/rsas/work/souper-rsas/third_party/z3-install/lib/libz3.so
-- Z3 installed version: 4.8.4
-- RE2C: RE2C-NOTFOUND
-- Skiping translation validation plugin
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/rsas/work/souper-rsas/third_party/alive2/build
ninja: Entering directory `third_party/alive2/build'
ninja: error: '/Users/rsas/work/souper-rsas/third_party/z3-install/lib/libz3.so', needed by 'alive', missing and no known rule to make it
```

rsas:
It is ```libz3.dylib``` in my case on osx, can we improve the hard-coded library name for alive somehow?

regehr:
> It is `libz3.dylib` in my case on osx, can we improve the hard-coded library name for alive somehow?

I'm not very good at cmake and was hoping one of my people would fix this, can you take a look @pranavk or @manasij7479 ?

pranavk:
how about just writing a small bash function that finds the library prefixes and suffixes after probing the host system? i guess that's what cmake's find_library does too. Having said that, I also think that proper solution might be to use cmake_packages as Manasij mentioned earlier but might be bit more work than just writing a small shell script function.

regehr:
I would hope that cmake has a canned solution to this problem, which must occur anytime someone wants to portably deal with shared library across linux and osx. it seems implausible that this problem has not been solved. please verify that before looking at interesting solutions.

if there indeed does not exist a solution to reuse, then implement whatever seems most pleasing and maintainable

PR #421, opened 2019-01-23
Init doxygen docs

pranavk:
`make doc` to generate HTML documentation


pranavk:
Doxygen is not added as hard dependency; so, if you don't have it installed, we won't complain. If it is installed, you are supposed to do `make doc` explicitly to generate docs; it's not added to the default `all` target.

regehr:
can you make it just build the docs if doxygen is found, and not add an extra cmake option?

pranavk:
Yes, I can merge the 'doc' target with 'all' so docs are always generated whenever doxygen is found but I thought we don't want to clutter build tree for users who don't want it; so, I didn't.

regehr:
I think it's better to clutter the build tree than to make configuration more complicated

regehr:
please provide a link to the produced documentation, I want to look at it before merging this

pranavk:
http://www.cs.utah.edu/~pranavk/souperdocs/

pranavk:
> I think it's better to clutter the build tree than to make configuration more complicated

My reason behind not cluttering was not to add to the build time. Those who think it's worth to generate the documentation, they can manually do it. 

Also, I don't think configuration is complicated, especially when I have a new CMakeLists.txt in the docs/ directory (separate from the main CMakeLists.txt file of the whole project). It's not cluttering the main CMakeLists.txt file.

regehr:
how much build time is added by doxygen?

regehr:
also the top doxygen page is all blank and it's hard to tell where to click, before this gets merged I'd like some added content

pranavk:
> how much build time is added by doxygen?

Actually not much as of now. ~1 second. 

> also the top doxygen page is all blank and it's hard to tell where to click, before this gets merged I'd like some added content

The LLVM Doxygen (http://llvm.org/doxygen/) also doesn't put anything that will teach users how to navigate the documentation. Couple of other projects I know also doesn't care about it. I think people who are used to using doxygen-based documentation will already know where to go (Classes -> Index, for example).

We can get some content added but I didn't feel initially that it's worth investing time (after all, it's going to be used by developers only).

regehr:
Ok, please make doxygen generation conditional only on presence of doxygen, and then this looks good!

pranavk:
Okay. If I just make the `doc` target depend on `ALL` that triggers doc generation *even when* nothing is built (i.e. when make does nothing). I need to figure out a way to *only* trigger this when there's something to do.

regehr:
the doc target should depend on all source and header files, right?

regehr:
LGTM please squash

pranavk:
done

PR #422, opened 2019-01-23
souperweb: be more informative; print port on console

pranavk:
just doing trivial fixes as i explore the codebase


regehr:
LGTM thanks

PR #435, opened 2019-02-06
Support for using alive backend in solver tests

manasij7479:
Don't merge until all tests pass.

### DIFF
--- a/lib/Extractor/Solver.cpp
+++ b/lib/Extractor/Solver.cpp
@@ -21,6 +21,7 @@
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/ErrorHandling.h"
 #include "souper/Extractor/Solver.h"
+#include "souper/Infer/AliveDriver.h"
 #include "souper/Infer/ExhaustiveSynthesis.h"
 #include "souper/Infer/InstSynthesis.h"
 #include "souper/KVStore/KVStore.h"
@@ -95,6 +96,17 @@ public:
         Guesses.emplace_back(IC.getConst(APInt(LHS->Width, -1)));
       for (auto I : Guesses) {
         InstMapping Mapping(LHS, I);
+
+        if (UseAlive) {
+          bool IsValid = isTransformationValid(Mapping.LHS, Mapping.RHS,
+                                               PCs, IC);
+          if (IsValid) {
+            RHS = I;
+            return std::error_code();
+          }
+          // TODO: Propagate errors from Alive backend, exit early for errors
+        }
+
         std::string Query = BuildQuery(IC, BPCs, PCs, Mapping, 0);
         if (Query.empty())
           return std::make_error_code(std::errc::value_too_large);
@@ -114,6 +126,16 @@ public:
       std::vector<llvm::APInt> ModelVals;
       Inst *I = IC.createVar(LHS->Width, "constant");
       InstMapping Mapping(LHS, I);
+
+      if (UseAlive) {
+        bool IsValid = isTransformationValid(Mapping.LHS, Mapping.RHS, PCs, IC);
+        if (IsValid) {
+          RHS = I;
+          return std::error_code();
+        }
+        // TODO: Propagate errors from Alive backend, exit early for errors
+      }
+
       std::string Query = BuildQuery(IC, BPCs, PCs, Mapping, &ModelInsts, /*Negate=*/true);
       if (Query.empty())
         return std::make_error_code(std::errc::value_too_large);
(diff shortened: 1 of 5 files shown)


lib/Extractor/Solver.cpp
```diff
@@ -61,6 +62,9 @@ static cl::opt<bool> InferInsts("souper-infer-inst",
 static cl::opt<bool> EnableExhaustiveSynthesis("souper-exhaustive-synthesis",
     cl::desc("Use exaustive search for instruction synthesis (default=false)"),
     cl::init(false));
+static cl::opt<bool> EnableAliveTest("souper-solver-alive",
+    cl::desc("Use alive for solver tests (default=false)"),
```
regehr: don't say "for solver tests" since there's nothing specific about testing here

let's have one alive command line option, don't have one here and one in synthesis

lib/Extractor/Solver.cpp
```diff
@@ -316,8 +320,18 @@ class MemCachingSolver : public Solver {
                           std::vector<std::pair<Inst *, llvm::APInt>> *Model)
     override {
     // TODO: add caching support for models.
-    if (Model)
+    if (Model) {
```
regehr: this is the memory caching layer, I don't think alive code wants to go here

manasij7479: This is the solver call invoked by souper-check though.
Hooking up alive here invokes it fine.
What am I missing?

regehr: if it's connected with alive it goes in the base isValid call, not in the caching layers

lib/Infer/AliveDriver.cpp
```diff
@@ -70,7 +70,7 @@ class FunctionBuilder {
   template <typename T>
   void assume(T &&V) {
     auto AI = std::make_unique<IR::Assume>(*std::move(V));
-    F.getBB("").addIntr(std::move(AI));
+    F.getBB("").addInstr(std::move(AI));
```
regehr: you probably should rebase your branch since you made these changes in a different PR, right?

lib/Extractor/Solver.cpp
regehr: ok there are three isValid() implementations in this file, right?
the top one is where stuff happens, that's the one you need to modify
the other two are in the caching layers, they absolutely do not talk to alive
not sure what else I can say here...

regehr: it's irrelevant if this code works, it's completely in the wrong place

lib/Extractor/Solver.cpp
regehr: I mean, the "souper-use-alive" flag should be used here, please do not add another flag for this

lib/Extractor/Solver.cpp
```diff
@@ -95,6 +99,20 @@ class BaseSolver : public Solver {
...
+          Inst *Ante = IC.getConst(APInt(1, true));
+          for (auto PC : PCs ) {
+            Inst *Eq = IC.getInst(Inst::Eq, 1, {PC.LHS, PC.RHS});
+            Ante = IC.getInst(Inst::And, 1, {Ante, Eq});
+          }
+          bool IsValid = isTransformationValid(Mapping.LHS, Mapping.RHS, Ante);
+          if (IsValid) {
+            RHS = I;
+          }
+          return std::error_code(); // TODO: Propagate errors from Alive backend
```
regehr: don't unconditionally return here, this piece only synthesized 0, 1, -1
we still want to synthesize arbitrary integers

lib/Extractor/Solver.cpp
```diff
@@ -114,6 +132,20 @@ class BaseSolver : public Solver {
...
+        Inst *Ante = IC.getConst(APInt(1, true));
+        for (auto PC : PCs ) {
+          Inst *Eq = IC.getInst(Inst::Eq, 1, {PC.LHS, PC.RHS});
+          Ante = IC.getInst(Inst::And, 1, {Ante, Eq});
+        }
+        bool IsValid = isTransformationValid(Mapping.LHS, Mapping.RHS, Ante);
+        if (IsValid) {
+          RHS = I;
+        }
+        return std::error_code(); // TODO: Propagate errors from Alive backend
```
regehr: same here, don't just return. if synthesizing an integer fails, we still want to fall down to exhaustive syntheiss (or CEGIS or whatever)

regehr:
ok great-- this is ready to merge as far as you're concerned @manasij7479 ?

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -51,9 +53,9 @@ namespace {
   static cl::opt<bool> EnableBigQuery("souper-exhaustive-synthesis-enable-big-query",
     cl::desc("Enable big query in exhaustive synthesis (default=false)"),
     cl::init(false));
-  static cl::opt<bool> UseAlive("souper-use-alive",
-    cl::desc("Use Alive2 as the backend"),
-    cl::init(false));
+  static cl::opt<bool, true>
```
regehr: please add a /* foo=*/ comment here saying what the true is for

manasij7479: Done

manasij7479:
Yes, feel free to merge this.
Lots of issues to fix before we can enable it by default though.

regehr:
LGTM thanks!

PR #437, opened 2019-02-06
Filter out guesses which differ in LSB

manasij7479:


regehr:
I'd like you to change build_deps a bit so that it checks out a specific revision of alive2 so we don't have these spurious test failures

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -731,6 +758,16 @@ ExhaustiveSynthesis::synthesize(SMTLIBSolver *SMTSolver,
     separatePCs(PCs, PCsCopy, InstCache, BlockCache, IC, &ConstMap, false);
 
     InstMapping Mapping2(LHS, I2);
+
+    if (LSBPruning && !GuessHasConstant) {
```
regehr: this idea should still work fine when we have synthesized constants on the RHS, right?

manasij7479: Yes, let's do that with another PR, added a TODO.

regehr: actually I don't think it's as easy as I thought-- so definitely save for later

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -472,6 +475,30 @@ bool exceeds64Bits(const Inst *I, std::set<const Inst *> &Visited) {
...
+
+  InstMapping NewMapping{NewLHS, NewRHS};
+
+  auto Query = BuildQuery(IC, BPCs, PCs, NewMapping, 0);
+
+  bool QueryIsSat;
+  auto EC = SMTSolver->isSatisfiable(Query, QueryIsSat, 0, 0, Timeout);
+  if (EC) {
+    if (DebugLevel > 1)
+      llvm::errs() << "error!\n";
```
regehr: should say something more interesting here, and probably at all debug levels

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -54,6 +54,9 @@ namespace {
   static cl::opt<bool> UseAlive("souper-use-alive",
     cl::desc("Use Alive2 as the backend"),
     cl::init(false));
+  static cl::opt<bool> LSBPruning("souper-lsb-pruning",
+    cl::desc("Try to prune guesses by looking for a difference in LSB"),
+    cl::init(true));
```
regehr: I'd say this is off by default until we've done some more extensive testing for effect on performance

lib/Infer/ExhaustiveSynthesis.cpp
manasij7479: Expanded the error message, note that just "error!\n" is used as an error message in multiple places in  this file.
As for all debug levels: probably not necessary because if the query fails, nothing bad happens; it just fails to prune early.

regehr:
LGTM, thanks

PR #441, opened 2019-02-11
WIP: Initial version of the z3 expression builder

rsas:
I clean version of the z3builder, easier to review.


rsas:
``make check`` fails because the ``llvm-lit`` (Python) cannot find ``libz3.dylib``. Python unsets the current ``LD_LIBRARY_PATH`` (on Darwin: ``DYLD_LIBRARY_PATH``) making bash exports useless. Any ideas how to fix this? Is there a cmake hack to tell the linker to use the absolute path the the z3 library?

LebedevRI:
Random thoughts: https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling

PR #445, opened 2019-02-12
Fixed a bug in harvest uses which potentially hangs the clang

zhengyangl:


manasij7479:
LGTM

regehr:
is this ready for review?

zhengyangl:
Not yet, will be ready tonight.

lib/Inst/Inst.cpp
```diff
@@ -77,6 +77,10 @@ bool Inst::operator<(const Inst &Other) const {
     return (*OpsA[I] < *OpsB[I]);
   }
 
+  if (HarvestKind != Other.HarvestKind)
+    return true;
```
regehr: does this work? this just them them compare less than, but then the opposite comparison will also be less than.
don't you need to return something like (int)HarvestKind) < (int)Other.HarvestKind?

lib/Inst/Inst.cpp
```diff
@@ -509,6 +513,8 @@ void Inst::Profile(llvm::FoldingSetNodeID &ID) const {
   default:
     if (!DemandedBits.isAllOnesValue())
       ID.Add(DemandedBits);
+    if (HarvestFrom)
+      ID.Add(HarvestFrom);
```
regehr: what about HarvestKind?

lib/Pass/Pass.cpp
```diff
@@ -386,102 +386,109 @@ struct SouperPass : public ModulePass {
         }
         AddToCandidateMap(CandMap, R);
       }
+    }
 
 
-      for (auto &Cand : CandMap) {
+    for (auto &Cand : CandMap) {
```
zhengyangl: After the block information is attached to the LHS, the for loop in Pass.cpp now gets reverted. Now it first collects all the candidates at one time, instead of collects the candidate by each block.

lib/Inst/Inst.cpp
zhengyangl: Fixed in the new commit.

lib/Inst/Inst.cpp
zhengyangl: Fixed in the new commit

regehr:
LGTM, thanks!
please squash

PR #449, opened 2019-02-14
Refactor ExhaustiveSynthesis into smaller functions

manasij7479:
The goto is still there, it looks cleaner than a while loop. :(


regehr:
LGTM thanks

PR #453, opened 2019-02-16
Fixed the operator precedence bug

zhengyangl:


lib/Pass/Pass.cpp
```diff
@@ -427,7 +427,7 @@ struct SouperPass : public ModulePass {
                                ReplacedValues, Builder, F->getParent());
 
       // if LHS comes from use, then NewVal should be a constant
-      assert(!Cand.Mapping.LHS->HarvestKind == HarvestType::HarvestedFromUse ||
+      assert(!(Cand.Mapping.LHS->HarvestKind == HarvestType::HarvestedFromUse) ||
```
regehr: please use != instead

PR #456, opened 2019-02-17
fix misnamed function

regehr:


manasij7479:
https://github.com/google/souper/pull/458
This PR uses the function mentioned here.
If you merge #456 before it, can you also make this function public?
That would make #458 a little bit more concise.

manasij7479:
Also, is it possible to make this function take const Inst *?

regehr:
oops I merged before seeing your second comment-- please make the const change as part of your PR

PR #460, opened 2019-02-20
Introduce pruning in the constant synthesis loop

manasij7479:


regehr:
ok it looks good but something like this really wants a test case or two, please add

PR #462, opened 2019-02-21
debug: print translated alive function as debug output

pranavk:


pranavk:
was playing with alive backend. Wanted to print transformed alive ir too; there wasn't anything existing.

regehr:
LGTM

PR #463, opened 2019-02-21
Support saturating instructions

pranavk:
This is a draft pull request. Work still in progress. I need to add more tests. Right now only one small test to test the basic functionality (which works).

### DIFF
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -508,6 +508,7 @@ Inst *ExprBuilder::buildHelper(Value *V) {
     LibFunc Func;
     if (auto II = dyn_cast<IntrinsicInst>(Call)) {
       Inst *L = get(II->getOperand(0));
+      Inst *R = get(II->getOperand(1));
       switch (II->getIntrinsicID()) {
         default:
           break;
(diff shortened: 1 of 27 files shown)


regehr:
great!
I don't want to merge until llvm 8 is really out
if you have time in the meantime, add the other saturating instructions
I'll look over your code now

lib/Extractor/Candidates.cpp
```diff
@@ -563,6 +563,11 @@ Inst *ExprBuilder::buildHelper(Value *V) {
           Inst *Overflow = IC.getInst(Inst::UMulO, 1, {L, R}, /*Available=*/false);
           return IC.getInst(Inst::UMulWithOverflow, L->Width+1, {Mul, Overflow});
         }
+        case Intrinsic::sadd_sat: {
+	  // TODO: Factor out *R
+	  Inst *R = get(II->getOperand(1));
+	  return IC.getInst(Inst::SAddSat, L->Width, {L, R}, /*Available=*/false);
```
regehr: why false for available?

regehr:
ok code looks great except:
- thing I noted
- we require parser support
- need quite a few more tests, think like Roman

regehr:
ok I see you made a change to the parser -- is this all that's required?

lib/Extractor/Candidates.cpp
pranavk: i am not sure really yet how it affects things. Yeah, from the pattern of other instructions above, it should be true (which is also the default if i don't mention it).

pranavk:
the one test that i added passed; so, i guess this is all that's required in the parser to make it work. i'll fix things as i notice them while adding newer tests

regehr:
look at @LebedevRI's tests for the funnel shift patch
your tests do not need to be as extensive as his (where were excellent) but you should cover that same sorts of things

pranavk:
added rest of the instructions; will add tests soon too

lib/Extractor/Candidates.cpp
regehr: should be true

pranavk: this is outdated. i have already fixed that.

regehr:
ready for review?

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -366,6 +366,50 @@ class KLEEBuilder : public ExprBuilder {
...
+      auto smaxExt = SExtExpr::create(smax, I->Width + 1);
+      auto pred = SleExpr::create(addExt, sminExt);
+
+      auto pred2 = SgeExpr::create(addExt, smaxExt);
+      auto select2 = SelectExpr::create(pred2, smax, add);
+
+      return SelectExpr::create(pred, smin, select2);
+    }
+    case Inst::UAddSat:
+    {
```
regehr: prefer putting this kind of curly brace on the same line as the case
(don't worry about it for this PR but do it generally)

regehr:
ok this seems almost ready to go

I want at least one test case which synthesizes a saturating operation

I want at least one test case which reasons about saturating arithmetic symbolically, for example use a path condition to express that an addition cannot overflow, in which case you should show that a saturating subtract can be replaced by a subnsw

pranavk:
ok. will add these tests and update you when it's ready

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -367,6 +367,46 @@ class KLEEBuilder : public ExprBuilder {
...
+      ref<Expr> add = AddExpr::create(get(Ops[0]), get(Ops[1]));
+      auto sextL = SExtExpr::create(get(Ops[0]), I->Width + 1);
+      auto sextR = SExtExpr::create(get(Ops[1]), I->Width + 1);
+      auto addExt = AddExpr::create(sextL, sextR);
+      auto smin = klee::ConstantExpr::alloc(llvm::APInt::getSignedMinValue(I->Width));
+      auto smax = klee::ConstantExpr::alloc(llvm::APInt::getSignedMaxValue(I->Width));
+      auto sminExt = SExtExpr::create(smin, I->Width + 1);
+      auto smaxExt = SExtExpr::create(smax, I->Width + 1);
+      auto pred = SleExpr::create(addExt, sminExt);
+
```
regehr: I feel like this is a pretty complicated way to detect signed overflow. why not reuse our signed overflow code or else just look for the result having a different sign than the inputs (you don't need to do any sign extension in that case)

test/Pass/syn-inst-sadd-sat.ll
```diff
@@ -0,0 +1,19 @@
+; REQUIRES: solver
```
regehr: you should try to add diversity in test cases
if the souper test used one way to detect overflow, use a different one here
if the souper test is 8 bits, make this one 16 or 32
etc.
you're trying to dig up bugs here, doing different things in different ways is part of the goal

test/Tool/saturating1.opt
```diff
@@ -0,0 +1,10 @@
+; REQUIRES: solver
+
+; RUN: %souper-check %solver -infer-rhs -print-replacement %s > %t1
+; RUN: %FileCheck %s -check-prefix=SUCCESS < %t1
+
+; SUCCESS: RHS inferred successfully
```
regehr: do you want to look for a particular RHS? or are you just testing that souper thinks it has succeeded here?

lib/Extractor/KLEEBuilder.cpp
pranavk: this (and others) is taken verbatim from alive2 to keep it consistent with it; I didn't want to come up with any clever way for this (with potential risks of getting wrong). Moreover, we want to get rid of KLEE anyways according to you; so, i didn't want to invest much time on it.

test/Tool/saturating1.opt
pranavk: yeah, this test is silly. I will remove it. I think we have enough tests for this one now.

pranavk: fixed

lib/Extractor/KLEEBuilder.cpp
regehr: ya, ok ok :)

regehr:
LGTM, thanks!

regehr:
do you know what's going on with the test failures?

pranavk:
i have seen this KLEE buidler recursively calling itself repeateadly in these tests before. So this should be unrelated to this PR.

pranavk:
ok. these tests fails on my local even with this PR reverted. So this is unrelated.

LebedevRI:
(That's just #400)

regehr:
this is a problem that'll go away once we move to alive!
not sure what to do about it right now, ugh. I'll merge this though.

regehr:
@LebedevRI but I don't see why it's showing up in CI suddenly, ugh....

PR #465, opened 2019-02-21
Update some instructions to reflect new LLVM instrinsics semantics

pranavk:


pranavk:
I think this can be improved a bit too. I will do that soon (tomorrow) : https://github.com/google/souper/blob/master/lib/Extractor/KLEEBuilder.cpp#L292-L313

pranavk:
and will try adding a test or two supporting above patches

lib/Parser/Parser.cpp
```diff
@@ -725,18 +725,8 @@ bool Parser::typeCheckInst(Inst::Kind IK, unsigned &Width,
 
   switch (IK) {
```
regehr: this should not be a switch anymore, but just an "if"

lib/Parser/Parser.cpp
```diff
@@ -725,18 +725,8 @@ bool Parser::typeCheckInst(Inst::Kind IK, unsigned &Width,
...
-    break;
-  case Inst::CtPop:
-  case Inst::Ctlz:
-  case Inst::Cttz:
-    if (Width !=8 && Width != 16 && Width != 32 &&
-        Width != 64 && Width != 256) {
-      ErrStr = std::string(Inst::getKindName(IK)) + " doesn't support " +
-        std::to_string(Width) + " bits";
+    if (Width % 16 != 0) {
+      ErrStr = "bswap doesn't support odd number of bytes";
```
regehr: a bit more informative would be "bswap argument must be a multiple of 16 bits"

also make sure it's >0 unless that's done elsewhere

pranavk: yup. > 0 check is already there just above this switch block

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -82,7 +82,7 @@ class KLEEBuilder : public ExprBuilder {
       Printer.setArrayValuesToGet(Arr);
     }
     Printer.generateOutput();
-  
+
```
regehr: if you want to make whitespace changes, these should go into a separate pull req

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -291,28 +291,13 @@ class KLEEBuilder : public ExprBuilder {
       return countOnes(get(Ops[0]));
     case Inst::BSwap: {
       ref<Expr> L = get(Ops[0]);
-      unsigned Width = L->getWidth();
-      if (Width == 16) {
-        return ConcatExpr::create(ExtractExpr::create(L, 0, 8),
-                                  ExtractExpr::create(L, 8, 8));
+      constexpr unsigned bytelen = 8;
```
regehr: this is fixed in stone so it's probably clearer (and is definitely shorter) to just use "8" everywhere

lib/Extractor/KLEEBuilder.cpp
pranavk: i have kept all the whitespace changes as a separate commit. When PR is approved, I will squash everything but that commit. So when it gets merged, it will still be separate. For the review purposes, can we just review this specific commit only (clicking the above whitespace commit, eg:). Can create a new PR too if it's hard to review.

regehr: it isn't hard to review, it's wrong to put whitespace changes into a feature commit-- please do a separate PR

lib/Extractor/KLEEBuilder.cpp
pranavk: hmm. i actually thought the opposite that it's clearer to use a variable. It's `constexpr` anyways, so there's no performance penalty, etc. Bunch of 8s around can make things confusing sometimes..

lib/Extractor/KLEEBuilder.cpp
pranavk: ok

pranavk:
okay; this is ready from my side. I removed whitespace changes and squashed everything into one.

regehr:
ok, LGTM

PR #469, opened 2019-02-22
Disabled the crashing test case, Blockpc is not supported in exhaustive synthesis yet

zhengyangl:


test/Infer/no-const-inputs.opt
```diff
@@ -1,8 +1,6 @@
 ; REQUIRES: solver
 ; RUN: %souper-check -infer-rhs %solver -souper-infer-inst -souper-synthesis-comps=add,const %s > %t1
 ; RUN: %FileCheck %s < %t1
-; RUN: %souper-check -infer-rhs %solver -souper-exhaustive-synthesis %s > %t2
-; RUN: %FileCheck %s < %t2
```
regehr: when you want to keep a test case, but it is expected to fail, you should not remove important parts of it, since we won't remember what do add back later

just mark it as XFAIL

zhengyangl:
But isn’t XFAIL works on whole test case, this case also tests InstSynthesis.

regehr:
this test case only tests CEGIS, it doesn't test exhaustive at all

regehr:
supporting blockpc in exhaustive synthesis will not help at all

regehr:
LGTM thanks!

PR #470, opened 2019-02-24
CEGIS loop for constant synthesis in alive2 backend

zhengyangl:


lib/Infer/AliveDriver.cpp
```diff
@@ -156,6 +157,60 @@ struct ReturnLHSRAII {
 };
 }
 
+std::map<souper::Inst *, llvm::APInt>
+performCegisFirstQuery(tools::Transform &t,
+                  std::map<std::string, souper::Inst *> &SouperConsts,
```
regehr: please indent these lines so they are even with "tools" in the line above

regehr:
does this code support multiple constants on the RHS? if yes, please add some test cases

lib/Infer/AliveDriver.cpp
```diff
@@ -260,6 +315,68 @@ souper::AliveDriver::synthesizeConstants(souper::Inst *RHS) {
...
+    // First Query
+    {
+      ReturnLHSRAII foo{t, LHSF};
+      t.src = std::move(LHSF);
+
+      Tried ++;
+      // exceeds MAX_TRIES, stop immediately
+      if (Tried > MaxTries) {
+        if (DebugLevel > 2)
+          llvm::errs()<<"Time of tries reached maximum\n";
```
regehr: put spaces around << please

lib/Infer/AliveDriver.cpp
```diff
@@ -156,6 +157,60 @@ struct ReturnLHSRAII {
...
+  for (auto &[Var, Val] : TgtState.getValues()) {
+    auto &Name = Var->getName();
+    if (startsWith("%reserved", Name)) {
+      auto App = Val.first.value.isApp();
+      assert(app);
+      SMTConsts[Name] = (Z3_get_app_arg(smt::ctx(), App, 1));
+    }
+  }
+
+  smt::Solver::check({{(Sv.first.value == Tv.first.value) && (TriedExpr),
```
regehr: I don't think this is right
@manasij7479 can you look?
this looks like an equivalence check, but we must check refinement

lib/Infer/AliveDriver.cpp
```diff
@@ -156,6 +157,60 @@ struct ReturnLHSRAII {
...
+          if (R.isUnsat()) {
+            if (DebugLevel > 3)
+              llvm::errs()<<"No more new possible guesses\n";
+            return;
+          } else if (R.isSat()) {
+            auto &&Model = R.getModel();
+            smt::expr TriedAnte(false);
+
+            for (auto &[name, expr] : SMTConsts) {
+              TriedAnte |= (expr != smt::expr::mkUInt(Model.getInt(expr), expr.bits()));
```
regehr: does this work for values >64 bits?

lib/Infer/AliveDriver.cpp
```diff
@@ -260,6 +315,68 @@ souper::AliveDriver::synthesizeConstants(souper::Inst *RHS) {
...
+      // exceeds MAX_TRIES, stop immediately
+      if (Tried > MaxTries) {
+        if (DebugLevel > 2)
+          llvm::errs()<<"Time of tries reached maximum\n";
+        return {};
+      }
+
+      ConstMap = performCegisFirstQuery(t, Consts, TriedExpr);
+
+      if (ConstMap.empty())
```
regehr: what does it mean when this happens?

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -66,6 +66,9 @@ namespace {
   static cl::opt<bool> EnableDataflowPruning("souper-dataflow-pruning",
     cl::desc("Enable pruning based on dataflow analysis (default=false)"),
     cl::init(false));
+  static cl::opt<bool> SynthesisConstWithCegisLoop("souper-synthesis-const-with-cegis",
+    cl::desc("Synthesis constants with CEGIS (default=false)"),
```
regehr: I think default this to true since the exists-forall method isn't working well

regehr:
is it correct that this CEGIS loop isn't specializing the query using some example inputs?
if so, you're going to work on that @manasij7479 ?

lib/Infer/AliveDriver.cpp
manasij7479: Yeah, I have some weird bugs with this check.
Also, this query needs to be explicit about PC (assumes do not seem to be automatically implicated)
@zhengyangl  Is there a way we can avoid this check and use alive's verification as a black box?

manasij7479:
Once this works (and is provably better than the forall query), I can implement the input specialization techniques. It seems I am going to doing that in the near future anyway.

lib/Infer/AliveDriver.cpp
zhengyangl: No more guesses produced, terminate immediately.

lib/Infer/ExhaustiveSynthesis.cpp
zhengyangl: This method is also not working well for now, since no input specialization is implemented.

Say it cannot synthesis mul(x, 100) with less than 100 guess, when the solver guesses from 0 and increase the guess by 1 each time. Input 0 will always make the check satisfied for any constant.

lib/Infer/AliveDriver.cpp
regehr: I mean: under what circumstances does this happen?

zhengyangl: This happens when the first query is unsat, see the implementation of performCegisFirstQuery

lib/Infer/AliveDriver.cpp
zhengyangl: Yes this is the equivalence check. Need further work on this.

Also another thought is the input specialization should not be done in the Souper land. The query LHS(x_0)=RHS(x_0, C) /\ LHS(x_1)=RHS(x_1, C) /\ LHS(x)=RHS(x, C) is wrong since any poison triggered by specialized input will poison the whole query. We need to handle UB in these specialized queries separately

lib/Infer/AliveDriver.cpp
zhengyangl: I'll write a test for this, seems it should work for large const.

lib/Infer/AliveDriver.cpp
regehr: ok-- I think this deserves a printout at some debug level so we can see this happening

lib/Infer/AliveDriver.cpp
regehr: OK-- please add a comment explaining this to this PR
we can figure out how to fix this later

regehr: @manasij7479 could something like this explain the unsound results you've been seeing/?

manasij7479: I don't know.

regehr:
is this ready?

zhengyangl:
Not yet, will be ready soon

lib/Infer/AliveDriver.cpp
zhengyangl: Unfortunately the Alive driver needs to update to support LHS with width > 64, but my part of code should handle value > 64 bits without change.

zhengyangl:
> does this code support multiple constants on the RHS? if yes, please add some test cases

Test case added for RHS with a pair of consts.

zhengyangl:
The PR is ready for review again

lib/Infer/AliveDriver.cpp
regehr: @manasij7479 please do PRs bumping our alive version from time to time

regehr:
LGTM, thanks!

PR #471, opened 2019-02-24
Mark the blockpc5.opt as UNSUPPORTED instead of XFAIL

zhengyangl:
The assertion will not be triggered in debug mode and this test cases passes if the assertion is disabled. If a test is actually passing and it is marked XFAIL, then lit will report failure. 
So, instead of marking the case XFAIL, let's make it UNSUPPORTED until blockpc is supported in exhaustive synthesis.


regehr:
LGTM thanks

PR #472, opened 2019-02-26
Refactor interpreter and abstract interpreters into separate TU

manasij7479:


include/souper/Infer/Interpreter.h
```diff
@@ -0,0 +1,67 @@
...
+
+namespace souper {
+
+struct EvalValue {
+  enum class Kind {Val, Poison, Undef, Unimplemented};
+
+  EvalValue() : K(Kind::Unimplemented) {}
+  EvalValue(llvm::APInt Val) : Value{Val}, K(Kind::Val) {};
+
+  llvm::Optional<llvm::APInt> Value;
```
regehr: I don't feel too strongly about it, but maybe you want to just have an APInt here and we don't care about its value if there UB or whatever?

include/souper/Infer/Interpreter.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include "llvm/Support/KnownBits.h"
+#include "llvm/IR/ConstantRange.h"
+
+#include "souper/Inst/Inst.h"
+
+namespace souper {
+
+struct EvalValue {
+  enum class Kind {Val, Poison, Undef, Unimplemented};
```
regehr: I think it would be best to rename this enum since Souper already has a type called Kind that is important and pervasive

lib/Infer/Interpreter.cpp
```diff
@@ -0,0 +1,269 @@
...
+EvalValue getValue(Inst *I, ValueCache &C) {
+  if (I->K == souper::Inst::Const) {
+      return {I->Val};
+  } else if (I->K == souper::Inst::Var
+      || I->K == souper::Inst::ReservedConst
+      || I->K == souper::Inst::ReservedInst) {
+    if (I->Name != "" && C.find(I) != C.end()) {
+      return C[I];
+    } else {
+      return EvalValue(); // unavailable
```
regehr: can you explain what unavailable means now?

lib/Infer/Interpreter.cpp
```diff
@@ -0,0 +1,269 @@
...
+    case souper::Inst::Ctlz:
+      return {llvm::APInt(Inst->Width,
+                          args[0].getValue().countLeadingZeros())};
+    case souper::Inst::Cttz:
+      return {llvm::APInt(Inst->Width,
+                          args[0].getValue().countTrailingZeros())};
+    case souper::Inst::BSwap:
+      return {args[0].getValue().byteSwap()};
+    default:
+      return EvalValue(); // Indicates an 'unavailable' value
```
regehr: please change mentions of unavailable to unimplemented or whatever is correct now

regehr:
LGTM, thanks!

PR #478, opened 2019-02-27
add a frontend for the souper interpreter, mostly so we can test it

regehr:
cc @manasij7479

### DIFF
--- a/tools/souper-interpret.cpp
+++ b/tools/souper-interpret.cpp
@@ -1,0 +1,138 @@
+// Copyright 2014 The Souper Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


tools/souper-interpret.cpp
```diff
@@ -0,0 +1,138 @@
...
+     "The larger the number is, the more fine-grained debug "
+     "information will be printed."));
+
+static bool parseInput(const std::string &S, std::string &Name,
+                       std::string &Val) {
+  if (S[0] != '%')
+    return false;
+  int Pos = 1;
+  while (Pos < S.size() &&
+         (S[Pos] >= '0' && S[Pos] <= '9' ||
```
manasij7479: More parentheses here and the next two lines.
CI reports a build failure otherwise, coming from a warning turned into an error.

regehr: argh!! thanks

PR #479, opened 2019-02-27
Fix: Always create new copy of reservedconst when doing wiring in multiple instruction synthesis

zhengyangl:
Before the fix, C1 below points the same reservedconst Inst::Var object:
```

Inst1 C1  Inst2 C1
  \  /      \  /
  mul        mul
```

This is wrong. The cache in getInstCopy() when filling in constants after first query might hit the wrong value. This patch gurantees C1 and C2 always point to different objects.

This patch also fixes a potential error: when creating an unary inst or a select inst with a reservedinst as its argument, always call getReservedInst() for a new object. Again, this keeps the cache from breaking things. Binary operators are already following this.

### DIFF
--- a/lib/Infer/ExhaustiveSynthesis.cpp
+++ b/lib/Infer/ExhaustiveSynthesis.cpp
@@ -202,12 +202,16 @@ void getGuesses(std::vector<Inst *> &Guesses,
         if (Comp->K == Inst::ReservedConst)
           continue;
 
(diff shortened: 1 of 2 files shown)


zhengyangl:
@pranavk This fixes the distributive law test failure.

manasij7479:
Wow, this was subtle.
Thanks.
Does this also fix the synthesis failures I posted on slack a few weeks ago?   
I can track those down tomorrow.

regehr:
thanks!!

PR #482, opened 2019-02-28
stop test blockpc5.opt from running

zhengyangl:


zhengyangl:
This fixes the crash when testing with debug build.

regehr:
thanks!

PR #490, opened 2019-03-02
tests: tweak overflow tests to do something useful

pranavk:
There was lot of LLVM IR not related to doing the actual testing of overflow intrinisics. I removed all of it and kept only the part that "actually" checks for the functionality of overflow intrinsics.


regehr:
thanks!

PR #498, opened 2019-03-03
add poison and UB support to the interpreter

regehr:


regehr:
please review @manasij7479 
this isn't all of the changes that I want to make to the interpreter but it gets a lot of the main stuff out of the way

manasij7479:
LGTM

PR #499, opened 2019-03-03
docs: Enable interactive class graphs/diagrams, if you have graphviz

pranavk:
Might turn out to be helpful in debugging, in some cases:

See: http://www.cs.utah.edu/~pranavk/souperdocs/DataflowPruning_8cpp.html


regehr:
cool!!!
LGTM

PR #501, opened 2019-03-04
Partial evaluation for abstract interpreters, some refactoring

manasij7479:


lib/Infer/Interpreter.cpp
```diff
@@ -2,6 +2,31 @@
 
 namespace souper {
 
+  bool hasReservedHelper(Inst *I, std::set<Inst *> &Visited) {
+    if (I->K == Inst::ReservedConst || I->K == Inst::ReservedInst)
+      return true;
+    if (I->K == Inst::Var && (I->Name.find(ReservedConstPrefix) != std::string::npos)) {
+      return true;
+    } else if (I->K == Inst::Var && (I->Name.find(ReservedInstPrefix) != std::string::npos)) {
```
regehr: I think it's better style to avoid the else when the if-part just returns (like you did above, but not right here)

lib/Infer/Interpreter.cpp
```diff
@@ -368,53 +404,200 @@ namespace souper {
...
+        Op0KB.One <<= Val;
+        Op0KB.Zero <<= Val;
+        Op0KB.Zero.setLowBits(Val);
+        // setLowBits takes an unsigned int, so getLimitedValue is harmless
+        return Op0KB;
+      } else {
+        return Result;
+      }
+    }
+//   case ShlNSW:
```
regehr: I believe all UB-flag versions of each instruciton can just use the normal version
this'll miss a bit of precision but it's a lot better than nothing

regehr: in other words, addnsw can maybe return a better answer than add, but the answer for add should be a sound answer for addnsw

regehr:
LGTM

Issue #334, opened 2018-07-02
Are utils/ bitrot?


LebedevRI:
`cache_dump` does not work out-of-the box
```diff
-    my %h = $r->hgetall($opt);
+    my %h;
+    if($r->TYPE($opt) eq "hash") {
+      %h = $r->hgetall($opt);
+    } else {
+      next;
+    }
```
`souper2llvm` seems to always generate invalid IR
Is there some other tools that are supposed to be used?

regehr:
cache_dump works out of the box for me, I was just using it yesterday. You'll need to provide standard bug report information (your platform details, steps to reproduce, observed output, etc.) for us to figure out why it isn't working for you.

souper2llvm is more of a research tool and not really something that we want to support, and it contains a bunch of known bugs. I'd suggest not relying on it for now. It was written by @rsas but he probably doesn't have time to support it at the moment.

rsas:
Hi @LebedevRI, as mentioned by @regehr, ``souper2llvm`` was created a while ago for testing purposes and was not updated/maintained since. It should not be difficult to update/fix it. Contributions are very welcome.

LebedevRI:
> cache_dump works out of the box for me

That is interesting. The reproduction steps are trivial.
1. This is debian sid, fully updated. 
```
 dpkg -l | grep redis
ii  libredis-perl                                    2:1.9910-1                                             all          persistent key-value database with network interface (Perl client library)
ii  redis                                            5:4.0.10-1                                             all          Persistent key-value database with network interface (metapackage)
ii  redis-server                                     5:4.0.10-1                                             amd64        Persistent key-value database with network interface
ii  redis-tools                                      5:4.0.10-1                                             amd64        Persistent key-value database with network interface (client)
```
2. Clone this repo. (i'm at b41a0944c0f0c5b106c71f0ba390290f6ef07b2c)
3. Run `$ ./build_deps.sh` without arguments
4. `$ mkdir souper_build && cd souper_build && cmake -DTEST_SOLVER="-z3-path=/usr/bin/z3" -GNinja ../ && ninja`
5. `$ redis-cli flushall`
6. Build something with `sclang` with `SOUPER_DYNAMIC_PROFILE` and `SOUPER_STATIC_PROFILE` set.
7. Run `./cache_dump`, and it fails with
(i apologize, this is from memory, i butchered the build at the moment):
[...]

PR #504, opened 2019-03-04
[broken draft wip] Souper codegen to LLVM IR

LebedevRI:
"Words are cheap, ..."

As discussed in https://github.com/google/souper/issues/334#issuecomment-468963713, current `souper2llvm` script is bitrot,
and unusable for some time now, even though it is super needed for triaging
of missed LLVM transforms. This is my yet-not-fully-scratched itch.
I suspect i might have some laying around, but i can't really triage them manually :/

Now, it should of course be possible to just perform some voodoo on the script,
and resurrect it until it is broken next time. But wouldn't it be nicer to have A Proper Way? :)

This is **extremely** partial, just a proof-of-concept.
It **does** handle every souper instr (a **LOT** of them was missed by Pass code :/),
but does not handle any `phi`/`pc`/whatever control flow.
There is _some_ test coverage. (If anything, that is already better than without this tool.)

Also, while this has the same name as the `souper2llvm` script that is being deleted here,
it is not functionally equivalent, it does not parse whole replacements, only RHS.


regehr:
Hi @LebedevRI !
I'm sorry to have let this PR languish. We definitely want to merge it. I think that one of us can spare some time to make phi work. Thanks for doing this awesome work (especially the test cases).

LebedevRI:
> I'm sorry to have let this PR languish.

Actually, it was me who stopped paying attention to it, i don't recall what was the problem.

regehr:
As long as it doesn't break anything that currently works, I don't know that there's any problem here. I'm fine to merge incomplete code, and then we can work on it more.

I read it a while ago, will re-read carefully in the next day or two.

regehr:
@zhengyangl can you please review this? I know you have looked at this code

regehr:
@LebedevRI could you modify this PR to avoid deleting the souper2llvm python program?
I agree it needs to die, but we should wait until some functionality which uses it has been migrated to the new souper2llvm

zhengyangl:
This patch LGTM :)

LebedevRI:
This is not any more ready than it was originally, even less so due to bitrot, but if that's what is wanted..

LebedevRI:
That link-time issue is new one, i would say llvm is now being miscompiled - why is it built without `-fPIC` in the first place? Feel free to either take this over or i'll close the PR in a day or so.

regehr:
please take this over @zhengyangl ?

zhengyangl:
@regehr yes sure

LebedevRI:
For reference, i did try to take a look what the problem is, and every `report_fatal_error()` is the problem. This doesn't look like a mistake in the diff.

regehr:
we'll dig into it. thanks Roman!

LebedevRI:
Resolved by #676.

regehr:
thanks again @LebedevRI -- we think we know what to do to fix up the remaining issues, or at least most of them

PR #506, opened 2019-03-05
linking: add another missing dependency

pranavk:
There can be another missing dependencies that we are not aware of yet. We are just fixing them as we are coming across them. And these linking errors are bit hard to reproduce; they sometime happen sometimes not.


regehr:
LGTM, thanks!

PR #508, opened 2019-03-05
Improve KnownBits interpreter & make it testable 

pranavk:
Work in progress ...


regehr:
looks good, please squash

regehr:
ah nevermind, I see it's a wip

pranavk:
yeah, this PR is just for tracking for now; i will let you know when it's ready to merge.

include/souper/Infer/AbstractInterpreter.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
```
regehr: if we're going to do this, let's rather do it across the board in a separate PR

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -1,6 +1,272 @@
...
+    }
+
+    llvm::KnownBits mul(const llvm::KnownBits &lhs, const llvm::KnownBits &rhs) {
+      llvm::KnownBits Result(lhs.getBitWidth());
+
+      auto trailingZeros0 = lhs.countMinTrailingZeros();
+      auto trailingZeros1 = rhs.countMinTrailingZeros();
+      Result.Zero.setLowBits(std::min(trailingZeros0 + trailingZeros1, lhs.getBitWidth()));
+
+      // check for leading zeros
```
regehr: I think there are good opportunties for improvement here by counting leading ones as well as zeroes, right? no need to fix it now but please add a comment sketching out the necessary code

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -1,6 +1,272 @@
...
+      return Result;
+    }
+
+    llvm::KnownBits and_(const llvm::KnownBits &lhs, const llvm::KnownBits &rhs) {
+      auto Op0KB = lhs;
+      auto Op1KB = rhs;
+
+      Op0KB.One &= Op1KB.One;
+      Op0KB.Zero |= Op1KB.Zero;
+      return Op0KB;
```
regehr: isn't there a less awkward way to express this computation in a functional style, instead of creating these temporary objects and mutating them?

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,432 @@
...
+      return merge(bruteForce(x, setLowest(y), Pred),
+		   bruteForce(x, clearLowest(y), Pred));
+    auto xc = x.getConstant();
+    auto yc = y.getConstant();
+
+    KnownBits res(x.getBitWidth());
+    switch (Pred) {
+    case Inst::AddNUW:
+    case Inst::AddNW:
+    case Inst::Add:
```
regehr: where is addnsw?

lib/Infer/AbstractInterpreter.cpp
pranavk: i just copied them from the original file to here; will see what we can do here.

regehr: it's fine for now if it's a copy

unittests/Interpreter/InterpreterTests.cpp
pranavk: added now in latest commit

regehr:
ok this is looking good

but please add a small unsoundness to one of the transfer functions and make sure that it is caught by the exhaustive testing framework, ok?

regehr:
nothing about that will appear in a commit, of course-- it's just a sanity check you need to do before committing this

pranavk:
> ok this is looking good
> 
> but please add a small unsoundness to one of the transfer functions and make sure that it is caught by the exhaustive testing framework, ok?

yup. i have done this in couple of transfer functions and the testing framework was able to catch the unsoundness. This is how the output looks like:

```
[pranavk@pranavk build]$ ./interpreter_tests 
[==========] Running 3 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 3 tests from InterpreterTests
[ RUN      ] InterpreterTests.KBTransferFunctions
Unsound!! mul
0001 mul 0001
Calculated: ???0
Expected: 0001
../unittests/Interpreter/InterpreterTests.cpp:376: Failure
Value of: testFn(Inst::Mul)
  Actual: false
Expected: true

```

regehr:
perfect!

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+		       << Res2.getBitWidth() << ", Given: " << Res1.getBitWidth() << '\n';
+	  return false;
+	}
+	if (Res1.hasConflict() || Res2.hasConflict()) {
+	  llvm::errs() << "Expected or Given result has a conflict\n";
+	  return false;
+	}
+
+	for (unsigned i = 0; i < Res1.getBitWidth(); i++) {
+	  if ((Res1.Zero[i] and Res2.One[i]) || (Res1.One[i] and Res2.Zero[i])) {
```
regehr: let's avoid the alternative operators like "and" unless there's a compelling reason to use

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+	  Res1 = BinaryTransferFunctionsKB::slt(x, y);
+	  break;
+	case Inst::Ule:
+	  Res1 = BinaryTransferFunctionsKB::ule(x, y);
+	  break;
+	case Inst::Sle:
+	  Res1 = BinaryTransferFunctionsKB::sle(x, y);
+	  break;
+	default:
+	  break;
```
regehr: would be best to make this an llvm::report_fatal_error

regehr: try to avoid writing code that could hide a bug

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+    {
+      res = KnownBits(1);
+      if (xc.sle(yc))
+	res.One.setBit(0);
+      else
+	res.Zero.setBit(0);
+    }
+    break;
+    default:
+      break;
```
regehr: same-- this shouldn't get triggered, make it a fatal error

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+      if (xc.slt(yc))
+	res.One.setBit(0);
+      else
+	res.Zero.setBit(0);
+    }
+    break;
+    case Inst::Ule:
+    {
+      res = KnownBits(1);
+      if (xc.ule(yc))
```
regehr: can you just return the result of the ule call? why do you need to make a new object of the knownbits type?

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+      auto rc = xc | yc;
+      res.One = rc;
+      res.Zero = ~rc;
+    }
+    break;
+    case Inst::Xor:
+    {
+      auto rc = xc ^ yc;
+      res.One = rc;
+      res.Zero = ~rc;
```
regehr: the repeated code for these is super ugly, why not just compute the APInt result for all operations and then at the bottom do the res.One and res.Zero thing?

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+    }
+    break;
+    case Inst::Mul:
+    {
+      auto rc = xc * yc;
+      res.One = rc;
+      res.Zero = ~rc;
+    }
+      break;
```
regehr: indentation

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+      auto rc = xc + yc;
+      res.One = rc;
+      res.Zero = ~rc;
+    }
+    break;
+    case Inst::SubNUW:
+    case Inst::SubNSW:
+    case Inst::SubNW:
+    case Inst::Sub:
+    {
```
regehr: put this kind of curly brace on the line above after the :

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+  }
+
+  KnownBits setLowest(KnownBits x) {
+    for (int i = 0; i < x.getBitWidth(); i++) {
+      if (!x.Zero[i] && !x.One[i]) {
+	x.One.setBit(i);
+	return x;
+      }
+    }
+    assert(false);
```
regehr: this should be an llvm::report_fatal_error or llvm::unreachable

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -0,0 +1,442 @@
...
+	x.Zero.clearBit(i);
+	x.One.setBit(i);
+	return true;
+      }
+      if (!x.Zero[i] && x.One[i]) {
+	x.Zero.clearBit(i);
+	x.One.clearBit(i);
+	continue;
+      }
+      assert(false && "error in nextKB");
```
regehr: report_fatal_error or unreachable

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -250,34 +438,38 @@ namespace souper {
...
-      }
+      return BinaryTransferFunctionsKB::ne(KB0, KB1);
+    }
+    case Inst::Ult: {
+      return BinaryTransferFunctionsKB::ult(KB0, KB1);
+    }
+    case Inst::Slt: {
+      return BinaryTransferFunctionsKB::slt(KB0, KB1);
+    }
+    case Inst::Ule: {
```
regehr: always avoid curly braces when not needed, like here

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -250,34 +438,38 @@ namespace souper {
       } else {
         return Result;
       }
-      if ((Other.Zero & 1) != 0) {
+
+      // Constants are never equal to 0
+      if (Other.Zero.isAllOnesValue()) {
         Result.Zero.setBit(0);
         return Result;
```
regehr: I know this isn't your code but please factor out the common "return Result"

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -1,6 +1,268 @@
...
+      llvm::KnownBits Result(1);
+      if (getSMax(lhs).slt(getSMin(rhs)))
+	Result.One.setBit(0);
+      if (getSMin(lhs).sge(getSMax(rhs)))
+	Result.Zero.setBit(0);
+      return Result;
+    }
+
+    llvm::KnownBits ule(const llvm::KnownBits &lhs, const llvm::KnownBits &rhs) {
+      llvm::KnownBits Result(1);
```
regehr: instead of creating a tempoarary object and mutating it, you should just construct the value that you want in the first place, there are several ways to do it, I think any is fine

regehr: return APInt(1, 0) or return APInt(1, 1) should work just fine

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -1,6 +1,268 @@
...
+      unsigned Leaders =
+        std::max(lhs.countMinLeadingZeros(), rhs.countMinLeadingZeros());
+      Result.resetAll();
+      Result.Zero.setHighBits(Leaders);
+      return Result;
+    }
+
+    llvm::KnownBits and_(const llvm::KnownBits &lhs, const llvm::KnownBits &rhs) {
+      auto result = lhs;
+
```
regehr: don't put a blank line here if you don't have an analogous blank line in the next function

lib/Infer/AbstractInterpreter.cpp
pranavk: We are not returning APInt here; we are returning KnownBits. KnownBits doesn't have any public constructor that can let you do that; hence, the reason for this thing. There's a private constructor with which we can accomplish something like this but I think that would be much more unreadable than the current form.

unittests/Interpreter/InterpreterTests.cpp
pranavk: result of the ule call is an APInt and KnownBits doens't have a public constructor that takes APInt.

lib/Infer/AbstractInterpreter.cpp
regehr: ugh.

pranavk:
Okay. i have squashed it. feel free to merge.

regehr:
LGTM!

PR #511, opened 2019-03-18
Allow running individual tests

pranavk:
Usage:

run_lit script now accepts an argument. If this argument is given, which
can either be a file or directory, only tests in that file or directory
are executed.

Eg:

./run_lit test/Solver/
will execute all tests in Solver/ directory
./run_lit test/Sover/saturating1.ll
will execute only one test contained in saturating1.ll


regehr:
LGTM, thanks!

PR #512, opened 2019-03-18
Add abstract interpreters in souper-interpreter executable

pranavk:
See commit messages for details.

### DIFF
--- a/tools/souper-interpret.cpp
+++ b/tools/souper-interpret.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "llvm/Support/KnownBits.h"
 #include "llvm/Support/MemoryBuffer.h"
 #include "souper/Infer/Interpreter.h"
 #include "souper/Parser/Parser.h"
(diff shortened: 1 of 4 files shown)


pranavk:
I know tests are missing but maybe it's a good idea to add them in a later PR when the Dataflow stuff is bit more stable so that we don't have to change tests again and again everytime we make some improvement in dataflow stuff?

regehr:
@pranavk can you give an example of a test that would be fragile under an improvement to the abstract interpreter?

pranavk:
improving precision for example in abstract interpreters will give different output. Current tests, for example, if gives XX00 for some combination of instructions, after improvement, we might be getting something like X100. Same for constant ranges. We can probably do something to make tests flexible but not sure if FileCheck can do that; so, it may involve a bit more work.

regehr:
a test that is fragile in this way is a bad test. but we can write better tests. it's not hard to ask the abstract interpreter to create a result at least as precise as a specified one.

pranavk:
okay. makes sense. i might have to look into non-FileCheck based tests then to acheive "at-least-this-precise" kind of thing

regehr:
filecheck is just a string matcher, don't try to get more out of it than that

precision-based tests will have to rely on a custom tool we write in C++ that prints something like "sound and precise" and then filecheck looks for that

we can compute optimally precise known bits using souper itself, that should be where the test oracle comes from here

regehr:
I think instead of specifying abstract interpreters you should just run all of them

regehr:
otherwise looks good-- please make that small change and squash

regehr:
@pranavk please tweak so it unconditionally runs all abstract interpreters instead of asking which ones to run on the command line

tools/souper-interpret.cpp
```diff
@@ -33,6 +34,35 @@ static cl::opt<unsigned> DebugLevel("souper-debug-level", cl::init(1),
...
+  };
+
+  CompareDataflowResult compareKnownBits(KnownBits &a, KnownBits &b) {
+    assert(!a.hasConflict() && !b.hasConflict() && "Can't compare conflicting KnownBits operands!");
+
+    APInt unknownA = ~(a.One ^ a.Zero);
+    APInt unknownB = ~(b.One ^ b.Zero);
+    if ((unknownA | unknownB).countPopulation() >
+	std::max(unknownA.countPopulation(), unknownB.countPopulation()))
+      return CompareDataflowResult::INCOMPARABLE;
```
regehr: I don't think this is right: two values conflict when one of them thinks a bit is 1 and the other thinks its 0

I would encourage you to implement "compare" for known bits and then just use the same code to compare known bits values as you use for constantranges

pranavk: yeah, you are right. i missed a case here.

pranavk: pushed a modified version. Now, I am checking for the conflicts first in the known bits and after we are clear that there is no conflict in known bits, we are moving forward with the code I had earlier.

regehr:
looks good
remove the command line option and I think this is ready

pranavk:
New output looks like this:

```
[pranavk@pranavk build]$ ninja; ./souper-interpret -input-values=%0=1,%3=0 -z3-path=/usr/bin/z3 ../../interpret1.opt 
[3/3] Linking CXX executable souper-interpret

 -------- KnownBits Interpreter ----------- 
KnownBits result: 
???????????????????????????????1
KnownBits result using solver: 
????????????????????????????????

Dataflow is more precise than solver.

 -------- ConstantRanges Interpreter ----------- 
ConstantRange result: 
[1,0)
ConstantRange result using solver: 
full-set

Dataflow is more precise than solver.
```

pranavk:
command line option has been removed

regehr:
great, LGTM

PR #517, opened 2019-03-20
fix debug build

pranavk:
1) llvm_unreachable accepts another parameter called message. So, the original code was never correct. The reason it worked was because we only tested it on release build where due to presence of NDEBUG, it's not compiled at all.

2) Let's use report_fatal_error because we don't want to not hit this error in Release builds.


regehr:
yes, report_fatal_error is usually the better choice

PR #518, opened 2019-03-21
Pass values from the command line to solver

pranavk:
This is a fallback from #512 

We also need to tell solver the values given by the user in the command line to have accurate results.


regehr:
good catch, I missed this!!

PR #522, opened 2019-03-21
Refactor: hasReservedInst -> hasGivenInst

pranavk:
Pass your lambda to check if instruction with given traits exists

+ The hasReservedInst function was not being used by anyone before


regehr:
LGTM

PR #523, opened 2019-03-22
Allow reservedinst to be given to souper-interpreter

pranavk:
This modifies the parser to read reservedinst correctly. So following thing is parsed correctly now:

`%0:i32 = reservedinst` 

as an instruction of Kind ReservedInst.

I also had to modify the KLEE backend to treat these instructions of ReservedInst like any other variable. I think this is correct but shout if not.


regehr:
there must have been a motivating example for this patch, can you turn it into a test case?

pranavk:
Added tests which were not possible earlier because parse didn't understand what `reservedinst` is.

regehr:
LGTM thanks

PR #524, opened 2019-03-22
Force reservedinst and reservedconst to stay in the designated slots

zhengyangl:


regehr:
is there an example that failed before this, that works now, that you can turn into a test case for this?

regehr:
LGTM

PR #528, opened 2019-04-09
add support for generating dot format for souper IR

pranavk:
./souper-check -emit-lhs-dot now also spits out the dot format of the
souper IR which was parsed which can be used to generate PNG like below:

dot -Tpng test.dot > a.png

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,6 +206,7 @@ add_library(souperInfer STATIC
 set(SOUPER_INST_FILES
   lib/Inst/Inst.cpp
   include/souper/Inst/Inst.h
+  include/souper/Inst/InstGraph.h
 )
 
 add_library(souperInst STATIC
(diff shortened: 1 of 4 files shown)


pranavk:
I haven't looked at that thing in detail but isn't that more suitable for LLVM IR? How easy it is to adapt it to any other IR like souper?

LebedevRI:
That is a wrapper around DOT language, not a DOT writer for LLVM IR.

regehr:
you should take a look at it, @pranavk, and give us a quick comparison of the approaches
I'm not opposed to merging this code since it's really simple-- but is it going to get more complicated?

pranavk:
@LebedevRI thanks. Inheriting couple of LLVM's GraphTrait classes and it worked. 

@regehr updated the PR to reuse LLVM's graph logic with updated test.

regehr:
LGTM, thanks!

PR #529, opened 2019-04-09
sclang/cache_dump script changes for pruning and redis_port

pranavk:
* support dataflow pruning in sclang

* Support redis port in sclang

* redis port support in cache_dump


regehr:
LGTM tnanks

PR #530, opened 2019-04-09
Dataflow pruning improvements

pranavk:


pranavk:
Everything is squashed in one commit except two assertion removal that were blind fixes. You can see those as separate commits above.

regehr:
what's with all the whitespace changes?

pranavk:
those are replacing tabs with spaces. Unfortunately, it will be very time-consuming now to separate those in a separate commit .

regehr:
I just want you to make sure all of your changes are in the correct direction

regehr:
also we should setup some sort of CI check to look for tabs

include/souper/Infer/AbstractInterpreter.h
```diff
@@ -39,6 +45,46 @@ namespace souper {
...
+                  bool ConsiderHoles = true);
+
+  class KnownBitsAnalysis {
+    std::unordered_map<Inst*, llvm::KnownBits> KBCache;
+
+    // Checks the cache or instruction metadata for knonwbits information
+    bool cacheHasValue(Inst* I);
+
+  public:
+    llvm::KnownBits findKnownBits(Inst* I,
```
regehr: please place * and & next to the idengtifier, not the type

include/souper/Infer/AbstractInterpreter.h
```diff
@@ -39,6 +45,46 @@ namespace souper {
...
+                                  ConcreteInterpreter& CI,
+                                  bool PartialEval = true);
+
+    static llvm::KnownBits findKnownBitsUsingSolver(Inst *I,
+                                                    Solver *S,
+                                                    std::vector<InstMapping> &PCs);
+
+    static std::string knownBitsString(llvm::KnownBits KB);
+
+    static llvm::KnownBits getMostPreciseKnownBits(llvm::KnownBits a, llvm::KnownBits b);
```
regehr: initial capitals for parameter names

regehr:
what's the file "test/Pass/.nop1.ll.swp"?

pranavk:
> what's the file "test/Pass/.nop1.ll.swp"?

silly me. Removed.

include/souper/Infer/AbstractInterpreter.h
```diff
@@ -39,6 +45,46 @@ namespace souper {
...
+
+  bool isConcrete(souper::Inst *I,
+                  bool ConsiderConsts = true,
+                  bool ConsiderHoles = true);
+
+  class KnownBitsAnalysis {
+    std::unordered_map<Inst*, llvm::KnownBits> KBCache;
+
+    // Checks the cache or instruction metadata for knonwbits information
+    bool cacheHasValue(Inst* I);
```
regehr: can you go through this patch and systematically write "Inst *I" instead of "Inst* I"?

lib/Infer/Pruning.cpp
```diff
@@ -41,71 +28,112 @@ std::string getUniqueName() {
...
-      if (!isConcrete(RHS)) {
-        auto CR = findConstantRange(RHS, InputVals[I]);
-        if (StatsLevel > 2)
-          llvm::errs() << "  RHS ConstantRange = " << CR << "\n";
-        if (!CR.contains(Val)) {
-          if (StatsLevel > 2) {
-            llvm::errs() << "  pruned using CR! ";
+    if (LHSHasPhi) {
+      auto LHSCR = LHSConstantRange[I];
+      auto RHSCR = ConstantRangeAnalysis().findConstantRange(RHS, ConcreteInterpreters[I]);
```
regehr: ok, so if I understand correctly, every time we invoke the known bits or constant range analysis, we're creating a new object of the analysis class in order to store the cache, and then throwing this away?

what we usually do in this case in souper is create a separate context object and use that-- is that not a good approach here?

regehr:
remove test/Pass/.nop1.ll.swp

regehr:
I was hoping for some new tests in this PR. what is there that can be tested here?

regehr:
@pranavk @manasij7479 I'm not sure which one of you likes to write ```Inst* I``` but this is just wrong. to see this, consider what it leads one to assume about ```int* x, y``` vs what that definition actually does

lib/Infer/Pruning.cpp
pranavk: Yes, that's true. The `ConstantRange` and `KnownBitsAnalysis` classes are just a wrapper around previous `findKnownBits`, `findConstantRange` functions to make them more efficient by making use of cache. This is helpful when some instruction has a lot of "uses".

Since all RHSes are different, we don't try to share the cache across different RHS evaluations as that will lead to misevaluations.

pranavk:
> @pranavk @manasij7479 I'm not sure which one of you likes to write Inst* I but this is just wrong. to see this, consider what it leads one to assume about int* x, y vs what that definition actually does

this is me. bad habits  :| I will fix this

include/souper/Infer/AbstractInterpreter.h
pranavk: should be done now.

pranavk:
> I was hoping for some new tests in this PR. what is there that can be tested here?

yes, i will think about it and add some

pranavk:
> I was hoping for some new tests in this PR. what is there that can be tested here?

i was thinking of following things:

1) Create instance of KB and CR analysis classes and check if they are caching the results as expected.
 2) Also check if they can read metadata from the instructions.
 3) Make sure ConcreteInterpreter doesn't write to cache except during construction -- we don't want individual calls of evaluateSingleInst to be cached by design.
 4) Check if `PartialEval` flag is working fine for both KB and CR.

But this is assuming you are happy with the overall design; otherwise, no point in writing tests like this.

regehr:
overall design is fine, please proceed with tests as described

include/souper/Infer/AbstractInterpreter.h
```diff
@@ -16,29 +16,75 @@
...
+    llvm::KnownBits ne(const llvm::KnownBits &LHS, const llvm::KnownBits &RHS);
+    llvm::KnownBits ult(const llvm::KnownBits &LHS, const llvm::KnownBits &RHS);
+    llvm::KnownBits slt(const llvm::KnownBits &LHS, const llvm::KnownBits &RHS);
+    llvm::KnownBits ule(const llvm::KnownBits &LHS, const llvm::KnownBits &RHS);
+    llvm::KnownBits sle(const llvm::KnownBits &LHS, const llvm::KnownBits &RHS);
   }
+
+  bool isConcrete(souper::Inst *I,
+                  bool ConsiderConsts = true,
+                  bool ConsiderHoles = true);
```
regehr: I think there don't exist any calls with Considerholes = false; if this is true, please eliminate this flag

include/souper/Infer/AbstractInterpreter.h
```diff
@@ -16,29 +16,75 @@
...
+  class KnownBitsAnalysis {
+    std::unordered_map<Inst*, llvm::KnownBits> KBCache;
+
+    // Checks the cache or instruction metadata for knonwbits information
+    bool cacheHasValue(Inst *I);
+
+  public:
+    llvm::KnownBits findKnownBits(Inst *I,
+                                  ConcreteInterpreter &CI,
+                                  bool PartialEval = true);
```
regehr: is this default value really needed? I really like to avoid default valued boolean flags when possible, they make using functions confusing

include/souper/Infer/AbstractInterpreter.h
```diff
@@ -16,29 +16,75 @@
...
+  class ConstantRangeAnalysis {
+    std::unordered_map<Inst*, llvm::ConstantRange> CRCache;
+
+    // checks the cache or instruction metadata for cr information
+    bool cacheHasValue(Inst *I);
+
+  public:
+    llvm::ConstantRange findConstantRange(souper::Inst *I,
+                                          ConcreteInterpreter &CI,
+                                          bool PartialEval = true);
```
regehr: same here-- if the only use case for turning off partial eval is testing, then we should just remove the parameter and adapt the test cases, unless there's a compelling argument why testing is better with partial evaluation turned off

include/souper/Infer/Interpreter.h
```diff
@@ -80,30 +81,39 @@ struct EvalValue {
...
+EvalValue evaluateSubNW(llvm::APInt A, llvm::APInt B);
+EvalValue evaluateUDiv(llvm::APInt A, llvm::APInt B);
+EvalValue evaluateURem(llvm::APInt A, llvm::APInt B);
+EvalValue evaluateShl(llvm::APInt A, llvm::APInt B);
+EvalValue evaluateLShr(llvm::APInt A, llvm::APInt B);
+EvalValue evaluateAShr(llvm::APInt A, llvm::APInt B);
+
+  class ConcreteInterpreter {
+    ValueCache _Cache;
+    bool _CacheWritable = false;
```
regehr: can you explain the purpose of this flag?
also, what is the rationale for prefixing an underscore on these field names?

include/souper/Infer/Pruning.h
```diff
@@ -29,33 +29,35 @@ typedef std::function<bool(Inst *, std::vector<Inst *> &)> PruneFunc;
 
 class PruningManager {
 public:
-  PruningManager(Inst *LHS_, std::vector< souper::Inst* >& Inputs_,
-                 unsigned int StatsLevel_, InstContext& IC_,
-                 SMTLIBSolver *SMTSolver_);
+  PruningManager(SynthesisContext &SC_, std::vector< souper::Inst* > &Inputs_,
```
regehr: do you have a rationale for putting spaces inside the angle brackets? if there's not a specific reason here, then please just be consistent with other code and don't do this

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -200,185 +217,236 @@ namespace souper {
...
-      if (PartialEval && isConcrete(I)) {
-        return evaluateInst(I, C);
-      }
-    }
+    else if (I->K == Inst::Var && !isReservedConst(I))
+      // evaluateInst will only give us input value of the variable; it doesn't
+      // evaluate anything.
+      return CI.evaluateInst(I);
+
+    if (PartialEval && isConcrete(I))
```
regehr: yeah, I can't think of a good reason not to just do partial evaluation whenever it's possible

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -200,185 +217,236 @@ namespace souper {
...
-#define VAL(INST) getValue(INST, C, PartialEval)
+#define KB0 findKnownBits(I->Ops[0], CI, PartialEval)
+#define KB1 findKnownBits(I->Ops[1], CI, PartialEval)
+#define KB2 findKnownBits(I->Ops[2], CI, PartialEval)
+#define VAL(INST) getValue(INST, CI, PartialEval)
+
+  llvm::KnownBits mergeKnownBits(std::vector<llvm::KnownBits> Vec) {
+    assert(Vec.size() > 0);
+
+    auto width = Vec[0].getBitWidth();
```
regehr: this is a pretty heavyweight kind of debugging code-- I'd say only do this when assertions are enabled or something

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -200,185 +217,236 @@ namespace souper {
...
+    for (unsigned i = 1; i < Vec.size(); i++) {
+      if (width != Vec[i].getBitWidth())
+        llvm::report_fatal_error("mergeKnownBits: bitwidth should be same of all inputs");
+    }
+
+    KnownBits Result(width);
```
regehr: this code here feels too slow... can't you just AND together the known zeros and the known ones and be done with it??

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -390,35 +458,54 @@ namespace souper {
...
 //   case LShrExact:
 //     return "lshrexact";
     case Inst::AShr:
       if (isReservedConst(I->Ops[1])) {
-	if (KB0.Zero[KB0.getBitWidth() - 1])
-	  Result.Zero.setHighBits(2);
-	if (KB0.One[KB0.getBitWidth() - 1])
-	  Result.One.setHighBits(2);
+        if (KB0.Zero[KB0.getBitWidth() - 1])
+          Result.Zero.setHighBits(2);
```
regehr: I don't understand this, is it saying that if we're going to synthesize a constant shift amount, we'll avoid synthesizing a zero there? and if so, is that true? please check with Zhengyang and make sure we never synthesize a zero

include/souper/Infer/AbstractInterpreter.h
pranavk: The only way I have used this till now is to test this thing when running experiments last month. This was already there before my modifications though. @manasij7479 do you've any comments?

include/souper/Infer/AbstractInterpreter.h
pranavk: tagging @manasij7479  as it was function signature was added by him originally.

include/souper/Infer/Interpreter.h
pranavk: Cache is only writable during the construction of ConcreteInterpreter. See it's ctor how it's being used. We don't want subsequent invocation of evaluateSingleInst to modify the cache. That's why this flag is there.

Rational behind underscore is to distinguish member variables with the local variables in a function body. Imagine a huge function and you are reading it. Some sort of distinction between member variable and local variables helps, I think. I have seen such a usage somewhere at least once in souper code; so I used them here too.

include/souper/Infer/Pruning.h
pranavk: It was already there. See the removed line. I didn't want to change the style, so kept the spaces.

lib/Infer/AbstractInterpreter.cpp
pranavk: @manasij7479 ^

include/souper/Infer/AbstractInterpreter.h
pranavk: @manasij7479

lib/Infer/AbstractInterpreter.cpp
pranavk: makes sense. will modify.

lib/Infer/AbstractInterpreter.cpp
pranavk: yes, this is one of the major assumption used in a lot of places in dataflow pruning code. @manasij7479 

@zhengyangl we don't synthesize 0 for a reservedconst, right?

zhengyangl: Yes. The current constant synthesis can synthesis 0 if the enumeration reaches the guess (ashr x, c). If the NOP synthesis is turned off and the ashr is set to the only operator allowed, we get this synthesis result:

```
liuz@intel:~/work/build$ cat b.opt
%0:i8 = var
%1:i8 = add %0, 0:i8
%2:i8 = add %1, 0:i8
infer %2
liuz@intel:~/work/build$ ./souper-check -infer-rhs -souper-exhaustive-synthesis -z3-path=/usr/bin/z3 b.opt
; RHS inferred successfully
%3:i8 = ashr %0, 0:i8
result %3
```

For now we there is no code for setting constraints on the constants for synthesis. However, for this assumption, I think in the short term we can tell in our current cost model we never synthesize 0, based on the fact that bringing a new instruction will always increase the cost and the guesses without ashr NOP comes earlier when doing enumeration.

lib/Infer/AbstractInterpreter.cpp
manasij7479: The unit tests rely on this feature, iirc.

include/souper/Infer/AbstractInterpreter.h
manasij7479: Feel free to adapt the tests, not very useful to keep this feature.

include/souper/Infer/AbstractInterpreter.h
manasij7479: I was using it for generating separate experimental results for pruning constants and holes.
We will need that again in future.

lib/Infer/AbstractInterpreter.cpp
manasij7479: We can synthesize 0.
It is rarely useful to do so.
If we can find a situation where it is useful to synthesize zero, we can fall back to consider that case specifically in the abstract interpreters.

include/souper/Infer/Interpreter.h
manasij7479: Aren't identifiers starting with _ considered reserved for the standard library?
(Not sure about this, the actual rule is probably more complicated.)

pranavk: That maybe the case for class names and other identifiers but this is inside the class.

include/souper/Infer/AbstractInterpreter.h
regehr: ugh, but ok

include/souper/Infer/AbstractInterpreter.h
regehr: can you take a stab at this @pranavk ?

include/souper/Infer/Interpreter.h
regehr: the generally accepted use for underscores is to find names for constructor parameters used only for initialization. don't ever do this for members or for local variables.

include/souper/Infer/Pruning.h
regehr: kill them please

lib/Infer/AbstractInterpreter.cpp
regehr: they should be changed

lib/Infer/AbstractInterpreter.cpp
regehr: this code is ok *only* if you guys are certain that we never synthesize a zero. if we can synthesize a zero (I can't tell for sure from the discussion above) then it needs to be changed.

and if we keep this code, it needs a comment stating the assumption it is based on. even so, it's pretty dangerous and nasty.

manasij7479: We can synthesize a zero, like in the example @zhengyangl has posted here.  
My argument is that there are no practical situations where synthesizing a zero produces an interesting RHS.

regehr: if we can shift by zero then this is wrong, correct? ```Result.Zero.setHighBits(2);```

manasij7479: If we prune using the assumption that the symbolic constant can not be zero, constant synthesis is avoided and we do not synthesize a zero.
The question here is: Is it a good idea to force synthesized constants to be non zero?
We discussed this a few months ago and came to the conclusion that it is a good idea.

regehr: all I want is this static analysis to be (1) sound with respect to the code and (2) documented, since it relies on a pretty subtle assumption

include/souper/Infer/AbstractInterpreter.h
pranavk: I just remembered that we also use this flag from the souper-interpret binary. The souper-interpret binary interprets first with partialEval off and second with on. If you think it's a useless feature, I will go ahead removing this PartialEval flag completely.

regehr: yes, I think this flag clutters up way too much code for the benefit it provides

lib/Infer/AbstractInterpreter.cpp
pranavk: yes, true. this is very inefficient. i have pushed a better version of it now.

lib/Infer/AbstractInterpreter.cpp
pranavk: i have documented this assumption in the code now. marking it as resolved.

pranavk:
@regehr Resolved all issues mentioned in the comment and added tests for things that could tested in this PR. You can review/merge this now.

include/souper/Infer/Pruning.h
```diff
@@ -25,37 +25,39 @@
 
 namespace souper {
 
-typedef std::function<bool(Inst *, std::vector<Inst *> &)> PruneFunc;
+typedef std::function<bool(Inst*, std::vector<Inst*>&)> PruneFunc;
```
regehr: please always prefer "Inst *" over "Inst*"

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -86,6 +86,7 @@ namespace {
 //   or look for feedback from solver, timeouts etc.
 // make sure path conditions work as expected
 // synthesize x.with.overflow
+// synthsize x.saturating
```
regehr: spelling error

regehr:
ok LGTM after a few tweaks, let me know when those are done

pranavk:
done. all squashed. ready to go

regehr:
thanks!

regehr:
but wait there are a couple of test fails, you need to look at those

pranavk:
huh! they were definetely passing earlier. I think I messed up when pushing or something; I will check this.

pranavk:
So, the test started failing after this commit (https://github.com/google/souper/commit/c5a88b02e3bac2e9d9c66364ab093d7d9824fc41)

PR #531, opened 2019-04-10
Make all constantsynthesis into one function

zhengyangl:


lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -728,6 +729,7 @@ void generateAndSortGuesses(InstContext &IC, Inst *LHS, SMTLIBSolver *Solver,
 
 typedef std::map<Inst*, std::vector<llvm::APInt>> InstConstList;
 
+#if 0
```
regehr: can you explain why you don't want to just delete this code?

regehr:
are there any more test cases you might be able to add, that test something that works now that failed before? maybe something with multiple constants?

lib/Infer/ExhaustiveSynthesis.cpp
zhengyangl: This function of code has the support of input specialization while the new code implements the G by S, I want to do more comparisons between these two

regehr:
anyway I'll merge, but if we can find some more constant synthesis tests we should add those

lib/Infer/ExhaustiveSynthesis.cpp
regehr: sounds good
in the future, if you plan to use the code again, protect it using "if (false)" instead of "#if 0" since then it has to pass the parser and typechecker

PR #536, opened 2019-04-15
precision testing of knownbits, negative, nonNeg, nonZero, power2, si…

jubitaneja:
…gnbits

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -409,6 +409,7 @@ set(SOUPER_PASS_PROFILE_ALL ${CMAKE_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}so
 set(PROFILE_LIBRARY ${CMAKE_BINARY_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}profileRuntime${CMAKE_STATIC_LIBRARY_SUFFIX})
 configure_file(${CMAKE_SOURCE_DIR}/utils/reduce.in ${CMAKE_BINARY_DIR}/reduce @ONLY)
 configure_file(${CMAKE_SOURCE_DIR}/utils/cache_dump.in ${CMAKE_BINARY_DIR}/cache_dump @ONLY)
+configure_file(${CMAKE_SOURCE_DIR}/utils/cache_dfa.in ${CMAKE_BINARY_DIR}/cache_dfa @ONLY)
 configure_file(${CMAKE_SOURCE_DIR}/utils/cache_import.in ${CMAKE_BINARY_DIR}/cache_import @ONLY)
 configure_file(${CMAKE_SOURCE_DIR}/utils/cache_infer.in ${CMAKE_BINARY_DIR}/cache_infer @ONLY)
 configure_file(${CMAKE_SOURCE_DIR}/utils/souper2llvm.in ${CMAKE_BINARY_DIR}/souper2llvm @ONLY)
--- a/include/souper/Extractor/Solver.h
+++ b/include/souper/Extractor/Solver.h
@@ -46,6 +46,36 @@ public:
                                                    const std::vector<InstMapping> &PCs,
                                                    Inst *LHS, InstContext &IC) = 0;
   virtual std::string getName() = 0;
+  virtual
+  std::error_code negative(const BlockPCs &BPCs,
+                           const std::vector<InstMapping> &PCs,
+                           Inst *LHS, bool &Negative,
+                           InstContext &IC) = 0;
+  virtual
+  std::error_code knownBits(const BlockPCs &BPCs,
+                            const std::vector<InstMapping> &PCs,
+                            Inst *LHS, llvm::KnownBits &Known,
+                            InstContext &IC) = 0;
+  virtual
+  std::error_code nonNegative(const BlockPCs &BPCs,
+                              const std::vector<InstMapping> &PCs,
+                              Inst *LHS, bool &NonNegative,
+                              InstContext &IC) = 0;
+  virtual
+  std::error_code powerTwo(const BlockPCs &BPCs,
+                           const std::vector<InstMapping> &PCs,
+                           Inst *LHS, bool &PowerTwo,
+                           InstContext &IC) = 0;
+  virtual
+  std::error_code nonZero(const BlockPCs &BPCs,
+                          const std::vector<InstMapping> &PCs,
+                          Inst *LHS, bool &NonZero,
+                          InstContext &IC) = 0;
+  virtual
+  std::error_code signBits(const BlockPCs &BPCs,
+                           const std::vector<InstMapping> &PCs,
+                           Inst *LHS, unsigned &SignBits,
+                           InstContext &IC) = 0;
 };
 
 std::unique_ptr<Solver> createBaseSolver(
--- a/lib/Extractor/Candidates.cpp
+++ b/lib/Extractor/Candidates.cpp
@@ -59,6 +59,30 @@ static llvm::cl::opt<bool> HarvestUses(
     "souper-harvest-uses",
     llvm::cl::desc("Harvest operands (default=false)"),
     llvm::cl::init(false));
+static llvm::cl::opt<bool> PrintNegAtReturn(
+    "print-neg-at-return",
+    llvm::cl::desc("Print negative dfa in each value returned from a function (default=false)"),
+    llvm::cl::init(false));
+static llvm::cl::opt<bool> PrintNonNegAtReturn(
+    "print-nonneg-at-return",
+    llvm::cl::desc("Print non-negative dfa in each value returned from a function (default=false)"),
+    llvm::cl::init(false));
+static llvm::cl::opt<bool> PrintKnownAtReturn(
+    "print-known-at-return",
+    llvm::cl::desc("Print known bits in each value returned from a function (default=false)"),
+    llvm::cl::init(false));
+static llvm::cl::opt<bool> PrintPowerTwoAtReturn(
+    "print-power-two-at-return",
+    llvm::cl::desc("Print power two dfa in each value returned from a function (default=false)"),
+    llvm::cl::init(false));
+static llvm::cl::opt<bool> PrintNonZeroAtReturn(
+    "print-non-zero-at-return",
+    llvm::cl::desc("Print non zero dfa in each value returned from a function (default=false)"),
+    llvm::cl::init(false));
+static llvm::cl::opt<bool> PrintSignBitsAtReturn(
+    "print-sign-bits-at-return",
+    llvm::cl::desc("Print sign bits dfa in each value returned from a function (default=false)"),
+    llvm::cl::init(false));
 
 extern bool UseAlive;
 
@@ -847,6 +871,61 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
   for (auto &BB : F) {
     std::unique_ptr<BlockCandidateSet> BCS(new BlockCandidateSet);
     for (auto &I : BB) {
+      if (isa<ReturnInst>(I)) {
+        if (PrintNegAtReturn) {
+          auto V = I.getOperand(0);
+          auto DL = F.getParent()->getDataLayout();
+          bool Negative = isKnownNegative(V, DL);
+          if (Negative)
+            llvm::outs() << "known negative from compiler: " << "true" << "\n";
+          else
+            llvm::outs() << "known negative from compiler: " << "false" << "\n";
+        }
+        if (PrintNonNegAtReturn) {
+          auto V = I.getOperand(0);
+          auto DL = F.getParent()->getDataLayout();
+          bool NonNegative = isKnownNonNegative(V, DL);
+          if (NonNegative)
+            llvm::outs() << "known nonNegative from compiler: " << "true" << "\n";
+          else
+            llvm::outs() << "known nonNegative from compiler: " << "false" << "\n";
+        }
+        if (PrintKnownAtReturn) {
+          auto V = I.getOperand(0);
+          auto DL = F.getParent()->getDataLayout();
+          unsigned Width = DL.getTypeSizeInBits(V->getType());
+          KnownBits Known(Width);
+          computeKnownBits(V, Known, DL);
+          llvm::outs() << "knownBits from compiler: " << Inst::getKnownBitsString(Known.Zero, Known.One) << "\n";
+        }
+        if (PrintPowerTwoAtReturn) {
+          auto V = I.getOperand(0);
+          auto DL = F.getParent()->getDataLayout();
+          bool PowerTwo = isKnownToBeAPowerOfTwo(V, DL);
+          if (PowerTwo)
+            llvm::outs() << "known powerOfTwo from compiler: " << "true" << "\n";
+          else
+            llvm::outs() << "known powerOfTwo from compiler: " << "false" << "\n";
+        }
+        if (PrintNonZeroAtReturn) {
+          auto V = I.getOperand(0);
+          auto DL = F.getParent()->getDataLayout();
+          bool NonZero = isKnownNonZero(V, DL);
+          if (NonZero)
+            llvm::outs() << "known nonZero from compiler: " << "true" << "\n";
+          else
+            llvm::outs() << "known nonZero from compiler: " << "false" << "\n";
+        }
+        if (PrintSignBitsAtReturn) {
+          auto V = I.getOperand(0);
+          auto DL = F.getParent()->getDataLayout();
+          unsigned NumSignBits = ComputeNumSignBits(V, DL);
+          if (NumSignBits > 1)
+            llvm::outs() << "known signBits from compiler: " << NumSignBits << "\n";
+          else
+            llvm::outs() << "known signBits from compiler: 1\n";
+        }
+      }
       // Harvest Uses (Operands)
       if (HarvestUses) {
         std::unordered_set<llvm::Instruction *> Visited;
--- a/lib/Extractor/Solver.cpp
+++ b/lib/Extractor/Solver.cpp
@@ -76,6 +76,162 @@ public:
   BaseSolver(std::unique_ptr<SMTLIBSolver> SMTSolver, unsigned Timeout)
       : SMTSolver(std::move(SMTSolver)), Timeout(Timeout) {}
 
+  bool testZeroMSB(const BlockPCs &BPCs,
+                   const std::vector<InstMapping> &PCs,
+                   Inst *LHS, InstContext &IC) {
+    unsigned W = LHS->Width;
+    Inst *Mask = IC.getConst(APInt::getOneBitSet(W, W-1));
+    InstMapping Mapping(IC.getInst(Inst::And, W, { LHS, Mask }), IC.getConst(APInt::getNullValue(W)));
+    bool IsSat;
+    std::error_code EC = SMTSolver->isSatisfiable(BuildQuery(IC, BPCs, PCs,
+                                                  Mapping, 0, /*Precondition=*/0),
+                                                  IsSat, 0, 0, Timeout);
+    if (EC) {
+      llvm::report_fatal_error("Error: SMTSolver->isSatisfiable() failed in testing zero MSB");
+      return false;
+    }
+    return !IsSat;
+  }
+
+  bool testOneMSB(const BlockPCs &BPCs,
+                  const std::vector<InstMapping> &PCs,
+                  Inst *LHS, InstContext &IC) {
+    unsigned W = LHS->Width;
+    Inst *Mask = IC.getConst(APInt::getOneBitSet(W, W-1));
+    InstMapping Mapping(IC.getInst(Inst::And, W, { LHS, Mask }), Mask);
+    bool IsSat;
+    std::error_code EC = SMTSolver->isSatisfiable(BuildQuery(IC, BPCs, PCs,
+                                                  Mapping, 0, /*Precondition=*/0),
+                                                  IsSat, 0, 0, Timeout);
+    if (EC) {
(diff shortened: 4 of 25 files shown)


jubitaneja:
This is WIP, adding test cases and refactoring some parts of code.

include/souper/Extractor/Solver.h
```diff
@@ -41,6 +41,36 @@ class Solver {
...
   virtual std::string getName() = 0;
+  virtual
+  std::error_code negative(const BlockPCs &BPCs,
+                           const std::vector<InstMapping> &PCs,
+                           Inst *LHS, llvm::APInt &Negative,
+                           InstContext &IC) = 0;
+  virtual
+  std::error_code knownBits(const BlockPCs &BPCs,
+                            const std::vector<InstMapping> &PCs,
+                            Inst *LHS, llvm::APInt &Zeros, llvm::APInt &Ones,
```
regehr: let's use llvm::KnownBits instead of a pair of APInts

include/souper/Extractor/Solver.h
```diff
@@ -41,6 +41,36 @@ class Solver {
...
+                            InstContext &IC) = 0;
+  virtual
+  std::error_code nonNegative(const BlockPCs &BPCs,
+                            const std::vector<InstMapping> &PCs,
+                            Inst *LHS, llvm::APInt &NonNegative,
```
regehr: I don't think it's necessary to allocate an APInt for facts like NonNegative, this should just be a bool

include/souper/Extractor/Solver.h
```diff
@@ -41,6 +41,36 @@ class Solver {
...
+                            InstContext &IC) = 0;
+  virtual
+  std::error_code powerTwo(const BlockPCs &BPCs,
+                            const std::vector<InstMapping> &PCs,
```
regehr: careful with indentation here and elsewhere -- make sure your browser has no tabs

lib/Extractor/Candidates.cpp
```diff
@@ -847,6 +871,64 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
...
     for (auto &I : BB) {
+      if (PrintNegAtReturn && isa<ReturnInst>(I)) {
+        auto V = I.getOperand(0);
+        auto DL = F.getParent()->getDataLayout();
+        bool Negative = 0;
+        Negative = isKnownNegative(V, DL);
+        if (Negative)
+          llvm::outs() << "known at return: " << "(negative)" << "\n";
+        else
+          llvm::outs() << "known at return: " << "" << "\n";
```
regehr: how about just print something like: "known negative?" followed by true or false? otherwise all these blank "known at return:" are hard to interpret

lib/Extractor/Candidates.cpp
```diff
@@ -847,6 +871,64 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
...
+      }
+      if (PrintNonNegAtReturn && isa<ReturnInst>(I)) {
+        auto V = I.getOperand(0);
+        auto DL = F.getParent()->getDataLayout();
+        bool NonNegative = 0;
```
regehr: false, not 0

lib/Extractor/Candidates.cpp
```diff
@@ -847,6 +871,64 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
...
+        NonNegative = isKnownNonNegative(V, DL);
```
regehr: but also no initial value needed if you're just going to overwrite it

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,164 @@ class BaseSolver : public Solver {
...
+    return !IsSat;
+  }
+
+  std::error_code negative(const BlockPCs &BPCs,
+                           const std::vector<InstMapping> &PCs,
+                           Inst *LHS, APInt &Negative,
+                           InstContext &IC) override {
+    unsigned W = LHS->Width;
+    Negative = APInt::getNullValue(W);
+    APInt NegativeGuess = Negative | APInt::getOneBitSet(W, W-1);
```
regehr: why do you want to do math using a variable that you just set to zero?

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,164 @@ class BaseSolver : public Solver {
...
+                              InstContext &IC) override {
+    unsigned W = LHS->Width;
+    NonNegative = APInt::getNullValue(W);
+    APInt NonNegativeGuess = NonNegative | APInt::getOneBitSet(W, W-1);
+    if (testZeroSign(BPCs, PCs, NonNegativeGuess, LHS, IC)) {
+      NonNegative = APInt::getNullValue(W);
+    } else if (testOneSign(BPCs, PCs, NonNegativeGuess, LHS, IC)) {
+      NonNegative = NonNegativeGuess;
+    } else {
+      NonNegative = NonNegativeGuess;
```
regehr: it looks like this else branch and the previous "else if" branch contain the same code, that cannot be correct

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,164 @@ class BaseSolver : public Solver {
...
+      InstMapping Mapping(Guess, True);
+      bool IsSat;
+      std::error_code EC = SMTSolver->isSatisfiable(BuildQuery(IC, BPCs, PCs, Mapping, 0, /*Precondition=*/0),
+                                                    IsSat, 0, 0, Timeout);
+      if (EC)
+        llvm::report_fatal_error("stopping due to error");
+
+      if (!IsSat) {
+        SignBits = I;
+        continue;
```
regehr: continue not needed

CMakeLists.txt
```diff
@@ -409,6 +409,12 @@ set(SOUPER_PASS_PROFILE_ALL ${CMAKE_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}so
 set(PROFILE_LIBRARY ${CMAKE_BINARY_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}profileRuntime${CMAKE_STATIC_LIBRARY_SUFFIX})
 configure_file(${CMAKE_SOURCE_DIR}/utils/reduce.in ${CMAKE_BINARY_DIR}/reduce @ONLY)
 configure_file(${CMAKE_SOURCE_DIR}/utils/cache_dump.in ${CMAKE_BINARY_DIR}/cache_dump @ONLY)
+configure_file(${CMAKE_SOURCE_DIR}/utils/cache_knownbits.in ${CMAKE_BINARY_DIR}/cache_knownbits @ONLY)
+configure_file(${CMAKE_SOURCE_DIR}/utils/cache_neg.in ${CMAKE_BINARY_DIR}/cache_neg @ONLY)
+configure_file(${CMAKE_SOURCE_DIR}/utils/cache_nonNeg.in ${CMAKE_BINARY_DIR}/cache_nonNeg @ONLY)
+configure_file(${CMAKE_SOURCE_DIR}/utils/cache_power.in ${CMAKE_BINARY_DIR}/cache_power @ONLY)
+configure_file(${CMAKE_SOURCE_DIR}/utils/cache_nonZero.in ${CMAKE_BINARY_DIR}/cache_nonZero @ONLY)
+configure_file(${CMAKE_SOURCE_DIR}/utils/cache_signbits.in ${CMAKE_BINARY_DIR}/cache_signbits @ONLY)
```
regehr: it seems like these are all very similar scripts, can't they be collapsed into one?

jubitaneja:
Thanks for the review comments. I will fix it soon.

Jubi

jubitaneja:
I have fixed and updated this PR. souper2llvm in this PR has changes to support dataflow facts only. About creating one big script for testing dataflow facts -- we can do that, but this will be a bit complicated to read and debug if required later. Maybe keeping individual scripts is an easy way to use it (just in my opinion). If you think we should really have just one perl script, i can spend some time putting it all together.

regehr:
I thought you were going to move the souper2llvm changes into a different pull request

regehr:
also I don't want a bunch of different scripts for inferring dataflow facts from a cache, you need to merge these. this code is impossible to maintain

jubitaneja:
I kept the minimal souper2llvm changes in this PR that only takes care of translating dataflow facts in Souper IR to LLVM IR. I will take away every change in that case and put it all in next PR. Also, that would mean we cannot test dataflow facts on LLVM side using commands like: 
` %souper2llvm %s | %llvm-as | %souper %solver -print-known-at-return`

jubitaneja:
Sure, will merge all scripts of caching into one.

lib/Extractor/Candidates.cpp
```diff
@@ -847,6 +871,64 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
   for (auto &BB : F) {
     std::unique_ptr<BlockCandidateSet> BCS(new BlockCandidateSet);
     for (auto &I : BB) {
+      if (PrintNegAtReturn && isa<ReturnInst>(I)) {
+        auto V = I.getOperand(0);
+        auto DL = F.getParent()->getDataLayout();
+        bool Negative;
```
regehr: in this case and others, you should prefer this style:
```bool Negative = isKnownNegative(V, DL);```

Comment (regehr) [reply lib/Extractor/Candidates.cpp]: of course "auto" would be fine too

File: lib/Extractor/Candidates.cpp [hunk 12]
```diff
@@ -847,6 +871,64 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
...
+        auto V = I.getOperand(0);
+        auto DL = F.getParent()->getDataLayout();
+        bool PowerTwo;
+        PowerTwo = isKnownToBeAPowerOfTwo(V, DL);
+        if (PowerTwo)
+          llvm::outs() << "known powerOfTwo from compiler: " << "true" << "\n";
+        else
+          llvm::outs() << "known powerOfTwo from compiler: " << "false" << "\n";
+      }
+      if (PrintNonZeroAtReturn && isa<ReturnInst>(I)) {
```
regehr: instead of these repeated tests of ```isa<ReturnInst>``` just test this once at the beginning and then test the individual flags

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,157 @@ class BaseSolver : public Solver {
   BaseSolver(std::unique_ptr<SMTLIBSolver> SMTSolver, unsigned Timeout)
       : SMTSolver(std::move(SMTSolver)), Timeout(Timeout) {}
 
+  bool testZeroMSB(const BlockPCs &BPCs,
+                    const std::vector<InstMapping> &PCs,
```
regehr: indentation problem

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,157 @@ class BaseSolver : public Solver {
...
+    return !IsSat;
+  }
+
+  std::error_code negative(const BlockPCs &BPCs,
+                           const std::vector<InstMapping> &PCs,
+                           Inst *LHS, bool &Negative,
+                           InstContext &IC) override {
+    unsigned W = LHS->Width;
+    Negative = false;
+    if (testZeroMSB(BPCs, PCs, LHS, IC))
```
regehr: I don't understand this, why are you calling a function and then using its result to set Negative to false, when you just initialized Negative to false one line earlier in the code?

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,157 @@ class BaseSolver : public Solver {
...
+
+  std::error_code nonNegative(const BlockPCs &BPCs,
+                              const std::vector<InstMapping> &PCs,
+                              Inst *LHS, bool &NonNegative,
+                              InstContext &IC) override {
+    unsigned W = LHS->Width;
+    NonNegative = false;
+    if (testZeroMSB(BPCs, PCs, LHS, IC)) {
+      NonNegative = true;
+    } else if (testOneMSB(BPCs, PCs, LHS, IC)) {
```
regehr: same here

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,157 @@ class BaseSolver : public Solver {
...
+      APInt OneGuess = Known.One | APInt::getOneBitSet(W, I);
+      if (testKnown(BPCs, PCs, Known.Zero, OneGuess, LHS, IC))
+        Known.One = OneGuess;
+    }
+    return std::error_code();
+  }
+
+  std::error_code powerTwo(const BlockPCs &BPCs,
+                           const std::vector<InstMapping> &PCs,
+                           Inst *LHS, APInt &PowTwo,
```
regehr: PowTwo should be a bool, not an APInt

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,157 @@ class BaseSolver : public Solver {
...
+    Inst *Zero = IC.getConst(APInt(W, 0, false));
+    Inst *True = IC.getConst(APInt(1, 1, false));
+    Inst *PowerTwoInst = IC.getInst(Inst::And, 1, {IC.getInst(Inst::Ne, 1, {LHS, Zero}),
+                                                   IC.getInst(Inst::Eq, 1, {PowerMask, Zero})});
+    InstMapping Mapping(PowerTwoInst, True);
+    bool IsSat;
+    std::error_code EC = SMTSolver->isSatisfiable(BuildQuery(IC, BPCs, PCs, Mapping, 0, /*Precondition=*/0),
+                                                  IsSat, 0, 0, Timeout);
+    if (EC)
+      llvm::report_fatal_error("stopping due to error");
```
regehr: in this case and others, say something useful when an error occurs, ```stopping due to error``` is not useful to someone running this code

lib/Extractor/Solver.cpp
```diff
@@ -76,6 +76,157 @@ class BaseSolver : public Solver {
...
+    if (!IsSat)
+      PowTwo = APInt(1, 1, false);
+    else
+      PowTwo = APInt(1, 0, false);
+    return std::error_code();
+  }
+
+  std::error_code nonZero(const BlockPCs &BPCs,
+                          const std::vector<InstMapping> &PCs,
+                          Inst *LHS, APInt &NonZero,
```
regehr: bool

test/Infer/precision-kb-2.opt
```diff
@@ -0,0 +1,11 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-known-bits %s
+; RUN: %souper2llvm %s | %llvm-as | %souper %solver -print-known-at-return
+
+; CHECK: knownBits from souper: xxxxxxx0
+; CHECK: knownBits from compiler: xxxxxxxx
```
regehr: it's great to test that souper returns the precise result, but do not test that the compiler returns the imprecise result, that is not something that we want our tests relying upon

test/Infer/precision-neg-1.opt
```diff
@@ -0,0 +1,12 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-neg %s
+; RUN: %souper2llvm %s | %llvm-as | %souper %solver -print-neg-at-return
+
+; CHECL: known negative from Souper:   true
```
regehr: mis-spelled word CHECL

test/Infer/precision-neg-2.opt
```diff
@@ -0,0 +1,12 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-neg %s
+; RUN: %souper2llvm %s | %llvm-as | %souper %solver -print-neg-at-return
+
+; CHECL: known negative from Souper:   true
```
regehr: same

test/Infer/precision-nonNeg-1.opt
```diff
@@ -0,0 +1,17 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-non-neg %s
+; RUN: %souper2llvm %s | %llvm-as | %souper %solver -print-nonneg-at-return
+
+; CHECL: known nonNegative from Souper:   true
```
regehr: same

test/Infer/precision-nonNeg-2.opt
```diff
@@ -0,0 +1,13 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-non-neg %s
+; RUN: %souper2llvm %s | %llvm-as | %souper %solver -print-nonneg-at-return
+
+; CHECL: known nonNegative from Souper:   true
```
regehr: mis-spelled word CHECL

test/Infer/precision-nonNeg-3.opt
```diff
@@ -0,0 +1,11 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-non-neg %s
+
+; CHECL: known nonNegative from Souper:   true
```
regehr: mis-spelled word CHECL

tools/souper-check.cpp
```diff
@@ -96,6 +121,93 @@ int SolveInst(const MemoryBufferRef &MB, Solver *S) {
...
+      bool Negative;
+      if (std::error_code EC = S->negative(Rep.BPCs, Rep.PCs, Rep.Mapping.LHS,
+                                           Negative, IC)) {
+        llvm::errs() << EC.message() << '\n';
+      }
+      std::string s;
+      if (Negative)
+        s = "true";
+      else
+        s = "false";
```
regehr: avoid this repeated code for making a string out of a bool. either factor it out into a helper function or else use the native C++ method for printing a bool as a string.

test/Infer/precision-kb-2.opt
jubitaneja: I added this check to only test what LLVM returns, but not with an intent to compare the precision of Souper vs LLVM. But, I will take off this LLVM check part from here and add relevant tests for LLVM where we can expect some useful dataflow information.

tools/souper-check.cpp
jubitaneja: std::boolalpha didn't work in context of llvm::outs(), so I added a small helper function.

jubitaneja:
Thanks for the detailed feedback. Updated the PR again.

test/Infer/precision-kb-5.opt
```diff
@@ -0,0 +1,19 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-known-bits %s
+; RUN: %souper2llvm %s | %llvm-as | %souper %solver -print-known-at-return
+
+
+; CHECK: knownBits from souper: 00000000000000xxxxxxxxxxxxxxxxxx
+; CHECK: knownBits from compiler: 0000000000000xxxxxxxxxxxxxxxxxxx
```
regehr: don't test what the compiler returns

test/Infer/precision-neg-1.opt
```diff
@@ -0,0 +1,12 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-neg %s
+; RUN: %souper2llvm %s | %llvm-as | %souper %solver -print-neg-at-return
+
+; CHECK: known negative from Souper:   true
+; CHECK: known negative from compiler: false
```
regehr: don't test what the compiler returns

test/Infer/precision-power-3.opt
```diff
@@ -0,0 +1,12 @@
+; REQUIRES: solver
+; RUN: %souper-check %solver -infer-rhs -infer-known-bits %s
+; RUN: %souper2llvm %s | %llvm-as | %souper %solver -print-power-two-at-return
+
+; CHECK: known powerOfTwo from souper:   true
+; CHECK: known powerOfTwo from compiler: false
```
regehr: fix all of these

tools/souper-check.cpp
```diff
@@ -96,6 +128,74 @@ int SolveInst(const MemoryBufferRef &MB, Solver *S) {
...
+      }
+      llvm::outs() << "known powerOfTwo from souper: " << convertToStr(PowTwo) << "\n";
+      return 0;
+    }
+
+    if (InferNonZero) {
+      bool NonZero;
+      if (std::error_code EC = S->nonZero(Rep.BPCs, Rep.PCs, Rep.Mapping.LHS,
+                                              NonZero, IC)) {
+        llvm::errs() << EC.message() << '\n';
```
regehr: in all of these functions you must check the error code, not just print it out. the dataflow result is not valid when there's an error, so the code can't just keep going.

jubitaneja: Fixed.

regehr:
LGTM, thanks!

LebedevRI:
> I kept the minimal souper2llvm changes in this PR that only takes care of translating dataflow facts in Souper IR to LLVM IR. I will take away every change in that case and put it all in next PR. Also, that would mean we cannot test dataflow facts on LLVM side using commands like:
> ` %souper2llvm %s | %llvm-as | %souper %solver -print-known-at-return`

I keep oscillating between writing too much words, and not enough words it seems.
I was suggesting making the `souper2llvm.in` fixes *FIRST*, because currently that tool is still broken.
Really looking forward to the next PR with those changes i guess..

(Related: is static optimization profile working?
`cache_dump` tells me it is always zero, even though `SOUPER_STATIC_PROFILE` is set)

jubitaneja:
@LebedevRI Thanks for checking in on souper2llvm status. @zhengyangl and I are working on it, will soon send a PR for review. In case, you want to have a look at current souper2llvm fixes only, you can check this branch: https://github.com/jubitaneja/souper/tree/fix-souper2llvm

LebedevRI:
Thanks!

> About static profile, I will have to check again. I remember using this script a couple of months ago and it worked fine.

Yes, that is my recollection, too..

PR #537, opened 2019-04-15
Setup infra for exhaustive CR testing

pranavk:
Taken shamefully from https://github.com/regehr/llvm-dataflow-research/blob/master/test-xfer.cpp

The PR depends on #530 ; For review purposes, only look the latest commit for now. There are still some changes to be made in testing infra before it's fully ready for review though.

### DIFF
--- a/lib/Infer/AbstractInterpreter.cpp
+++ b/lib/Infer/AbstractInterpreter.cpp
@@ -87,8 +87,8 @@ namespace {
   KnownBits intersect(const KnownBits &KB1, const KnownBits &KB2) {
     // ugh the constructor we want is private
     KnownBits Res(KB1.getBitWidth());
-    Res.Zero = KB1.Zero | KB2.Zero;
-    Res.One = KB1.One | KB2.One;
+    Res.Zero = KB1.Zero & KB2.Zero;
+    Res.One = KB1.One & KB2.One;
     return Res;
   }
 } // anonymous
--- a/lib/Infer/AbstractInterpreter.cpp
+++ b/lib/Infer/AbstractInterpreter.cpp
@@ -45,10 +45,81 @@ namespace {
     return Max;
   }
 
+ ConstantRange KBToCR(const KnownBits &KB) {
+    ConstantRange CR1(getSMin(KB), getSMax(KB));
+    ConstantRange CR2(getSMin(KB), getSMax(KB));
+    return CR1.getSetSize().ult(CR2.getSetSize()) ? CR1 : CR2;
+  }
 
+  KnownBits CRToKB(const ConstantRange &CR) {
+    KnownBits Res(CR.getBitWidth());
+    if (CR.isFullSet())
+      return Res;
+    if (CR.isEmptySet()) {
+      Res.setAllZero();
+      return Res;
+    }
+    if (CR.isWrappedSet())
+      return Res; // TODO look for opportunities in wrapped case
+    const APInt &L = CR.getLower();
+    const APInt &U = CR.getUpper();
+    for (unsigned I = 0; I < L.getBitWidth(); ++I) {
+      // initial guess is bit from lower bound
+      bool V = L[I];
+      APInt Boundary = L;
+      // gotta be multiple faster ways to do this...
+      for (unsigned J = 0; J < I; ++J)
+        Boundary.setBit(J);
+      if (U.ugt(Boundary + 1))
+        continue;
+      if (V)
+	Res.One.setBit(I);
+      else
+	Res.Zero.setBit(I);
+    }
+    return Res;
+  }
+
+  unsigned numKnown(const KnownBits &KB) {
+    return KB.Zero.countPopulation() + KB.One.countPopulation();
+  }
+
+  KnownBits intersect(const KnownBits &KB1, const KnownBits &KB2) {
+    // ugh the constructor we want is private
+    KnownBits Res(KB1.getBitWidth());
+    Res.Zero = KB1.Zero | KB2.Zero;
+    Res.One = KB1.One | KB2.One;
+    return Res;
+  }
 } // anonymous
 
 namespace souper {
+
+  // approximate reduced product for known bits + constant range
+  void improveKBCR(KnownBits &KB, ConstantRange &CR) {
+    while (1) {
+      // measure
+      APInt SS = CR.getSetSize();
+      unsigned NK = numKnown(KB);
+      // mutual shootdown
+      auto KB2 = CRToKB(CR);
+      auto CR2 = KBToCR(KB);
+      CR = CR.intersectWith(CR2);
+      KB = intersect(KB, KB2);
+      // done when no more improvements
+      assert(CR.getSetSize().ule(SS));
+      assert(numKnown(KB) >= NK);
+      if (SS == CR.getSetSize() && NK == numKnown(KB))
+	break;
+    }
+  }
+
+  bool KnownBitsAnalysis::isConflictingKB(const KnownBits &A, const KnownBits &B) {
+    if ((A.One & B.Zero) != 0 || (A.Zero & B.One) != 0) {
+      return true;
+    }
+    return false;
+  }
 
   KnownBits KnownBitsAnalysis::getMostPreciseKnownBits(KnownBits A, KnownBits B) {
     unsigned unknownCountA =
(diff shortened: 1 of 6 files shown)


regehr:
wow, this is a lot of changes, I did not realize this would be so much work!!

regehr:
thanks for all the code style fixes, yikes!

unittests/Interpreter/InterpreterInfra.h
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    template<int SetSize>
+    llvm::KnownBits abstractizeKB(std::bitset<SetSize> &ResultSet) {
+      llvm::KnownBits ResultKB(WIDTH);
+      if (ResultSet.none())
+	return ResultKB;
+
+      unsigned I = 0;
+      while (!ResultSet[I++]);
```
regehr: I don't understand this loop, or the use of its results

unittests/Interpreter/InterpreterInfra.h
```diff
@@ -0,0 +1,153 @@
...
+      }
+      return ResultSet;
+    }
+
+    template<int SetSize>
+    llvm::ConstantRange abstractizeCR(std::bitset<SetSize> &ResultSet) {
+      llvm::ConstantRange ResultCR(WIDTH, false);
+      for (unsigned I = 0; I < SetSize; I++) {
+	if (ResultSet[I]) {
+	  ResultCR = ResultCR.unionWith(llvm::APInt(WIDTH, I));
```
regehr: I don't think this is going to give the most precise result, can you look back at my code and see if there's a better way to do this? I believe I had a different approach

regehr: the problem is ConstantRange isn't a lattice and there's no unique LUB of arbitrary sets of elements

unittests/Interpreter/InterpreterInfra.h
```diff
@@ -0,0 +1,153 @@
...
+
+      std::bitset<SetSize> ResultSet(1);
+      for (int Cnt = 0; Cnt < SetSize; Cnt++) {
+
+	llvm::KnownBits ValueKB;
+	auto Tmp = llvm::APInt(KB.getBitWidth(), Cnt);
+	ValueKB.One = Tmp;
+	ValueKB.Zero = ~Tmp;
+
+	if (KnownBitsAnalysis::isConflictingKB(ValueKB, KB))
```
regehr: how can there be a conflict given that you initialized One to Tmp and Zero to ~Tmp?

unittests/Interpreter/InterpreterInfra.h
pranavk: ResultSet is a bitvector. It's bit is 1 if the element is present and 0 otherwise. Here, we are finding the first element which is present and then initializing the value of {And,Or}Result later on with this index.

unittests/Interpreter/InterpreterInfra.h
pranavk: `Tmp` and `ValueKB` is the APInt, KnownBits representation respectively of the number `Cnt`. I am checking if this KnownBits representation of the number conflicts with the `KB` which is the KnownBits object we want to concretize. 

There would be some value of `Cnt` which will not agree with the `KnownBits` value that we are concretizing.

unittests/Interpreter/InterpreterInfra.h
pranavk: >  can you look back at my code and see if there's a better way to do this? I believe I had a different approach

I am not aware of any code you wrote related to abstractizing ConstantRanges. Your  code that I copied to this PR was just related to  CR KB reduction.

regehr:
is my improveKBCR() sound? is it precise?

unittests/Interpreter/InterpreterTests.cpp
```diff
@@ -25,345 +27,69 @@
...
-    }
-    report_fatal_error("faulty setLowest!");
-  }
 
-  KnownBits clearLowest(KnownBits x) {
-    for (int i = 0; i < x.getBitWidth(); i++) {
-      if (!x.Zero[i] && !x.One[i]) {
-        x.Zero.setBit(i);
-        return x;
+      if (!ExhaustiveCR.contains(CalculatedCR)) {
```
regehr: are you sure this is right? shouldn't it be the CalculatedCR that contains the ExhaustiveCR?

the way to think about this is: the ExhaustiveCR must be the smallest one, so why would it contain a worse one?

regehr: also this contains property (whichever way it should go) doesn't necessarily hold since the constantrange abstract domain isn't a lattice

unittests/Interpreter/InterpreterInfra.h
regehr: but why? why is this the right thing to do?

unittests/Interpreter/InterpreterInfra.h
regehr: https://github.com/regehr/llvm-dataflow-research/blob/master/test-xfer.cpp#L74

regehr:
OK this needs some more work, please try to get it done before you leave or else while you're away

unittests/Interpreter/InterpreterInfra.h
pranavk: It's the right thing to do because we want first non-zero bit to initialize And/Or Result with. Eg: If WIDTH=4, the result set will be 16 bits long. Let's say that only number 4 and 12 are present in the ResultSet, and we want to abstractize it. In this case, the ResultSet will be : `0000100000001000`. We want to get the AndResult and OrResult of 4 and 12 which means we quickly want to access the first non-zero bit so that we can initialize `AndResult` and `OrResult` with "some" number and then keep ANDing or ORing as we encounter more non-zero bits in the ResultSet.

unittests/Interpreter/InterpreterInfra.h
pranavk: So, this thing (the link pasted) that you are talking about is already taken care of in the `exhaustiveCR` function. See the call to `bestCR` in that function. 

The quoted code you have commented against is just abstractizing the ResultSet for CR. I believe the LLVM's `unionWith` function, which looks quite elaborate and correct, to be enough for this purpose. Are you saying that LLVM's `unionWith` function won't give the most precise constant range given set of numbers?

pranavk:
> is my improveKBCR() sound? is it precise?

It's unsound. Right now, I don't mark the test as failure if it's imprecise; let's just make sure that it's atleast sound.

Here's the results:

```
Unsound!! CR KB reduction for CR
Original KB: ????
Original CR: [1,0)
CalculatedCR: [-8,7)
ExhaustiveCR: [1,0)
../unittests/Interpreter/InterpreterTests.cpp:87: Failure
Value of: false
  Actual: false
Expected: true
```

unittests/Interpreter/InterpreterTests.cpp
pranavk: oh, yes, you are right. Let me fix that.

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -45,11 +45,82 @@ namespace {
...
+	Res.Zero.setBit(I);
+    }
+    return Res;
+  }
 
+  unsigned numKnown(const KnownBits &KB) {
+    return KB.Zero.countPopulation() + KB.One.countPopulation();
+  }
+
+  KnownBits intersect(const KnownBits &KB1, const KnownBits &KB2) {
```
regehr: wait, this looks like a union operation, not an intersection. please double-check this code @pranavk

regehr: I mean that I would expect "intersect" to return the abstraction of the intersection of the concretization sets, but this function returns the abstraction of the union of the concretization sets

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -45,11 +45,82 @@ namespace {
...
+      KB = intersect(KB, KB2);
+      // done when no more improvements
+      assert(CR.getSetSize().ule(SS));
+      assert(numKnown(KB) >= NK);
+      if (SS == CR.getSetSize() && NK == numKnown(KB))
+	break;
+    }
+  }
+
+  bool KnownBitsAnalysis::isConflictingKB(const KnownBits &A, const KnownBits &B) {
```
regehr: wait there's already a public member function hasConflict(), that should be used instead of this, right?

regehr: oops nevermind

unittests/Interpreter/InterpreterInfra.cpp
```diff
@@ -0,0 +1,527 @@
...
+  } while(nextKB(x));
+
+  return true;
+}
+
+
+// CRTesting Implementation
+// -----------------------------
+
+bool CRTesting::ok(const ConstantRange &R, const bool Table[]) {
```
regehr: let's name this something better than "ok"

lib/Infer/AbstractInterpreter.cpp
pranavk: yeah, i just took this code as it is from what you had in your branch[1]. But indeed, this says intersection but is actually a union. Anyhow, I will fix it as you confirmed that this is incorrect.

>  I mean that I would expect "intersect" to return the abstraction of the intersection of the concretization sets, but this function returns the abstraction of the union of the concretization sets

It takes two abstract KnownBits as parameters, not concretization sets. So, it just returns KB abstraction from two KB abstractions.

[1] https://github.com/regehr/souper/commit/cc95254608aeef5c8038b602d0a8a77ba3394c2f#diff-e2d75e847ce8ee98960879bf86c6ea33R95

pranavk:
After fixing the intersect bug that you commented above, here's the new unsoundness results:

```
Unsound!! CR KB reduction for CR
Original KB: ????
Original CR: [0,-8)
CalculatedCR: [0,7)
ExhaustiveCR: [0,-8)
../unittests/Interpreter/InterpreterTests.cpp:87: Failure
Value of: false
  Actual: false
Expected: true
```

unittests/Interpreter/InterpreterInfra.h
regehr: I'm telling you that there is no most precise constant range

unittests/Interpreter/InterpreterInfra.h
regehr: this is just a weird way to do it Pranav, you can make this a lot simpler:
- eliminate the first loop
- in the second loop, just and together the ones (as you are doing) and then also and together the zeros, and don't flip the bits at the end

please do this.

regehr: the first loop is not just useless but also confusing

pranavk: okay. i have simplified this. see the updated PR

lib/Infer/AbstractInterpreter.cpp
regehr: so how could any of this code be working at all if this basic operation was wrong? please figure out what is going on here, I don't want to merge code that isn't working at all.

in other words, if the testing code had a very basic mistake and it wasn't caught via the testing code, then this testing code is bogus and needs to be fixed. how can we be confident in this code now? this is all making me much less confident in it.

regehr: "It takes two abstract KnownBits as parameters, not concretization sets. So, it just returns KB abstraction from two KB abstractions."

I'm talking about the specification of the code, not the implementation

pranavk: > so how could any of this code be working at all if this basic operation was wrong? please figure out what is going on here, I don't want to merge code that isn't working at all.

wait, this function `intersect` was used by `improveKBCR` function, part of the changes I picked from your branch, *only*. And testing infrastructure never said that `improveKBCR` is sound.

> in other words, if the testing code had a very basic mistake and it wasn't caught via the testing code, then this testing code is bogus and needs to be fixed. how can we be confident in this code now? this is all making me much less confident in it.

since this bug was only in the function used by `improveKBCR` function, and testing code never promised that `improveKBCR` is correct, I don't think we can conclude that testing code is bogus (yet :).

regehr:
thanks!

PR #538, opened 2019-04-16
Count Trunc/SExt/ZExt as regular synthesis operators

zhengyangl:
* Now after ReservedInsts gets the widths assigned, we turn it into "Inst::Hole". This is similar to the conversion between ReservedConst and Vars. This makes the code clear to read.


regehr:
cool!

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -128,19 +128,6 @@ void hasConstant(Inst *I, std::set<Inst *> &ConstSet) {
   hasConstantHelper(I, Visited, ConstSet);
 }
 
-// TODO small vector by reference?
-std::vector<Inst *> matchWidth(Inst *I, unsigned NewW, InstContext &IC) {
```
regehr: so happy to see this code get deleted, I hated it as soon as I wrote it!

regehr:
LGTM!

PR #541, opened 2019-04-22
Fixed the bug which leads to the misses of Unary operator guesses in exhaustive synthesis

zhengyangl:
After the trunc/sext/zext patch, syn-ctpop now cost around 5 min.


regehr:
LGTM thanks!

PR #545, opened 2019-05-06
Constant Range Support

zhengyangl:


regehr:
I don't understand the parts of this patch that are doing stuff with demanded bits, why is that in here?

zhengyangl:
@regehr Before this PR, all the constants we're synthesizing stay in the RHSs. But in constant range synthesis, we now have reserved constants in LHSs: For query [(x<= LHS') && (x<LHS'+c) == True], here, (x<= LHS) && (x< LHS+c) is the LHS and True is the RHS. Since the reserved constant now occurs in LHS, after the first query, we need to call getInstCopy() on LHS to fill it with the model returned from the solver. This triggers the bug where the demandedbits are not initialized properly in function getPhi() and getInstCopy(), which leads to wrong synthesis results.

zhengyangl:
Before this pull request, the model filling only works on RHS, and the bug is hidden because we're not having phi's synthesized.

zhengyangl:
Also I've thought about reverse the query, like [True == (X<= LHS') && (X<LHS'+c)], where True as LHS and (x<= LHS') && (x<LHS'+c). By setting the query this way we can eliminate the extra getInstCopy() call on LHS. However, the refinement check are needed here and the direction matters. The reversed query can lead to over-conservative results.

lib/Extractor/Solver.cpp
```diff
@@ -472,6 +472,94 @@ class BaseSolver : public Solver {
     return !IsSat;
   }
 
+  std::error_code testRange(const BlockPCs &BPCs,
+                            const std::vector<InstMapping> &PCs,
+                            Inst *LHS, llvm::APInt &C,
+                            llvm::APInt &X,
```
regehr: can you find a better name for X and SynthesisX?

zhengyangl: How about naming X as "Result" and naming SynthesisX as "ReservedResult"?

lib/Extractor/Solver.cpp
```diff
@@ -472,6 +472,94 @@ class BaseSolver : public Solver {
...
+      IsFound = false;
+    } else {
+      IsFound = true;
+      X = ResultMap[SynthesisX];
+    }
+
+    return EC;
+  }
+
+  std::error_code range(const BlockPCs &BPCs,
```
regehr: can you make this function have a better name too?

lib/Extractor/Solver.cpp
```diff
@@ -472,6 +472,94 @@ class BaseSolver : public Solver {
...
+    APInt BinSearchResultX, BinSearchResultC;
+    bool BinSearchHasResult = false;
+
+    while (L.ule(R)) {
+      APInt M = L + ((R - L)).lshr(1);
+      APInt BinSearchX;
+      bool Found = false;
+      EC = testRange(BPCs, PCs, LHS, M, BinSearchX, Found, IC);
+      if (EC) {
+        llvm::report_fatal_error("stopping due to error");
```
regehr: does this ever happen in practice?

lib/Extractor/Solver.cpp
```diff
@@ -472,6 +472,94 @@ class BaseSolver : public Solver {
...
+      }
+      if (Found) {
+        R = M - 1;
+
+        // record result
+        BinSearchResultX = BinSearchX;
+        BinSearchResultC = M;
+        BinSearchHasResult = true;
+      } else {
+        if (L == R) break;
```
regehr: put break on a new line please

regehr:
looks great except for a few minor changes

lib/Extractor/Solver.cpp
zhengyangl: Sorry, it wont happen, after calling testRange, either we found the range or we don't. I'll remove this line.

lib/Extractor/Solver.cpp
regehr: sounds good

zhengyangl:
Fixed the issues, please review @regehr

regehr:
LGMT, thanks!

PR #547, opened 2019-05-07
only update range when the source has the attribute

zhengyangl:
The variable range in parser is initialized to full set, however, no matter the source defines the range or not, the variable will be assigned to range [Lower, Upper). Since Lower and Upper are initialized to 0, if source does not define range, the range will be assigned to [0, 0), which is the empty set.
This patch forces it to change the range only when the range is defined in the source file.


regehr:
thanks!

PR #548, opened 2019-05-11
Let the constant synthesis caller decide how to handle solver errors

zhengyangl:
Instead of terminate immediately, let the constant synthesis caller decide how to handle solver error. Before this patch, souper terminates immediately when the a single solver query times out.


regehr:
great, thanks, we should keep eyes open for lazy code that terminates too easily when something outside our control goes wrong

PR #549, opened 2019-05-17
refactor dataflow code and print range info

jubitaneja:


jubitaneja:
This patch refactors the dataflow code, and also prints range information as computed by LLVM in the extractor. This was missing from the previous PR on constant range.

jubitaneja:
@zhengyangl Please review this as well. It refactors the code and adds the missing case for printing range computed by LLVM.

lib/Extractor/Candidates.cpp
```diff
@@ -860,6 +864,70 @@ std::tuple<BlockPCs, std::vector<InstMapping>> GetRelevantPCs(
...
 }
 
+void PrintDataflowInfo(Function &F, Instruction &I, LazyValueInfo *LVI,
+                       ScalarEvolution *SE) {
+  auto V = I.getOperand(0);
+  auto DL = F.getParent()->getDataLayout();
+  if (PrintNegAtReturn) {
+    bool Negative = isKnownNegative(V, DL);
+    if (Negative)
+      llvm::outs() << "known negative from compiler: " << "true" << "\n";
```
regehr: don't keep duplicating strings like this, just print the first part of this before the conditional

lib/Extractor/Candidates.cpp
```diff
@@ -860,6 +864,70 @@ std::tuple<BlockPCs, std::vector<InstMapping>> GetRelevantPCs(
...
+      if (Instruction *I = dyn_cast<Instruction>(V)) {
+        BasicBlock *BB = I->getParent();
+        auto LVIRange = LVI->getConstantRange(V, BB);
+        auto SC = SE->getSCEV(V);
+        auto R1 = LVIRange.intersectWith(SE->getSignedRange(SC));
+        auto R2 = LVIRange.intersectWith(SE->getUnsignedRange(SC));
+        Range = R1.getSetSize().ult(R2.getSetSize()) ? R1 : R2;
+      }
+    }
+    llvm::outs() << "known range from compiler: [" << Range.getLower() << ","
```
regehr: just range, not "known range"

lib/Extractor/Candidates.cpp
```diff
@@ -860,6 +864,70 @@ std::tuple<BlockPCs, std::vector<InstMapping>> GetRelevantPCs(
...
+    bool NonZero = isKnownNonZero(V, DL);
+    if (NonZero)
+      llvm::outs() << "known nonZero from compiler: " << "true" << "\n";
+    else
+      llvm::outs() << "known nonZero from compiler: " << "false" << "\n";
+  }
+  if (PrintSignBitsAtReturn) {
+    unsigned NumSignBits = ComputeNumSignBits(V, DL);
+    if (NumSignBits > 1)
+      llvm::outs() << "known signBits from compiler: " << NumSignBits << "\n";
```
regehr: only say "known" for the actual knownBits analysis, it's part of the name of that analysis, but it's not part of the name of the other ones

jubitaneja: sure, will fix it. I used known as a verb here, but I agree it gets confusing with the actual analysis name.

lib/Extractor/Candidates.cpp
```diff
@@ -861,6 +865,75 @@ std::tuple<BlockPCs, std::vector<InstMapping>> GetRelevantPCs(
   return std::make_tuple(RelevantBPCs, RelevantPCs);
 }
 
+void PrintDataflowInfo(Function &F, Instruction &I, LazyValueInfo *LVI,
+                       ScalarEvolution *SE) {
+  auto V = I.getOperand(0);
+  auto DL = F.getParent()->getDataLayout();
+  if (PrintNegAtReturn) {
+    bool Negative = isKnownNegative(V, DL);
+    std::string s = "negative from compiler: ";
```
regehr: fine but why not just print this right here? why make another variable??

lib/Extractor/Candidates.cpp
```diff
@@ -861,6 +865,75 @@ std::tuple<BlockPCs, std::vector<InstMapping>> GetRelevantPCs(
...
+    std::string s = "nonZero from compiler: ";
+    if (NonZero)
+      llvm::outs() << s << "true" << "\n";
+    else
+      llvm::outs() << s << "false" << "\n";
+  }
+  if (PrintSignBitsAtReturn) {
+    unsigned NumSignBits = ComputeNumSignBits(V, DL);
+    std::string s = "signBits from compiler: ";
+    if (NumSignBits > 1)
```
regehr: why is this test needed?

lib/Extractor/Candidates.cpp
```diff
@@ -861,6 +865,71 @@ std::tuple<BlockPCs, std::vector<InstMapping>> GetRelevantPCs(
...
 
+void PrintDataflowInfo(Function &F, Instruction &I, LazyValueInfo *LVI,
+                       ScalarEvolution *SE) {
+  auto V = I.getOperand(0);
+  auto DL = F.getParent()->getDataLayout();
+  if (PrintNegAtReturn) {
+    bool Negative = isKnownNegative(V, DL);
+    llvm::outs() << "negative from compiler: ";
+    if (Negative)
+      llvm::outs() << "true\n";
```
regehr: ok this is getting there, thanks.

there's one more very easy way to simplify this code, please see here: http://www.cplusplus.com/reference/ios/boolalpha/

just set this flag at the start of PrintDataflowInfo() and clear it at the end

jubitaneja: Thanks for sharing this flag. Hope I have used it the right way.

zhengyangl: @regehr std::boolalpha won't work together with llvm::errs() and outs().

regehr:
ok, good
see how much cleaner and nicer the code is when you do this right?

jubitaneja:
Yes. :)

PR #550, opened 2019-05-28
Refactor exhaustive synthesis

zhengyangl:


manasij7479:
LGTM

regehr:
LGTM but remember that we have poor test cases for exhaustive synthesis, and should work on improving that, or else it becomes hard to have confidence in patches that touch that code

PR #554, opened 2019-06-28
Fix #553

### LINKED ISSUE #553: Incorrect handling on intrinsic arguments
Hi,

In [line 535](https://github.com/google/souper/blob/master/lib/Extractor/Candidates.cpp#L535) of `lib/Extractor/Candidates.cpp`, not every intrinsic functions have at least one argument. Thus sometime it will crash due to out-of-bound argument fetching.

I think we should be more generalized here by either checking the amount of operands or fetching intrinsic arguments only when needed.

Thank you

mshockwave:
Check number of operands before fetching intrinsic arguments.


regehr:
thank you!

PR #560, opened 2019-08-24
Fixed a bug which leads to error copy of constant in enumeration synthesize

zhengyangl:
The bug is caused by missing handling of Const in instJoin() function. Before the commit the Const is getting copied by getInst, which caused the loss of its value. 

The patch also added the missing cache in instJoin().


regehr:
LGTM, thanks Zhengyang!

PR #565, opened 2019-09-03
Restricted bits precision testing

manasij7479:
Run with: ./interpreter_tests --check-rb-precision


regehr:
LGTM but in the future, for this kind of situation a global flag is cleaner that passing this around everywhere

PR #568, opened 2019-09-04
Bump alive version

manasij7479:


regehr:
can we just use the upstream instead of your fork?

manasij7479:
> can we just use the upstream instead of your fork?

That's the long term plan, I guess.

PR #571, opened 2019-09-04
this is count pruning, not cost pruning

pranavk:


pranavk:
@manasij7479 this code was added by you. I don't think //TODO comment is relevant anymore. Also, i renamed the function to reflect exactly what it's meant to do.

regehr:
wait a second, why do we want to do this at all? what's going on here @manasij7479?

PR #573, opened 2019-09-07
Misc alive backend changes, switch z3 to master

manasij7479:


build_deps.sh
```diff
@@ -27,7 +27,7 @@ klee_branch=pure-bv-qf-llvm-7.0
 alive_commit=9823174bb34fcb9c8e33c37e7e04d46bfe3a29a5
 alive_repo=https://github.com/manasij7479/alive2.git
 z3_repo=https://github.com/Z3Prover/z3.git
-z3_branch=z3-4.8.4
+z3_branch=master
```
regehr: don't switch to master, switch to a specific commit
if you switch to master, our tool randomly changes behavior as z3 gets commits

regehr:
this is fine except for the one thing

build_deps.sh
manasij7479: Fixed.

regehr:
thanks!

PR #575, opened 2019-09-09
Prune sub, shift ops with same operands

pranavk:


lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -303,6 +302,16 @@ void getGuesses(std::vector<Inst *> &Guesses,
         if (K == Inst::Sub && V2->SynthesisConstID != 0)
           continue;
 
+        // PRUNE: don't synthesize "{shl, ashr, lshr} x, x" as its results in poison
```
regehr: comment isn't accurate, this operation results in either 0 or poison

pranavk: fixed it now

regehr:
LGTM

PR #576, opened 2019-09-09
Disable undef input for alive2

manasij7479:


regehr:
since this is unsound, just doing it unconditionally is bad
if it's absolutely necessary to default to unsound behavior, then you should make souper print a warning about unsoundness and also you need a flag to go back to the right behavior

PR #580, opened 2019-09-11
misc changes makes test suite happy again

zhengyangl:


test/Infer/alive-cegis.opt
```diff
@@ -1,4 +1,4 @@
-; REQUIRES: solver, synthesis
+; REQUIRES: solver, synthesis, unsupported
```
zhengyangl: Will work on this very soon.

lib/Infer/ExhaustiveSynthesis.cpp
```diff
@@ -166,6 +166,9 @@ void getGuesses(std::vector<Inst *> &Guesses,
     }
   }
 
+  Inst *I1 = IC.getReservedInst();
```
zhengyangl: This reverts some wrong changes made recently. A case failed because of that change.

regehr:
thanks!!

PR #582, opened 2019-09-11
Simplify enum syn

pranavk:
Okay. I missed this piece of code which was already pruning if both operands are same. It's better to put new saturating and shift operators in this list rather than a new pruning if condition


regehr:
LGTM thanks

PR #584, opened 2019-09-12
Rely on ConcreteInterpreter instead of KLEE backend for LHS simplification

zhengyangl:


zhengyangl:
This patch also adds the support of simplification of phi nodes in constant synthesis.

manasij7479:
Check hasValue() before ConcreteInterpreter::getValue(), just in case evaluation fails (/produces poison, undef, etc).

PR #587, opened 2019-09-12
fix illegal return width for cmp operators in restricted bits analysis

pranavk:


pranavk:
@regehr pushed a test for this that crashes souper-check atleast in Debug build (some assertion failure in LLVM). I don't have a release build to test yet but i think it will crash release build as well down the line.

Can update you again on the results of release build if you prefer that. Otherwise, feel free to merge this.

regehr:
nah, I don't care about release mode, just wanted something that triggers this bug, thanks!

PR #589, opened 2019-09-13
KB transfer functions for funnel shift ops

manasij7479:


lib/Infer/AbstractInterpreter.cpp
```diff
@@ -144,6 +144,25 @@ namespace souper {
...
 
+  llvm::KnownBits concatKnownBits(llvm::KnownBits A, llvm::KnownBits B) {
+    auto W0 = A.Zero.getBitWidth();
+    auto W1 = B.Zero.getBitWidth();
+
+    llvm::KnownBits Result(W0 + W1);
+    Result.Zero = 0;
+    Result.One = 0;
+
+    Result.Zero |= A.Zero;
```
regehr: want to make this a little simpler? instead of initialize to zero and then OR it with some stuff, just initialize with the stuff you want there in the first place

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -715,10 +734,32 @@ namespace souper {
...
     }
-//   case FShl:
-//     return "fshl";
-//   case FShr:
-//     return "fshr";
+    case Inst::FShl: {
+      auto NewKB0 = concatKnownBits(KB0, KB1);
+      if (KB2.isConstant()) {
+        auto Shift = KB2.getConstant();
+        NewKB0.Zero = NewKB0.Zero.rotl(Shift);
```
regehr: this does not look right. funnel shift does not concatenate and then rotate.

regehr: this would have been caught if you had done the testing

regehr:
what is the holdup on this?

manasij7479:
Have fixed the transfer functions but haven't implemented the testing yet
(the plumbing to exhaustively test ternary instructions is absent).
Will do it soon and submit.

lib/Infer/AbstractInterpreter.cpp
manasij7479: done

manasij7479:
This is ready for review now

regehr:
LGTM thanks!

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -715,10 +723,38 @@ namespace souper {
...
-//     return "fshl";
-//   case FShr:
-//     return "fshr";
+    case Inst::FShl: {
+      auto NewKB0 = concatKnownBits(KB0, KB1);
+      if (KB2.isConstant()) {
+        auto Shift = KB2.getConstant().urem(I->Width);
+        NewKB0.Zero = NewKB0.Zero.shl(Shift);
+        NewKB0.One = NewKB0.One.shl(Shift);
+        Result = NewKB0.trunc(I->Width);
```
pranavk: Are you sure this is correct? 

LangRef for fshl: 
"and the most significant bits are extracted to produce a result"

And `trunc` removes the most significant bits. LangRef for `trunc` says:
"The ‘trunc’ instruction truncates the high order bits in value "

manasij7479:
Yeah, this was wrong. Will fix.

lib/Infer/AbstractInterpreter.cpp
pranavk: I have filed #631 for this so that we don't forget about it.

PR #592, opened 2019-09-13
null pointer derefs found in performance testing

jubitaneja:


regehr:
please add regression tests triggering these

zhengyangl:
This only happens when using souper functions within ValueTracking, since these information are not available in an analysis pass. However, it is all correct if we're harvesting from Souper's own pass.

zhengyangl:
I think we may just keep another Souper branch for performance test. We need to keep the behavior of crash in master. It can tells something wrong happened when these analysis results from other passes are missing instead of silently continue and reason imprecisely.

regehr:
LGTM, thanks

jubitaneja:
@zhengyangl I agree, we should have these in a separate branch.

PR #593, opened 2019-09-13
fix precision of powerTwo, add FileCheck for precision tests, and no …

jubitaneja:
…infer-rhs for DFA computation


regehr:
LGTM thanks

PR #597, opened 2019-09-13
Some improvements to gdb pretty printer

pranavk:
After using it for couple of days, it's useful but it also prints
inst* pointer that appear in backtrace which is annoying sometimes.

Let's change the behavior to print the instruction only when
dereferenced. So you have to do `p *I` to pretty print it, instead of
just doing `p I`


regehr:
LGTM, thanks!

PR #598, opened 2019-09-13
Support for overflow instructions in enumerative synthesis

pranavk:


pranavk:
@regehr @zhengyangl a review would be great.

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -216,27 +226,30 @@ void getGuesses(std::vector<Inst *> &Guesses,
...
+          (*I)->K == Inst::ReservedConst)
         continue;
 
       // Prune: I1 should only be the first argument
       if ((*I)->K == Inst::ReservedInst && (*I) != I1)
         continue;
 
       // PRUNE: don't try commutative operators both ways
-      auto Start = Inst::isCommutative(K) ? I : Comps.begin();
+      auto Start = (Inst::isCommutative(K) || Inst::isOverflowIntrinsicMain(K) || Inst::isOverflowIntrinsicSub(K)) ? I : Comps.begin();
```
regehr: why not just make these intrinsics commutative instead of adding a special case here?

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -207,7 +217,7 @@ void getGuesses(std::vector<Inst *> &Guesses,
   Comps.push_back(I2);
 
   for (auto K : BinaryOperators) {
-    if (Inst::isCmp(K) && Width != 1)
+    if ((Inst::isCmp(K) || Inst::isOverflowIntrinsicSub(K)) && Width != 1)
```
regehr: wait, why only subtract here?

lib/Infer/EnumerativeSynthesis.cpp
regehr: also, avoid lines >80 columns

pranavk: Because sadd.with.overflow is actually not commutative in internal souper representation. It's first argument is basic `add` operation and second argument is one among those AddO variants. However, it is commutative in enumerative synthesis algorithm and on a souper IR level. Hence, this special case.

lib/Infer/EnumerativeSynthesis.cpp
pranavk: This is not subtract. This is "Sub" component which refers to "AddO" variants. See this comment: https://github.com/google/souper/pull/598/files/ac7ad6e88b7fba740b87a1a41c265f6e9537a8e3#diff-a263ca9d697aae31475ef84f4ec04418R51-R54 for explanation how these are treated differently.

lib/Infer/EnumerativeSynthesis.cpp
regehr: are all of these lines <= 80 columns?

lib/Inst/Inst.cpp
```diff
@@ -447,11 +447,17 @@ const char *Inst::getKindName(Kind K) {
   case Hole:
     return "hole";
   case SAddO:
+    return "saddo";
```
zhengyangl: Are these operators supposed to be printed?

zhengyangl:
Just an open question, can you teach Souper to prefer non-overflow instructions if there is no use of the overflow bit?

regehr:
this should be easy to do with the cost model.

the question I have is about the fact that these instructions produce two results, whereas souper always optimizes a single value. it's likely the case that we're going to generate an add-with-overflow and not eliminate the bare add. perhaps gvn or another LLVM pass will clean that up for us. experiments are needed.

lib/Inst/Inst.cpp
pranavk: no, i don't think so. The reason i added this here is to help in debugging where sometimes i want to print a DAG.

lib/Infer/EnumerativeSynthesis.cpp
pranavk: i will fix this before merging, if not.

pranavk:
> this should be easy to do with the cost model.

yeah, the current cost model is setup such that sadd.with.overflow and extractvalue has a cost of 1 each. So if overflow bit is not used, the total cost of this thing will be 2 while cost of normal add would be 1. Add should be prioritized against sadd.with.overflow.

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -114,12 +124,25 @@ namespace {
...
+    auto NSExt = IC.getInst(Inst::SExt, TargetWidth, { RHS });
+    auto NZExt = IC.getInst(Inst::ZExt, TargetWidth, { RHS });
+    addGuess(NSExt, TargetWidth, IC, MaxCost, Guesses, TooExpensive);
+    addGuess(NZExt, TargetWidth, IC, MaxCost, Guesses, TooExpensive);
+  }
+  else if (TargetWidth < RHS->Width) {
+    auto NTrunc = IC.getInst(Inst::Trunc, TargetWidth, { RHS });
+    addGuess(NTrunc, TargetWidth, IC, MaxCost, Guesses, TooExpensive);
+  }
+  else {
```
regehr: plz always put else on the same line as the closing curly brace of the previous block

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -114,12 +124,25 @@ namespace {
...
+void addGuess(Inst *RHS, unsigned TargetWidth, InstContext &IC, int MaxCost, std::vector<Inst *> &Guesses,
               int &TooExpensive) {
-  if (souper::cost(RHS) < MaxCost)
-    Guesses.push_back(RHS);
-  else
-    TooExpensive++;
+  if (TargetWidth > RHS->Width) {
+    auto NSExt = IC.getInst(Inst::SExt, TargetWidth, { RHS });
+    auto NZExt = IC.getInst(Inst::ZExt, TargetWidth, { RHS });
+    addGuess(NSExt, TargetWidth, IC, MaxCost, Guesses, TooExpensive);
```
regehr: why do you have a recursive addguess call before incrementing the cost? this seems wrong

pranavk: what do you mean "before incrementing the cost"? These recursive calls are a way to convert an existing about-to-be-added guess which doesn't matches the `TargetWidth`. The recursive call will make sure that the new converted guess is still less in cost than the LHS

lib/Pass/Pass.cpp
```diff
@@ -80,6 +80,22 @@ static const bool DynamicProfileAll = false;
...
+  Value* getOperand(Inst* I, unsigned index, Instruction *ReplacedInst,
+                    ExprBuilderContext &EBC, DominatorTree &DT,
+                    std::map<Inst *, Value *> &ReplacedValues,
+                    IRBuilder<> &Builder, Module *M) {
+    Value *Result = nullptr;
+    if (Inst::isOverflowIntrinsicMain(I->K)) {
+      assert(I->Ops.size() == 2 && I->Ops[0]->Ops.size() == 2);
+      Result = getValue(I->Ops[0]->Ops[index], ReplacedInst, EBC, DT, ReplacedValues, Builder, M);
+    }
+    else {
```
regehr: else on same line as closing brace please

lib/Infer/EnumerativeSynthesis.cpp
regehr: ok good

regehr:
LGTM thanks-- will merge when tests pass

PR #601, opened 2019-09-18
precision testing of demanded bits

jubitaneja:


jubitaneja:
@zhengyangl Adding you as a reviewer as well because you've worked on this stuff.

jubitaneja:
I will create a separate PR for souper2llvm script to handle demanded bits and a demanded bits precision testing script that uses s2l.

lib/Extractor/Candidates.cpp
```diff
@@ -935,6 +939,22 @@ void ExtractExprCandidates(Function &F, const LoopInfo *LI, DemandedBits *DB,
     for (auto &I : BB) {
       if (isa<ReturnInst>(I))
         PrintDataflowInfo(F, I, LVI, SE);
+      if (PrintDemandedBitsAtReturn && I.getType()->isIntegerTy()) {
+        if (auto BO = dyn_cast<BinaryOperator>(&I)) {
+          auto AddOp = BO->getOpcode();
+          if (AddOp == Instruction::Add) {
```
regehr: I don't understand this line
why are you looking specifically for add instructions?

jubitaneja: The API 'getDemandedBits' works on an Instruction and to avoid computing demanded bits for each instruction in a given LLVM IR, I special-cased it to 'add x, 0' instruction only. To handle this, for a given LHS in Souper IR, all input vars are translated to 'add with zero' instruction first in souper2llvm. For instance, 

Input Souper IR:
%0:i8 = var
%1:i1 = slt %0, 0:i8
infer %1

is translated to LLVM IR as:
define i1 @foo(i8 %add_0, i64 %_phiinput) {
entry:
  %var_0 = add i8 %add_0, 0
  %xxx0 = icmp slt i8 %var_0, 0
  ret i1 %xxx0
}
And, the above code in Extractor will only compute demanded bits for instruction labelled by '%var_0'.

regehr: ok, that's kind of a weird hack
how about explain this in detail in a comment?

jubitaneja: I agree it's a hack, but I will add a detailed comment.

regehr: fyi the better way to do this would have been using metadata, which is easy to insert and look at

regehr: you don't have to do that here, but at least document that it's a nasty hack

jubitaneja: I have moved the details in a commit message. About the metadata, sure that would be an appropriate approach. But this piece of code will only be used when we run a precision testing experiment. If this hacky thing is not ok, I will fix it with the metadata approach.

regehr:
LGTM, thanks!

PR #604, opened 2019-09-19
Respect MaxNumInstructions parameters more strictly

pranavk:
The description of this parameter says that it is a way to bound the
maximum number of synthesized instructions but it was never respected.
We were calling instCount on a RHS which will also count the components
reused from the LHS.

### DIFF
--- a/include/souper/Inst/Inst.h
+++ b/include/souper/Inst/Inst.h
@@ -270,6 +270,7 @@ struct SynthesisContext {
 };
 
 int cost(Inst *I, bool IgnoreDepsWithExternalUses = false);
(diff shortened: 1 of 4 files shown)


pranavk:
Discussed with @zhengyangl on slack and we agreed `holes` should be counted as 1.

pranavk:
Ok. So before this patch, a test which should never have passed was passing without any problem. It was synthesizing 2 instrs even though we only asked it to synthesize one. This PR catching this bug provides some assurance about the changes in this PR :)

regehr:
looks good plz squash

pranavk:
@regehr squashed now.

regehr:
great, thanks

PR #606, opened 2019-09-20
make extractvalue 2nd arg a 32bit const

pranavk:
we emit a 32 bit constant when transforming llvm bc to souper IR. so
let's be consistent here as well.


regehr:
thanks!

PR #607, opened 2019-09-24
KB testing for Select

manasij7479:


unittests/Interpreter/InterpreterInfra.cpp
```diff
@@ -223,6 +247,62 @@ bool KBTesting::nextKB(llvm::KnownBits &x) {
   return false;
 }
 
+bool testKB(llvm::KnownBits Calculated, EvalValueKB Expected,
+            Inst::Kind K, std::vector<llvm::KnownBits> KBS) {
+    // expected value is poison/ub; so let binary transfer functions do
```
regehr: plz remove spurious space in front of //

unittests/Interpreter/InterpreterInfra.cpp
```diff
@@ -223,6 +247,62 @@ bool KBTesting::nextKB(llvm::KnownBits &x) {
...
+    llvm::KnownBits y(Op1W);
+    do {
+      llvm::KnownBits z(Op2W);
+      do {
+        InstContext IC;
+        auto Op0 = IC.getInst(Inst::Var, Op0W, {});
+        auto Op1 = IC.getInst(Inst::Var, Op1W, {});
+        auto Op2 = IC.getInst(Inst::Var, Op2W, {});
+        auto I = IC.getInst(K, WIDTH, {Op0, Op1, Op2});
+        std::unordered_map<Inst *, llvm::KnownBits> C{{Op0, x}, {Op1, y}, {Op2, z}};
```
regehr: it doesn't really matter in this kind of testing code, so you don't need to fix, but creating a map in a tight loop is pretty inefficient, something like a SmallMapVector would be better since it won't hit the allocator

unittests/Interpreter/InterpreterInfra.cpp
manasij7479: fixed

regehr:
LGTM thanks

PR #609, opened 2019-09-25
Double check synthesis results with alive

manasij7479:


regehr:
great, please add a test or two using this feature, or else modify an existing test or two to do so

manasij7479:
done

regehr:
LGTM, thanks! this is definitely a good safety check, we should turn it on during some of our synthesis runs

PR #611, opened 2019-09-27
LLVM 9

manasij7479:


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ execute_process(
   OUTPUT_VARIABLE LLVM_SYSTEM_LIBS
   OUTPUT_STRIP_TRAILING_WHITESPACE
 )
-set(LLVM_LIBS "${LLVM_LIBS} ${LLVM_SYSTEM_LIBS}")
+set(LLVM_LIBS "${LLVM_LIBS} ${LLVM_SYSTEM_LIBS} ${CMAKE_SOURCE_DIR}/third_party/llvm/${CMAKE_BUILD_TYPE}/lib/libclangASTMatchers.a")
```
regehr: this isn't the right solution, see how we set LLVM_LIBS in the first place and see if there's a flag that can be added there which gets the correct libraries

manasij7479: llvm-config does not seem to have a relevant flag

regehr:
plz also bump z3 to latest release

CMakeLists.txt
regehr: well, I assert that there is a solution to this which is different from bypassing llvm-config completely

lib/Infer/Pruning.cpp
```diff
@@ -14,7 +14,7 @@
 
 #include "souper/Infer/AbstractInterpreter.h"
 #include "souper/Infer/Pruning.h"
-
+#include "souper/Extractor/Candidates.h"
```
regehr: what necessitated this change?

manasij7479: Added a getSetSize declaration in that header.

PR #612, opened 2019-09-28
prevent cttz, ctlz, bitreverse on bitreverse

pranavk:


regehr:
I think you should explain this patch more, I would have imagined that doing what the title of this patch says is much simpler than this

regehr:
like all of the other ad-hoc pruning rules are 1 or 2 lines, what's going on here?

pranavk:
This pruning rule is different in nature from all other ad-hoc pruning rules because it depends on 2 instructions - we have to prune situations where a bitreverse result being is used by cttz, ctlz, bitreverse or ctpop instruction. Other ad-hoc pruning rules were only concerned about a single instruction (had nothing to do with uses, etc. of that instruction).

I also factored out the other code in this PR which bumped up the size of this PR slightly. All the main pruning code is in EnumerativeSynthesis.cpp

pranavk:
What do you think about it?

regehr:
sorry, keep trying to find time to take a closer look, maybe today!

regehr:
LGTM thanks!

PR #615, opened 2019-10-09
Phi should also take poison from their chosen input

zhengyangl:
This fixes the `LLVM ERROR: the model returned from second query evaluates to poison for LHS` bug.


regehr:
ack!!! correct comment and incorrect code :)

regehr:
LGTM thanks!

Issue #616, opened 2019-10-09
Constant Synthesize failure


zhengyangl:
Opened this issue just for reference. 

Constant synthesis cannot handle the following opt.

```
%0:i128 = var ; 1
%1:i128 = uadd.sat %0, %0
%2:i128 = sub %1, %0
infer %2
%3:i128 = reservedconst
%4:i128 = ashr %0, %3
%5:i128 = xor %0, %4
result %5
```

PR #617, opened 2019-10-09
Query Structure Fix

zhengyangl:
This fixes the issue #616.
Consider the first query in constant synthesize CEGIS loop:

```(lhs-ub-free => (lhs != rhs && rhs-ub-free && precondition)) == 0```

This query is problematic: look at the right hand side of the implication, `(lhs != rhs && rhs-ub-free && precondition)`, it is conjunction of three parts: `lhs != rhs`, `rhs-ub-free` and `precondition`. Any one of these three parts evaluating to false will lead to a first query model. Some cases like this one in issue #616, most guesses for the reserved constant leads to poison in RHS, thus `rhs-ub-free` is likely to be false. As a result, most models generated from the first query are garbage and later get discarded in the second query.

The cause of this error is the negation, when the query preprocessor do negation, it is not enough to just negate the `(LHS == RHS)` part, but it needs to negate the whole right hand side of the implication. After applying the negation correctly, the first query looks like

```(lhs-ub-free => (lhs != rhs || !rhs-ub-free || !precondition)) == 0```

The model returned from this new query will not contain any guesses that evaluate to UB in RHS; also it handles the precondition correctly.

In general, this patch can boost the constant synthesize performance when RHS is likely to be evaluated to poison. The attached test case shows the efficiency of the change.

### DIFF
--- a/lib/Extractor/ExprBuilder.cpp
+++ b/lib/Extractor/ExprBuilder.cpp
@@ -951,21 +951,22 @@ Inst *ExprBuilder::GetCandidateExprForReplacement(
   if (RHSUB == LIC->getConst(llvm::APInt(1, false)))
     return nullptr;
 
-  if (Negate) // (LHS != RHS)
-    Result = LIC->getInst(Inst::Ne, 1, {LHS, RHS});
-  else        // (LHS == RHS)
-    Result = LIC->getInst(Inst::Eq, 1, {LHS, RHS});
-
+  Result = LIC->getInst(Inst::Eq, 1, {LHS, RHS});
   if (Precondition)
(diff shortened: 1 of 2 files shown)


regehr:
this is a great catch @zhengyangl 
hey @rsas are you available to give this a quick look?
regardless, I want to make sure I understand this all before merging, will do it soon

regehr:
LGTM, thanks again, this is great work

rsas:
LGTM, this part of Souper had issues with poison on the RHS as well, have you guys addressed these?

zhengyangl:
Hi @rsas , could you please explain the issues with poison on RHS in detail? I would like to have those issues fixed :)

PR #618, opened 2019-10-11
Fix input association in alive translation

manasij7479:


include/souper/Infer/AliveDriver.h
```diff
@@ -44,6 +44,8 @@ class AliveDriver {
   Inst *LHS, *PreCondition;
 
   Cache LExprCache, RExprCache;
+  std::map<const Inst *, std::string> NameMap;
+  void copyInputs(Cache& From, Cache &To, IR::Function &RHS);
```
regehr: not that big of a deal, but you need to consistently put the & (or *) next to the name, not next to the type

lib/Infer/AliveDriver.cpp
```diff
@@ -433,10 +433,23 @@ souper::AliveDriver::synthesizeConstantsWithCegis(souper::Inst *RHS, InstContext
   return ConstMap;
 }
 
+void souper::AliveDriver::copyInputs(souper::AliveDriver::Cache& From,
+                                         souper::AliveDriver::Cache& To,
```
regehr: actually just fix all these quickly, please

include/souper/Infer/AliveDriver.h
manasij7479: fixed

lib/Infer/AliveDriver.cpp
manasij7479: fixed

regehr:
LGTM thanks!

PR #620, opened 2019-10-12
no sense to put ::Select here (it's a ternary op)

pranavk:
Also, make the comment order same as code


regehr:
so where are we pruning ```select x, y, y``` now?

pranavk:
in the ternaryOp for loop just below this one. Select is not contained in BinaryOp variable anyways, so this couldn't have done anything.

regehr:
great!

PR #621, opened 2019-10-12
support for alive2 overflow instrs

pranavk:


lib/Infer/AliveDriver.cpp
```diff
@@ -530,12 +536,21 @@ std::string getUniqueName() {
...
   }
 
+  auto Ops = I->Ops;
+  if (souper::Inst::isOverflowIntrinsicMain(I->K)) {
+    Ops = Ops[0]->Ops;
+  }
+
   FunctionBuilder Builder(F);
-  for (auto &&Op : I->Ops) {
+  for (auto Op : Ops) {
```
regehr: plz explain why you chose to make a copy of each op instead of using rvalue references

pranavk:
Oh, yeah, this can be changed now. I was doing a different approach first
and then changed to another approach. This removal of rvalue ref is remnant
of the previous approach. I will fix it

lib/Infer/AliveDriver.cpp
pranavk: fixed it.

lib/Infer/AliveDriver.cpp
```diff
@@ -530,12 +536,21 @@ std::string getUniqueName() {
 bool souper::AliveDriver::translateAndCache(const souper::Inst *I,
                                             IR::Function &F,
                                             Cache &ExprCache) {
+  if (souper::Inst::isOverflowIntrinsicSub(I->K)) {
```
regehr: it seems odd to special-case this one instruction, what's going on here?

pranavk: we are special casing out the SAddO instruction as they are not real instructions. The selected function checks if the given instruction kind is any of those 'AddO' type.

regehr: ok-- add a brief comment to this effect, that these are souper-internal things that we don't bother alive with

pranavk: done

regehr:
LGTM, thanks!

PR #622, opened 2019-10-13
DAG matching to prevent joining redundant instructions

pranavk:
We have several pruning rules in our enumerative algorithm that prevents us from enumerating on silly combinations. An example of such a rule would be `sub %0, %0`. The pruning rules work great but these pruning rules are not respected when holes are involved and we have to join two instructions. 

Consider this example:

```
%0:i8 = var ; 0
%1:i1 = hole
%2:i1 = hole
%3:i1 = usub.sat %1, %2
infer %2
```

%1 can take the form of trunc %0 and then %2 can also take the same form afterwards. The problem is that when joining the instruction, the instruction joiner is not smart enough to know that it would lead to a silly instruction at %3 (for which a pruning rule exists) where we will be substracting the same thing from the same thing.

One way to solve this problem which also seems correct is to fix instruction joiner such that whenever a new form is taking the place of a hole, we check if that form has already been present in the current 
candidate. If it's already present, enumerative algorithm should already have taken care of reusing that operand as many times as required or it wouldn't have reused it if pruning rule exist to prevent same operand from being used (like our previous example). 

DAG matching:
To be able to accomplish this, I did a DAG matching in the candidate. I will take the new instruction just generated and try to find if an "equivalent" of it already exists in the candidate containing the hole. All the instruction are matched by their op type and all the terminal insts like reserved const, holes, and variables are matched by their address. If a DAG is found in the tree, we ignore the guess and move on.

This is a O(mn) algorithm where m = number of nodes in haystack DAG and n = number of nodes in needle DAG. Since in our case m and n will mostly be small, we should be fine with this.

### DIFF
--- a/lib/Inst/Inst.cpp
+++ b/lib/Inst/Inst.cpp
@@ -1174,9 +1174,56 @@ Inst *souper::getInstCopy(Inst *I, InstContext &IC,
   return Copy;
 }
 
+bool isTerminalInst(Inst *I) {
+  bool terminal = false;
+  switch (I->K) {
+    case Inst::Const:
+    case Inst::UntypedConst:
+    case Inst::Var:
+    case Inst::Phi:
+    case Inst::Hole:
+      terminal = true;
+  }
+
+  return terminal;
+}
+
+// Return true if NewInst is equivalent to I
+bool isEquivalent(Inst *NewInst, Inst *I) {
+  // bfs on smaller instruction (NewInst)
+  if (I->K != NewInst->K || I->Ops.size() != NewInst->Ops.size())
+    return false;
+
+  if (isTerminalInst(I) && NewInst != I) {
+    return false;
+  }
+
+  for (int i = 0; i < I->Ops.size(); i++) {
+    if (!isEquivalent(I->Ops[i], NewInst->Ops[i]))
+      return false;
+  }
+
+  return true;
+}
+
+// Return true if NewInst is present in Cand.
+bool DAGMatch(Inst *NewInst, Inst *Cand) {
+  std::vector<Inst*> Results;
+  souper::findInsts(Cand, Results, [NewInst](Inst *I) {
+    if (isEquivalent(NewInst, I))
+      return true;
+    return false;
+  });
+  return Results.size();
+}
+
 Inst *souper::instJoin(Inst *I, Inst *EmptyInst, Inst *NewInst,
                        std::map<Inst *, Inst *> &InstCache,
                        InstContext &IC) {
+  if (DAGMatch(NewInst, I)) {
+    return nullptr;
+  }
+
   if (InstCache.count(I))
     return InstCache.at(I);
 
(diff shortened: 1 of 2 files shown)


regehr:
plz refresh, it looks like the diffs shown here are including the other pull req too

regehr:
I'd like you to do a quick experiment where you generate all guesses (for 3-4 instructions -- make it take some time), but don't call the solver at all, and measure the time before and after this patch, let's make sure it's not slowing things down too much

pranavk:
image

Here's the data. Not much difference for num-instrs<=3. Noticeable but still small difference when num-instrs=4.

regehr:
perfect. thanks. I think this amount of overhead of totally acceptable, since it allows us to avoid expensive solver calls.

regehr:
sorry, one more thing: how many fewer guesses do you get after this patch at n=4?

pranavk:
image

(5_2 means 5th example and num-instrs=2, etc.). All other RHS were only tested for num-instrs=2 as search space was quite huge because I am ignoring the cost of RHSes while doing this experiment.

This is the data. Here are the LHSes.
0:
```
%0:i32 = var
%1 = sub %0, 1
%2 = add %1, %1
%3 = shl %2, 1
infer %3
```

1:
```
%0:i32 = var
%1 = add %0, 1
%2 = and %1, 5
%3 = or %2, 42
infer %3
```

2:
```
%A:i8 = var
%B:i8 = var
%na = xor %A, -1
%op1 = xor %na, %B
%r = or %A, %op1
infer %r
```

3:
```
%0:i64 = var
%1:i64 = var
%2:i64 = var
%1b = xor %1, 23513542
%2b = xor %2, 234905723498
%3 = shl %0, %1b
%4 = lshr %3, %2b
infer %4
```

4:
```
%0:i32 = var
%1 = add %0, 1
%2 = and %1, 5
%3 = or %2, 42
%4 = sle %3, 1000
%5 = select %4, %1, %2
infer %5
```
5:
```
%0:i8 = var
%1:i16 = var
%2:i16 = sext %0
%3 = slt %2, %1
infer %3
```

lib/Inst/Inst.cpp
```diff
@@ -1174,16 +1174,68 @@ Inst *souper::getInstCopy(Inst *I, InstContext &IC,
   return Copy;
 }
 
+bool isTerminalInst(Inst *I) {
+  bool terminal;
```
regehr: when writing this kind of function, don't use a state variable, just directly return the correct thing

lib/Inst/Inst.cpp
```diff
@@ -1174,16 +1174,68 @@ Inst *souper::getInstCopy(Inst *I, InstContext &IC,
...
+  }
+
+  return true;
+}
+
+// Return true if NewInst is present in Cand.
+bool DAGMatch(Inst *NewInst, Inst *Cand) {
+  std::vector<Inst*> Results;
+  souper::findInsts(Cand, Results, [NewInst](Inst *I) {
+    if (isEquivalent(NewInst, I))
```
regehr: just "return isEquivalent(NewInst, I)" the conditional is only confusing

pranavk: yeah, i agree. silly me.

lib/Inst/Inst.cpp
pranavk: ok. will fix. I remember you have told me this thing before too but for some reason, I am naturally inclined to do it this way. I will change it.

pranavk:
I discussed it with @zhengyangl the other day and we were convinced that this will be missing some cases. Given that the benefit of this is not huge (and risks of missing cases are higher), let's just close this.

PR #628, opened 2019-10-25
Print for Inst*

pranavk:


regehr:
I don't understand adding a function that isn't used, what's the use case?

pranavk:
pretty helpful in debugging. Otherwise, you end up doing ReplacementContext RC; RC.printInst(bla bla). But feel free to not merge if you think it's not useful. I will create a template in my IDE to accomplish this.

regehr:
the code is fine, but if you are adding a function with no calls in our code base, you need to explain this in a comment or something

regehr:
ping

pranavk:
completely forgot about it. Done now.

regehr:
lgtm thanks

PR #633, opened 2019-11-04
bump to llvm9

zhengyangl:
This PR fixes the two failed assertions in separateBPC(). The cause of the assertion failure is because an unnecessary and wrong call to separateBPC() in isConcreteCandidateSat() in EnumerativeSynthesis.cpp.

This PR also fixes an wrong assertion in ConstantSynthesis.cpp. Previously I made a wrong assumption that RHS should never contain phi because we're not synthesizing phis. This is not correct because RHS may use phis harvested from LHS. I removed the assertion in this patch.

All the test cases passed after this patch in a debug build.

### DIFF
--- a/lib/Infer/EnumerativeSynthesis.cpp
+++ b/lib/Infer/EnumerativeSynthesis.cpp
@@ -599,16 +599,9 @@ std::error_code synthesizeWithAlive(SynthesisContext &SC, Inst *&RHS,
 
 std::error_code isConcreteCandidateSat(SynthesisContext &SC, Inst *RHSGuess, bool &IsSat) {
   std::error_code EC;
-  BlockPCs BPCsCopy;
-  std::vector<InstMapping> PCsCopy;
-  std::map<Inst *, Inst *> InstCache;
-  std::map<Block *, Block *> BlockCache;
-  separateBlockPCs(SC.BPCs, BPCsCopy, InstCache, BlockCache, SC.IC, {}, false);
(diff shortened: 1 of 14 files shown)


zhengyangl:
@googlebot I consent.

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -599,16 +599,9 @@ std::error_code synthesizeWithAlive(SynthesisContext &SC, Inst *&RHS,
 
 std::error_code isConcreteCandidateSat(SynthesisContext &SC, Inst *RHSGuess, bool &IsSat) {
   std::error_code EC;
-  BlockPCs BPCsCopy;
-  std::vector<InstMapping> PCsCopy;
-  std::map<Inst *, Inst *> InstCache;
-  std::map<Block *, Block *> BlockCache;
-  separateBlockPCs(SC.BPCs, BPCsCopy, InstCache, BlockCache, SC.IC, {}, false);
-  separatePCs(SC.PCs, PCsCopy, InstCache, BlockCache, SC.IC, {}, false);
-
```
zhengyangl: These two calls are redundant and cause errors.

regehr:
looks great, thanks!!
please add a comment to getSetSize explaining that we've copied the routine from LLVM 8 since LLVM 9 removed this API

manasij7479:
@googlebot I consent.

PR #645, opened 2019-11-12
Generate and verify concrete candidates in buckets(of size MaxV)

manasij7479:


manasij7479:
@zhengyangl Please review.
We might need some tweaks to the order in which 'components' are specified to match the global cost order.

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -751,70 +812,40 @@ void generateAndSortGuesses(SynthesisContext &SC,
...
+  uint64_t GuessCount = 0;
+
+  auto Generate = [&SC, &Guesses, &RHS, &EC, &GuessCount](Inst *Guess) {
+    GuessCount++;
+    Guesses.push_back(Guess);
+    if (Guesses.size() >= MaxV && !SkipSolver) {
+      sortGuesses(Guesses);
+      EC = verify(SC, RHS, Guesses);
+      Guesses.clear();
+      return RHS == nullptr; // Continue if no RHS
```
zhengyangl: Which RHS does this lambda capture?

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -751,70 +812,40 @@ void generateAndSortGuesses(SynthesisContext &SC,
...
-  getGuesses(Guesses, Cands, SC.LHS->Width,
-             LHSCost, SC.IC, nullptr, nullptr, TooExpensive, PruneCallback);
-  if (DebugLevel >= 1) {
-    DataflowPruning.printStats(llvm::errs());
-  }
+
+  std::vector<Inst *> Guesses;
+  uint64_t GuessCount = 0;
+
+  auto Generate = [&SC, &Guesses, &RHS, &EC, &GuessCount](Inst *Guess) {
```
zhengyangl: Instead of wrapping the function into lambda with complex capture and passing it around, why not just use a normal function, that would make code much more readable :)

lib/Infer/EnumerativeSynthesis.cpp
manasij7479: The result.

lib/Infer/EnumerativeSynthesis.cpp
manasij7479: That won't work because of the captures. One solution is making a class with operator () overloaded. Not sure if that's cleaner than this.

lib/Infer/EnumerativeSynthesis.cpp
zhengyangl: I see, RHS is changed in verify(). It would better to rely on EC for this early termination.

zhengyangl:
LGTM. Need to get the failed cases passing, though.

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -73,6 +73,9 @@ namespace {
   static cl::opt<unsigned> MaxNumInstructions("souper-enumerative-synthesis-num-instructions",
     cl::desc("Maximum number of instructions to synthesize (default=1)."),
     cl::init(1));
+  static cl::opt<unsigned> MaxV("souper-enumerative-synthesis-max-verification-load",
+    cl::desc("Maximum number of guesses verified at once (default=300)."),
+    cl::init(3000));
```
regehr: make these consistent (300 vs 3000)

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -242,7 +258,17 @@ void getGuesses(std::vector<Inst *> &Guesses,
   Inst *I2 = IC.getReservedInst();
   Comps.push_back(I2);
 
-  for (auto K : BinaryOperators) {
+  std::vector<Inst::Kind> BinOpsSorted = BinaryOperators;
+  if (Width == 1) {
+    std::stable_sort(BinOpsSorted.begin(), BinOpsSorted.end(), [](Inst::Kind A, Inst::Kind B) {
+      if (A == Inst::Kind::Eq || A == Inst::Kind::Ne) {
```
regehr: this shouldn't be an if/else, just return the boolean value you calculated

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -470,24 +498,36 @@ void getGuesses(std::vector<Inst *> &Guesses,
...
     }
 
     // if there exist empty slots, then call getGuesses() recursively
     // and fill the empty slots
     if (prune(JoinedGuess, CurrSlots)) {
-      for (auto S : CurrSlots)
-        getGuesses(Guesses, Inputs, S->Width,
-                   LHSCost, IC, JoinedGuess, S, TooExpensive, prune);
+      for (auto S : CurrSlots) {
+        if(!getGuesses(Inputs, S->Width,
```
regehr: space after if

lib/Infer/EnumerativeSynthesis.cpp
manasij7479: done

lib/Infer/EnumerativeSynthesis.cpp
```diff
@@ -193,7 +209,7 @@ void getGuesses(std::vector<Inst *> &Guesses,
 
   std::vector<Inst *> PartialGuesses;
 
-  std::vector<Inst *> Comps(Inputs.begin(), Inputs.end());
+  std::deque<Inst *> Comps(Inputs.begin(), Inputs.end());
```
regehr: dequeues are pretty slow, you're sure you want to do this?

regehr: if you're sorting anyway, why do you need to add guesses to the front?

manasij7479: No, can leave it as a vector for now.
This will have to change eventually though.

regehr:
LGTM

PR #659, opened 2019-11-16
Restrict input bits for analysis implementations in the presence of PC

manasij7479:


regehr:
I need more here. what's going on? why is it correct?

regehr:
add comments in the code explaining the strategy and rationale for correctness.

regehr:
remove commented-out lines of code unless they serve a specific purpose, in which case explain them

regehr:
ping

manasij7479:
Oh, forgot about this.
Will rebase, add comments for explanation and update.

manasij7479:
updated

lib/Infer/Pruning.cpp
```diff
@@ -119,6 +119,32 @@ std::pair<llvm::APInt, llvm::APInt> knownBitsNarrowing
   return {KnownNotZero, KnownNotOne};
 }
 
+std::unordered_map<Inst *, llvm::APInt>
+PruningManager::computeInputRestrictions() {
+  std::unordered_map<Inst *, std::pair<llvm::APInt, llvm::APInt>> Seen;
```
regehr: you don't need to change it here but overall you should avoid using a std::pair for same-typed things like this, and instead pick an implementation which assigns appropriate names to things. here two separate variables would work, or else a struct with members called Zero and One or similar

lib/Infer/Pruning.cpp
```diff
@@ -119,6 +119,32 @@ std::pair<llvm::APInt, llvm::APInt> knownBitsNarrowing
...
+  // ^ SeenZero, SeenOne
+
+  for (auto &&V : InputVars) {
+    Seen[V].first = llvm::APInt(V->Width, 0);
+    Seen[V].second = Seen[V].first;
```
regehr: initialize this to llvm::APInt(V->Width, 0)

lib/Infer/Pruning.cpp
```diff
@@ -119,6 +119,32 @@ std::pair<llvm::APInt, llvm::APInt> knownBitsNarrowing
...
+  std::unordered_map<Inst *, llvm::APInt> Result;
+  for (auto &&V : InputVars) {
+    Result[V] = ~(Seen[V].first & Seen[V].second);
+  }
+  return Result;
+}
+
 // TODO : Comment out debug stmts and conditions before benchmarking
 bool PruningManager::isInfeasible(souper::Inst *RHS,
                                  unsigned StatsLevel) {
```
regehr: indentation is off here

regehr:
I still have no idea why this does what you want it to do, you'll have to explain it more either in the code or in person

regehr:
@manasij7479 this is one of several things you need to get back to and finish up

manasij7479:
Addressed the comments, not sure what to do about the explanation.

regehr:
how about some examples which show this working, then?

regehr:
I don't like random, complicated code which is neither tested nor well-explained. I know you can do better than this, please make an effort here.

regehr:
it seems pretty obvious what to do here:
- add restricted bits to souper-interpret
- write test cases checking for sound processing of souper expressions with pcs

regehr:
I have no idea why you would be confident enough in this code's correctness to submit it for inclusion in souper without having done this

PR #663, opened 2019-11-17
tests for ternary op for RB

pranavk:


regehr:
what's status of this?

pranavk:
this got merged into master by one of Manasij's commit. However, this version avoids code duplicacy and merge ternary support into the same exhaustive functions. I think for now, we want to keep the ternary functions separate. So, let's close this PR.

PR #664, opened 2019-11-17
Make concrete interpreter optional for constant synthesis, Use the KLEE for G by S by default

zhengyangl:


regehr:
LGMT thanks!

PR #665, opened 2019-11-18
RB exhaustive tester fix

manasij7479:


regehr:
please supply a correct but conservative transfer function for each instruction, instead of commenting out the tests

regehr:
also please protect debug output behind a command-line flag instead of commenting out stuff

manasij7479:
updated

regehr:
LGTM

PR #669, opened 2019-11-18
maximally precise sle,ule,slt,ult

pranavk:


lib/Infer/AbstractInterpreter.cpp
```diff
@@ -936,10 +936,15 @@ namespace souper {
       case Inst::Sle:
       case Inst::Ult:
       case Inst::Slt:
-        if (RB0 == 0 && RB1 == 0)
-          Result = AllZeroes;
+      {
```
regehr: are we REALLY sure that the same transfer function is maximally precise for all of these operations? this seems kind of suspicious!! please triple check everything here
also please put the open curly brace on the previous line like this ```Slt: {```

regehr:
status?

pranavk:
abandoning this because we found a bug in the exhaustive tester after this PR was made. So this PR was based on an incorrect table generated from exhaustive tester.

PR #670, opened 2019-11-19
Cannot expect each LHS variable in RHS

pranavk:
This assert is incorrect. It assumes that each variable in LHS will be
used in the RHS which is not always true.


regehr:
PLZ review this @manasij7479

pranavk:
@manasij7479 we discussed this and you agreed that this is a bug. Can you approve here?

pranavk:
This assertion is hit when optimizing aaom.opt in the pruning-paper repo.

PR #674, opened 2019-11-26
freeze support

pranavk:
This PR adds support for synthesizing freeze instruction.


pranavk:
I will start working on it once Freeze support in LLVM settles down. There are still some outstanding patches there.

pranavk:
The CI will always fails because docker contains previous LLVM 9 version. Everything seems to be okay except one test `Solver/extract.ll` which is currently failing after I updated to LLVM 10 version in this PR. I am taking a look at it.

In the meantime, you can go ahead with the review.

pranavk:
@zhengyangl I spent some time investigating why Solver/extract.ll is failing with LLVM 10 and it turns out this is due to LLVM 10 harvesting `range` metadata for one of the variable.

The test is:
```
%struct = type { i32, [4 x i8] }

define i32 @test1() {
  %A = extractvalue %struct { i32 2, [4 x i8] c"foo\00" }, 0
    ret i32 %A
    ; CHECK-LABEL: @test1(
    ; CHECK: ret i32 2
}
```
The difference in the code starts at this point (https://github.com/google/souper/blob/master/lib/Tool/CandidateMapUtils.cpp#L213). Cand.Mapping.LHS at this point for LLVM 9 was : 
`%0:i32 = var `
but with LLVM 10, it is:
`%0:i32 = var (range=[2,3))`. The new metadata is making L213 give us a non-null RHS while we were getting null RHS before. This is making the test to fail. The code goes deep into constant synthesis and diverges even further. You may have good insight on this one. Can you see where things may have broken or suggest where should I look?

regehr:
good job tracking this down.
probably best to just replace this with a different file that tests the same thing, but doesn't allow LLVM to do the constantrange inference

pranavk:
I am not sure this test is even necessary at all. Functionality of `extractvalue` is tested at several other places in the test suite as well.

regehr:
fine to remove if zhengyang agrees this is OK

zhengyangl:
Weird, I need to have a look before accepting the removal.

zhengyangl:
TL DR: the test case extract.ll should be 
```
; REQUIRES: solver

; RUN: %llvm-as -o %t %s
; RUN: %souper %solver -check %t

target datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128"

%struct = type { i32, [4 x i8] }

define i32 @test1() {
  %A = extractvalue %struct { i32 2, [4 x i8] c"foo\00" }, 0, !expected !0
  ret i32 %A
}

!0 = !{i32 2}
```

zhengyangl:
The original test case is problematic. It is designed to test if Souper can transform the extractvalue into a constant. It tries to use the FileCheck format for the text comparison, but on the header of the test case, there is no call to FileCheck in the `RUN` line, so it never works the way it is designed to do.

zhengyangl:
The `souper` command with `-check` checks if the the synthesized const is equal to the `expected` metadata, it only fails the test case when they are mismatch or the `expected` metadata for the instruction is missing. If there is no synthesized const for the LHS, the test case passes. Before bumping to new LLVM, the `souper -check` succeed because Souper cannot synthesized the constant 2 from the extractvalue (the value is harvested as var), so the test case passes. But after bumping, LLVM assigns range DFA fact to the extractvalue, and Souper is now able to synthesize the constant with the range DFA fact. Unfortunately, there is no `expected` metadata assigned, thus the test cases fails.

zhengyangl:
The fix is easy: add expected metadata and remove the unnecessary CHECK lines.

pranavk:
Great. Thank you for the investigation. I will fix this.

pranavk:
@regehr This PR is now ready to review. I have tested both release and debug builds and all tests passes.

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -18,6 +18,7 @@
 #include "souper/Extractor/ExprBuilder.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Analysis/LoopInfo.h"
+#include "llvm/Support/CommandLine.h"
```
regehr: what's this for?

lib/Extractor/KLEEBuilder.cpp
```diff
@@ -141,6 +142,8 @@ class KLEEBuilder : public ExprBuilder {
       }
       return E;
     }
+    case Inst::Freeze:
+      return get(Ops[0]);
```
regehr: is this just ignoring freeze in kleebuilder?

lib/Infer/Interpreter.cpp
```diff
@@ -426,6 +426,13 @@ namespace souper {
       else
         return {ARG0.getHiBits(1).trunc(1)};
     }
+    case Inst::Freeze: {
+      if (Args[0].K == EvalValue::ValueKind::Undef || Args[0].K == EvalValue::ValueKind::Poison) {
+        return {llvm::APInt(Args[0].BitWidth,
+                std::rand() % llvm::APInt::getAllOnesValue(Args[0].BitWidth).getLimitedValue())};
```
regehr: is this the right thing to do? it seems like this will make testing non-deterministic, etc.

test/Codegen/argument-order-XFAIL.opt
```diff
@@ -10,7 +10,7 @@ result %4
 ; CHECK: ; ModuleID = 'souper.ll'
 ; CHECK-NEXT: source_filename = "souper.ll"
 
-; CHECK: define i8 @fun(i16, i4) {
+; CHECK: define i8 @fun(i16 %0, i4 %1) {
```
regehr: I wasn't expecting you to need to do this, can you explain?

regehr:
ok thanks. looks good overall, left a few comments.
can you please add a few more test cases for freeze though?

test/Codegen/argument-order-XFAIL.opt
regehr: ah nevermind, I can't believe they flipped this behavior again

lib/Extractor/KLEEBuilder.cpp
pranavk: The file uses cl:: namespace without including this header file. It worked with LLVM 9 probably because CommandLine.h was indirectly added by some other header file we are including here but this doesn't work with LLVM 10. Better to be explicit, as they say.

lib/Infer/Interpreter.cpp
pranavk: Yes, this will make testing hard to do. But I don't think it's right idea either to make `freeze` always come up with same result?

lib/Extractor/KLEEBuilder.cpp
pranavk: Sort of. I cannot put a explicit assert here to crash (as you suggested) as it results in several test failures because freeze is also synthesized now. So, it does need some sort of klee support for those RHSes to interpret successfully.

lib/Infer/Interpreter.cpp
regehr: Well, this needs to be thought about carefully, since it could have an impact on soundness. Talk to @manasij7479 about it please.

manasij7479: Not really sure what to do here, other than: https://xkcd.com/221/  
Maybe a 'freeze fuzzing' mode which runs the tests with freeze producing arbitrary results?

regehr: what all are we using the interpreter for? the solution we choose has to be appropriate for that.

manasij7479: We interpret both sides to see if they can produce a different result.
If all freezes produce the same result, this technique will not rule out some invalid transformations, which will then be sent off to the solver.
Similarly for pruning, we interpret the LHS and abstract-interpret the RHS.
It seems to me that producing the same value will (rarely, if at all) inhibit pruning, not introduce unsoundness.

pranavk:
Okay, finally was able to synthesize freeze. I added the test case above. There were some minor modifications I made to enumerative algorithm as well to prevent it from generating freezes of freezes.

include/souper/Infer/Interpreter.h
```diff
@@ -48,14 +51,20 @@ struct EvalValue {
     return Value;
   }
 
-  static EvalValue poison() {
+  static EvalValue poison(unsigned Width) {
+    if (Width == 0)
```
regehr: since there is not supposed to be a situation where this happens, please make it an assertion, not a regular error check (error checks are for things that can happen)

include/souper/Infer/Interpreter.h
```diff
@@ -48,14 +51,20 @@ struct EvalValue {
...
+      llvm::report_fatal_error("poison of zero width!");
     EvalValue Result;
     Result.K = ValueKind::Poison;
+    Result.BitWidth = Width;
     return Result;
   }
-  static EvalValue undef() {
+  static EvalValue undef(unsigned Width) {
+    if (Width == 0)
```
regehr: same

lib/Extractor/KLEEBuilder.cpp
regehr: we don't want silent unsound behavior, so something besides this needs to be done here

regehr: you could add a flag to not synthesize freeze, and use it in the tests that crash

test/Infer/Interpreter/freeze.opt
```diff
@@ -0,0 +1,12 @@
+; RUN: %souper-interpret %solver -input-values=%0=1 %s > %t 2>&1
+; RUN: %FileCheck -check-prefix=TEST1 %s < %t
+; TEST1: Value: 126
+
+; RUN: %souper-interpret %solver -input-values=%0=5 %s > %t 2>&1
+; RUN: %FileCheck -check-prefix=TEST2 %s < %t
+; TEST2-NOT: Value: Poison
```
regehr: I do not understand this test case: a freeze instruction can never generate poison

regehr:
looks good except for noted items, and also please switch to the LLVM 10 branch

include/souper/Infer/Interpreter.h
pranavk: this is to make sure that even if we do a release build without assertions, it doesn't go unnoticed. I vaguely remember some time back when we changed the asserts to this "fatal error" to make sure we don't skip asserts under any circumstances.

test/Infer/Interpreter/freeze.opt
pranavk: This is what this test is asserting. Note the "-NOT" after TEST2. "-NOT" means that it never generates poison.

include/souper/Infer/Interpreter.h
regehr: souper has a lot of assertions in it, we never got rid of them

regehr: if you're simply guarding against internal inconsistency, assertion is usualyl the way to go, and it is what you should be using here

test/Infer/Interpreter/freeze.opt
regehr: ah, got it

include/souper/Infer/Interpreter.h
pranavk: ok. i will change it to assert.

regehr:
thanks!

PR #676, opened 2019-12-09
Codegen

zhengyangl:
Take over from Roman's nice patch https://github.com/google/souper/pull/504 :) 

Some test cases are not passing. I'll reply here once the PR is good for review.


zhengyangl:
@googlebot I consent.

LebedevRI:
@googlebot does not work on PR's with commits by more than one author.

zhengyangl:
This patch is ready for review. @regehr

regehr:
what happens to the old souper2llvm in this patch? shouldn't you rename it as old-souper2llvm or something? otherwise we end up with two programs of the same name and this is likely very confusing

zhengyangl:
> what happens to the old souper2llvm in this patch? shouldn't you rename it as old-souper2llvm or something? otherwise we end up with two programs of the same name and this is likely very confusing

@regehr fixed :)

CMakeLists.txt
```diff
@@ -421,7 +437,7 @@ configure_file(${CMAKE_SOURCE_DIR}/utils/cache_dump.in ${CMAKE_BINARY_DIR}/cache
 configure_file(${CMAKE_SOURCE_DIR}/utils/cache_dfa.in ${CMAKE_BINARY_DIR}/cache_dfa @ONLY)
 configure_file(${CMAKE_SOURCE_DIR}/utils/cache_import.in ${CMAKE_BINARY_DIR}/cache_import @ONLY)
 configure_file(${CMAKE_SOURCE_DIR}/utils/cache_infer.in ${CMAKE_BINARY_DIR}/cache_infer @ONLY)
-configure_file(${CMAKE_SOURCE_DIR}/utils/souper2llvm.in ${CMAKE_BINARY_DIR}/souper2llvm @ONLY)
+configure_file(${CMAKE_SOURCE_DIR}/utils/souper2llvm.in ${CMAKE_BINARY_DIR}/py_souper2llvm @ONLY)
```
regehr: great, but please also rename the actual file here, to py_souper2llvm.in

regehr:
thanks!

PR #677, opened 2019-12-10
remove nop synthesis

zhengyangl:


regehr:
thanks!!! I've wanted that code to be gone for quite a while

PR #680, opened 2019-12-24
fix wrapped range check to handle special case in which upper bound is 0

jubitaneja:
LLVM-9.0 redefined isWrappedSet() as: `Lower.ugt(Upper) && !Upper.isNullValue()`
to ignore the case in which upper bound is zero, and added a new API
isUpperWrapped() as: `Lower.ugt(Upper)`.

However, in LLVM-8.0, isWrappedSet() checked `Lower.ugt(Upper)` to cover all
cases of wrapped range.

For reference, This change was made in the LLVM's commit id: 58c0bdde21ea


jubitaneja:
@zhengyangl Adding you as reviewer too.

jubitaneja:
ok, I have updated this PR and it's ready for review. It would be good to have a test case for the change that I made in the Abstract Interpretation part. Maybe @manasij7479 or someone else can take a look as well? If the test case for this part is not important, we can ignore it. There is a test case for another part though.

regehr:
LGTM, thanks!

PR #681, opened 2019-12-30
remove deprecated Z3_EXECUTABLE cmake option

zhengyangl:
The Z3_EXECUTABLE cmake option for building LLVM is deprecated since https://reviews.llvm.org/D54978 .


regehr:
thanks!

PR #682, opened 2019-12-30
Use orderedOps in codegen to eliminate the ambiguous generation of commutativity instructions

zhengyangl:
Several test cases fail on OS X due to the ambiguous code generation of commutativity operators. This patch fixes the issue by using Inst::orderedOps instead of the raw operand vector Inst::Ops in LLVM codegen. 

Note that now LLVM code generation of commutativity operators are shaped by the less than operator of souper::Inst. The ordering of operands does not always follow the way LLVM InstCombine canonicalization prefers. For example, the most common canonicalization rule is that if there is a constant in the operand list and the operator is commutative, then place the constant on the RHS. However, the current Souper code will always put the constant on the LHS. This "constant on LHS" ordering is already spread all over the Souper test suite. I can give a patch for fixing that if @regehr agrees that constant should be placed on the RHS.

### DIFF
--- a/lib/Codegen/Codegen.cpp
+++ b/lib/Codegen/Codegen.cpp
@@ -49,6 +49,7 @@ llvm::Type *Codegen::GetInstReturnType(llvm::LLVMContext &Context, Inst *I) {
 }
 
 llvm::Value *Codegen::getValue(Inst *I) {
+  const std::vector<Inst *> &Ops = I->orderedOps();
   if (I->K == Inst::UntypedConst) {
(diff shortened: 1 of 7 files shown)


zhengyangl:
ref: https://github.com/llvm/llvm-project/blob/948e745270de615cb83bc9c68cdfa4da62925d0f/llvm/lib/Transforms/InstCombine/InstructionCombining.cpp#L23

regehr:
LGTM thanks!

PR #683, opened 2020-01-01
remove the unnecessary libc build in build_deps.sh

zhengyangl:
The script build_deps.sh failed when building llvm/libc on freebsd. I found out that libc is not used in Souper, so I just removed it from build_deps.sh. Test suite ran good on osx, ubuntu and freebsd on both debug and release for this patch.


regehr:
thanks!

PR #684, opened 2020-01-03
Add support for optimal DFA on LLVM bitcode

zhengyangl:
- backward analysis (Demanded bits) is not supported

### DIFF
--- a/lib/Tool/CandidateMapUtils.cpp
+++ b/lib/Tool/CandidateMapUtils.cpp
@@ -148,6 +148,7 @@ bool SolveCandidateMap(llvm::raw_ostream &OS, CandidateMap &M,
           if (std::error_code EC = S->nonNegative(Cand.BPCs, Cand.PCs, Cand.Mapping.LHS,
                                                   NonNegative, IC)) {
             llvm::errs() << "Error: " << EC.message() << '\n';
+            return false;
           } else {
             OS << "; nonNegative from souper: "
                << convertToStr(NonNegative) << "\n";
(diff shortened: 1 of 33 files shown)


lib/Tool/CandidateMapUtils.cpp
```diff
@@ -43,6 +72,18 @@ void souper::AddModuleToCandidateMap(InstContext &IC, ExprBuilderContext &EBC,
   }
 }
 
+static bool isInferDFA() {
+  return InferNeg || InferNonNeg || InferKnownBits || InferPowerTwo ||
+         InferNonZero || InferSignBits || InferRange;
+}
+
+static std::string convertToStr(bool Fact) {
```
regehr: there are already two functions in Souper that do this, do we really need a third?

tools/souper.cpp
```diff
@@ -44,24 +44,6 @@ OutputFilename("o", cl::desc("Override output filename"),
 static cl::opt<bool> StaticProfile("souper-static-profile", cl::init(false),
     cl::desc("Static profiling of Souper optimizations (default=false)"));
 
-static cl::opt<bool> PrintNeg("print-neg", cl::init(false),
```
regehr: hmm, what were these options doing in the code? @jubitaneja ?

zhengyangl: These are dead codes

lib/Tool/CandidateMapUtils.cpp
zhengyangl: Oops I’ll get it fixed

tools/souper.cpp
regehr: yeah I was wondering how they got in there in the first place, so we can do a better job avoiding nasty stuff like this in the future

regehr:
thanks-- is this ready to go?

zhengyangl:
yes!

regehr:
LGTM thanks!
great job with all the test cases!!!

PR #691, opened 2020-01-23
Do not attempt pruning if abstract interpreters provably return TOP

manasij7479:


manasij7479:
This produces a ~20% speedup when pruning is enabled.

regehr:
you'll have to explain this code more.
for example, in that big switch statement, why is "add" in one category and "addnsw" and "or" in a different one?

regehr:
LGTM

PR #696, opened 2020-01-27
constrain constant synthesis so that single-instruction nops are not …

regehr:
…produced


zhengyangl:
LGTM, but I am not sure if this is the best way to organize the code to facilitate dataflow based pruning. Maybe @manasij7479 can have a look?

regehr:
right, it's not clear that Manasij can reuse this code directly, but if he can then he will need to speak up

manasij7479:
I'll add further constrains derived from pruning, there is no need to future proof this right now.
LGTM

PR #699, opened 2020-01-30
Only fill one hole per each getGuesses call

zhengyangl:
This patch fixes a bug when trying to enumerate large RHSs.


regehr:
thanks!

PR #711, opened 2020-02-12
adjust max-cand commandline options to get the test cases pass

zhengyangl:


regehr:
thanks!

but I don't understand why reducing MaxLHSCands did not cause any local test failures for me, do you have any idea?

zhengyangl:
These test cases are marked long-duration-synthesis and it’s not enabled by default. TravisCI has this enabled.

regehr:
ah, thanks. I have -DTEST_SYNTHESIS=ON enabled, but this is another one, I had forgotten about that

PR #716, opened 2020-02-21
Pruning with partially restricted bits and abstract input specialization

manasij7479:
Makes pruning 30% faster and prunes a few percent more candidates.
Subsumes some of the slower techniques.


manasij7479:
WIP

regehr:
is this ready?
if not, you should mark it as a draft pull request

manasij7479:
This is ready for review.

regehr:
test cases are still coming, I assume?

manasij7479:
This technique was not effective.
Reason: restricting inputs can only make output more restricted. Pruning is done of the basis of 'unrestrictedness'.  
I was wrong about the numbers because of a bug.
Closing.

PR #717, opened 2020-02-21
Build third_party parallelized with all possible cores when ninja is not found

zhengyangl:


regehr:
please support mac OS also ```sysctl -n hw.ncpu```

regehr:
thanks!

PR #727, opened 2020-03-05
Pruning refactoring and misc bugfixes

manasij7479:


regehr:
if you're getting rid of the test that uses -souper-suppress-foldable-operations=false, I want you to just eliminate that flag, as part of this patch

manasij7479:
Done

regehr:
sorry, please just nuke the AvoidNops parameter, unless there's some use case for it? I can't think of one

manasij7479:
souper-check -infer-const uses (AvoidNops = false).
I don't think we should completely get rid of our ability to synthesize a zero when constant synthesis is invoked separately from enumerative synthesis.

regehr:
you'll need to explain why you think AvoidNops == false makes sense. be concrete.

regehr:
the thing you said "rid of our ability to synthesize a zero when constant synthesis is invoked separately from enumerative synthesis" isn't what happens.

regehr:
so either do what I asked or else provide a better justification for not doing what I asked

manasij7479:
Here is a concrete use case which we can not abandon: synthesizing ranges with zero as a boundary.
This test case fails, for example:
Dataflow/precision-range-4.ll:24:15: error: CHECK-NEXT: expected string not found in input
; CHECK-NEXT: ; range from souper: [0,53)
              ^
<stdin>:31:1: note: scanning from here
; range from souper: [-2,53)

regehr:
this is an actual souper output when you disable that flag?

manasij7479:
Yes, there are 17 such test cases which need a range with a zero on one side.

manasij7479:
Failing Tests (17):
    Souper :: Dataflow/precision-multiple.ll
    Souper :: Dataflow/precision-multiple.opt
    Souper :: Dataflow/precision-range-1.ll
    Souper :: Dataflow/precision-range-1.opt
    Souper :: Dataflow/precision-range-2.ll
    Souper :: Dataflow/precision-range-2.opt
    Souper :: Dataflow/precision-range-3.ll
    Souper :: Dataflow/precision-range-3.opt
    Souper :: Dataflow/precision-range-4.ll
    Souper :: Dataflow/precision-range-4.opt
    Souper :: Dataflow/precision-range-5.ll
    Souper :: Dataflow/precision-range-5.opt
    Souper :: Dataflow/precision-range-7.opt
    Souper :: Dataflow/precision-range-8.ll
    Souper :: Dataflow/precision-range-8.opt
    Souper :: Dataflow/precision-range-9.ll
    Souper :: Dataflow/precision-range-9.opt

regehr:
LGTM

PR #732, opened 2020-03-22
Fix demanded bits pruning condition

manasij7479:


include/souper/Infer/Pruning.h
```diff
@@ -60,6 +60,7 @@ class PruningManager {
   HoleAnalysis HA;
   llvm::KnownBits LHSKnownBitsNoSpec;
   InputVarInfo LHSMustDemandedBits;
+  bool EnableDemandedBitsPruning = false;
```
regehr: does this really want to be a state variable, as opposed to a parameter that gets passed around?

manasij7479: I think state variable is appropriate, because it is not intended for an user.
The pruning initialization function sets it to true if the 'must demanded' analysis of the LHS returns an useful result.

PR #742, opened 2020-03-31
dedup guesses in enuemrative synthesis

zhengyangl:


regehr:
please make it print the number of guesses before and after deduplication (when debug level == 3 or something like that) so we can see the extent of the duplication

zhengyangl:
ready for review for this one.

regehr:
thanks!

PR #754, opened 2020-04-29
Link Z3 dynamically

pranavk:
Z3 is used in two places in souper's dependencies. 1) LLVM and 2)
libalive2.so. When both of these dependencies use Z3 statically, there
are two copies of z3 and it creates horrible horrible problems in
AliveDriver in souper. Let's dynamically link Z3 so that we use only one
copy of Z3 and avoid these nasty problems.


pranavk:
CI failure is expected because it needs changes to docker container. On my machine, all tests pass.

PR #756, opened 2020-05-18
Generate weakest known-bits precondition disjunction.

manasij7479:
Limited to one variable functions now.

### DIFF
--- a/lib/Extractor/Solver.cpp
+++ b/lib/Extractor/Solver.cpp
@@ -262,6 +262,154 @@ public:
     return std::error_code();
   }
 
+  std::error_code abstractPrecondition(const BlockPCs &BPCs,
+                  const std::vector<InstMapping> &PCs,
+                  InstMapping &Mapping, InstContext &IC,
+                  bool &FoundWeakest) override {
+    bool Valid;
+    PrintReplacement(llvm::outs(), BPCs, PCs, Mapping);
+
+    std::vector<std::pair<Inst *, APInt>> Models;
+    if (std::error_code EC = isValid(IC, BPCs, PCs, Mapping, Valid, &Models)) {
+      llvm::errs() << EC.message() << '\n';
+    }
+
+    if (Valid) {
+      llvm::outs() << "Already valid.\n";
+      return {};
+    }
+
+    std::vector<Inst *> Vars;
+    findVars(Mapping.LHS, Vars);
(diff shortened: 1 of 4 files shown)


regehr:
is this a draft, or something we should merge?

lib/Extractor/Solver.cpp
```diff
@@ -262,6 +262,154 @@ class BaseSolver : public Solver {
...
+          llvm::outs() << "Transformation is not valid for any input.\n";
+          return EC;
+        } else {
+          FoundWeakest = true;
+          llvm::outs() << "Exhausted search space.\n";
+          break;
+        }
+      }
+
+      for (unsigned I=0; I< W; I++) {
```
regehr, line 348: avoid sloppy whitespace like I< W

test/Infer/ap/kb.opt
```diff
@@ -0,0 +1,48 @@
...
+%0:i8 = var
+%1 = add %0, 1
+%2 = or %1, 7
+infer %2
+%3 = add %0, 2
+%4 = or %3, 7
+result %4
+;CHECK: xxxxxxx1
+;CHECK: xxxxx0xx
+;CHECK: xxxxxx0x
```
regehr, line 37: I don't understand the multiple CHECK lines, does your code print multiple solutions if these exist?

test/Infer/ap/kb.opt
```diff
@@ -0,0 +1,48 @@
...
+%0:i8 = var (knownBits=xxxxxxx0)
+%1:i8 = add 1:i8, %0
+%2:i8 = or 7:i8, %1
+infer %2
+%3:i8 = add 2:i8, %0
+%4:i8 = or 7:i8, %3
+result %4
+;CHECK: xxxxx0xx
+;CHECK: xxxxxx0x
+
```
regehr, line 48: how about add one test case where the result is all unknown bits?

test/Infer/ap/kb.opt
manasij7479: Yes, it prints all the solutions.
So the weakest precondition is the disjunction of the results.

test/Infer/ap/kb.opt
manasij7479: All unknown exits early saying the optimization is already valid.
I'll add a test checking for that.

PR #761, opened 2020-05-27
Add constant guess to enumerative synthesis

manasij7479:
-souper-enumerative-synthesis-num-insts=0 can now synthesize a constant along with the nop guesses.


regehr:
this PR needs a bit more.
your first option is to add a few test cases where you explicitly disable the existing iN synthesis code
your section option is to remove the existing iN synthesis code from Solver.cpp so that our existing constant synthesis tests all go through this code here
either is fine

regehr:
nevermind, I'll take over this code

PR #768, opened 2020-05-31
remove redundant LHS/BPCs/PCs duplication

zhengyangl:
The use of the duplicated LHS/BPCs/PCs can be just replaced by the use of original LHS/BPCs/PCs in constant synthesis. There is no point for duplicating them and duplicating itself is triggering the issue in blockpc-invalid1.opt. This patch removes the duplication calls.

### DIFF
--- a/lib/Infer/ConstantSynthesis.cpp
+++ b/lib/Infer/ConstantSynthesis.cpp
@@ -407,9 +407,9 @@ ConstantSynthesis::synthesize(SMTLIBSolver *SMTSolver,
 
     std::map<Inst *, Inst *> InstCache;
     std::map<Block *, Block *> BlockCache;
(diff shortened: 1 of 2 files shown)


lib/Infer/ConstantSynthesis.cpp
```diff
@@ -421,16 +421,11 @@ ConstantSynthesis::synthesize(SMTLIBSolver *SMTSolver,
       }
     }
 
-    BlockPCs BPCsCopy;
-    std::vector<InstMapping> PCsCopy;
-    separateBlockPCs(BPCs, BPCsCopy, InstCache, BlockCache, IC, &ConstMap, false);
```
zhengyangl, line 426: After this patch both functions (separateBlockPCs and separatePCs) are no longer be used. Their implementation are perserved because there may be some potential G by S improvement that requires these routines in the future.

PR #784, opened 2020-07-24
Fix comment in test

fitzgen:
Found this while making sure that my parser could handle every `.opt` file under `souper/test/**`.

@regehr, please take a look, thanks!


regehr:
LGTM, thanks Nick!

PR #785, opened 2020-08-25
Translate phi as a constrained variable

manasij7479:


lib/Infer/AliveDriver.cpp
```diff
@@ -625,10 +634,31 @@ bool souper::AliveDriver::translateAndCache(const souper::Inst *I,
...
+                       "%" + std::to_string(InstNumbers++));
+
+        auto PhiConstraint = Builder.iCmp(getType(1),
+             "%" + std::to_string(InstNumbers++),
+             IR::ICmp::EQ, FreeVar, ExprCache[I->Ops[0]]);
+
+        for (int i = 1; i < I->Ops.size(); ++i) {
+          auto Current = Builder.iCmp(getType(1),
+             "%" + std::to_string(InstNumbers++),
+             IR::ICmp::EQ, FreeVar, ExprCache[I->Ops[0]]);
```
zhengyangl, line 647: The last argument should be ExprCache[I->Ops[i]]

lib/Infer/AliveDriver.cpp
```diff
@@ -625,10 +634,31 @@ bool souper::AliveDriver::translateAndCache(const souper::Inst *I,
...
+             "%" + std::to_string(InstNumbers++),
+             PhiConstraint, Current, IR::BinOp::Or);
+        }
+
+        //TODO: Add blockpc constraints
+
+        Builder.assume(PhiConstraint);
+
+        ExprCache[I] = FreeVar; // No longer a free var
+        return true;
```
zhengyangl, line 658: redundant return here :)

lib/Infer/AliveDriver.cpp
zhengyangl: Are you sure that the index 0 of I->Ops is the value that comes from predecessor 0 rather than the block id? I forgot, let me check.

manasij7479: Block id is a separate member in Inst.
Fixed Ops[i]

lib/Infer/AliveDriver.cpp
manasij7479: fixed

PR #786, opened 2020-09-16
Decompose PHI nodes

manasij7479:
ready for review

### DIFF
--- a/include/souper/Infer/AliveDriver.h
+++ b/include/souper/Infer/AliveDriver.h
@@ -29,7 +29,8 @@ namespace souper {
 class AliveDriver {
   typedef std::unordered_map<const Inst *, IR::Value *> Cache;
 public:
-  AliveDriver(Inst *LHS_, Inst *PreCondition_, InstContext &IC_);
+  AliveDriver(Inst *LHS_, Inst *PreCondition_, InstContext &IC_,
+               std::vector<Inst *> ExtraInputs = {});
 
   std::map<Inst *, llvm::APInt> synthesizeConstants(souper::Inst *RHS);
   std::map<Inst *, llvm::APInt> synthesizeConstantsWithCegis(souper::Inst *RHS, InstContext &IC);
--- a/lib/Infer/AbstractInterpreter.cpp
+++ b/lib/Infer/AbstractInterpreter.cpp
@@ -1114,6 +1114,10 @@ namespace souper {
         auto RB1 = RB.findRestrictedBits(I->Ops[1]);
         // Take bivalent bits of opposite operand to independent variables
         for (auto V : IVARS) {
+          if (V->Width != I->Ops[0]->Width ||
+              V->Width != I->Ops[1]->Width) {
+            continue;
+          }
           if (A.find(V) == A.end())
             A[V] = APInt::getNullValue(V->Width);
           if (B.find(V) == B.end())
--- a/lib/Infer/AliveDriver.cpp
+++ b/lib/Infer/AliveDriver.cpp
@@ -23,6 +23,7 @@
 #include <set>
 #include <sstream>
 #include <string_view>
+#include <unordered_set>
 #include <z3.h>
 
 extern unsigned DebugLevel;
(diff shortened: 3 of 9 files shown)


include/souper/Infer/AliveDriver.h
```diff
@@ -29,7 +29,7 @@ namespace souper {
 class AliveDriver {
   typedef std::unordered_map<const Inst *, IR::Value *> Cache;
 public:
-  AliveDriver(Inst *LHS_, Inst *PreCondition_, InstContext &IC_);
+    AliveDriver(Inst *LHS_, Inst *PreCondition_, InstContext &IC_, std::vector<Inst *> ExtraInputs = {});
```
regehr, line 32: is the whitespace change intentional?

lib/Infer/AliveDriver.cpp
```diff
@@ -342,18 +342,26 @@ synthesizeConstantUsingSolver(tools::Transform &t,
...
+souper::AliveDriver::AliveDriver(Inst *LHS_, Inst *PreCondition_, InstContext &IC_, std::vector<Inst *> ExtraInputs)
     : LHS(LHS_), PreCondition(PreCondition_), IC(IC_) {
   IsLHS = true;
   InstNumbers = 101;
   //FIXME: Magic number. 101 is chosen arbitrarily.
   //This should go away once non-input variable names are not discarded
 
+  // If alive ever supports src and tgt with different number of arguments in future, this can go away.
+  for (auto Extra : ExtraInputs) {
+    if(!translateAndCache(Extra, LHSF, LExprCache)) {
```
regehr, line 354: put a space after the "if"

what is the consequence of this message? what is the souper user intended to learn from it?

lib/Infer/AliveDriver.cpp
```diff
@@ -783,7 +791,66 @@ IR::Type &souper::AliveDriver::getType(int n) {
   }
   return *TypeCache[n];
 }
+namespace souper {
+struct Problem {
```
regehr, line 795: problem is a very vague name for a class, please try to give it something more descriptive

also, add a comment somewhere around here documenting the solution that is being implemented

lib/Infer/AliveDriver.cpp
manasij7479: It is a hack to make sure that src and tgt have the same number of arguments.
We want to show that src(x) = x, and tgt(x, y) = x are same in this case.
This hack adds the extra argument to src.  
The comment is to help diagnose potential issues because of adding magic arguments.

include/souper/Infer/AliveDriver.h
manasij7479: No, new IDE version seems to have reset my settings and wants to force 4-spaces. :(

lib/Infer/AliveDriver.cpp
regehr: but I am asking what is the consequence of the message appearing. 
it sounds like you are saying this would indicate a fatal flaw in souper.
if this is correct, you need to change this to either an assertion or else an llvm::report_fatal_error()

regehr:
@zhengyangl can you take a look at this?

@manasij7479 please add test cases, or reuse existing ones by adding alive double checking to them

lib/Infer/AbstractInterpreter.cpp
```diff
@@ -1114,6 +1114,10 @@ namespace souper {
         auto RB1 = RB.findRestrictedBits(I->Ops[1]);
         // Take bivalent bits of opposite operand to independent variables
         for (auto V : IVARS) {
+          if (V->Width != I->Ops[0]->Width ||
+              V->Width != I->Ops[1]->Width) {
```
regehr, line 1118: why is this happening?

lib/Infer/AliveDriver.cpp
```diff
@@ -351,18 +352,27 @@ synthesizeConstantUsingSolver(tools::Transform &t,
...
+                                 std::vector<Inst *> ExtraInputs)
     : LHS(LHS_), PreCondition(PreCondition_), IC(IC_) {
   IsLHS = true;
   InstNumbers = 101;
   //FIXME: Magic number. 101 is chosen arbitrarily.
   //This should go away once non-input variable names are not discarded
 
+  // If alive ever supports src and tgt with different number of arguments in future, this can go away.
+  for (auto Extra : ExtraInputs) {
+    if(!translateAndCache(Extra, LHSF, LExprCache)) {
```
regehr, line 365: white space after if

lib/Infer/AliveDriver.cpp
```diff
@@ -814,7 +827,114 @@ IR::Type &souper::AliveDriver::getType(int n) {
...
+
+    // Recursively call ReplacePhi, because Result might have Phi`s
+    return Result.ReplacePhi(IC, Change);
+  }
+  bool operator == (const RefinementProblem &P) const {
+    return LHS == P.LHS && RHS == P.RHS && Pre == P.Pre;
+  }
+  struct Hash
+  {
+    std::size_t operator()(const RefinementProblem& P) const
```
regehr, line 884: & goes next to P, please try to be consistent about little things like this

lib/Infer/AbstractInterpreter.cpp
manasij7479: I didn't originally account for inputs with differing bitwidth.

regehr: I just want to make sure you're not hiding initialization bugs

manasij7479: Not here, there are a couple of similar bugs which I am fixing now which are problematic.

regehr:
LGTM

PR #787, opened 2020-09-23
Encode DemandedBits

manasij7479:


regehr:
of course this needs test cases. probably you can reuse our existing demanded bits test cases by adding a new run line to them and turning on alive double checking in the new run lines, right?

manasij7479:
Turned on alive2 flags for the demanded bits tests, except for the 2 test cases which involve phi.

lib/Infer/AliveDriver.cpp
```diff
@@ -777,6 +791,26 @@ souper::AliveDriver::translateDataflowFacts(const souper::Inst* I,
...
 
+bool
+souper::AliveDriver::translateDemandedBits(const souper::Inst* I,
+                     IR::Function& F,
+                     souper::AliveDriver::Cache& ExprCache) {
+  FunctionBuilder Builder(F);
+
+  auto DemandedBits = IsLHS ? I->DemandedBits : LHS->DemandedBits;
+
+  if (DemandedBits.isAllOnesValue() || DemandedBits.getBitWidth() != I-> Width) {
```
regehr, line 802: this second clause,  what is it doing?

lib/Infer/AliveDriver.cpp
```diff
@@ -777,6 +791,26 @@ souper::AliveDriver::translateDataflowFacts(const souper::Inst* I,
   }
 }
 
+bool
```
regehr, line 794: if this function cannot fail, please make it return void

test/Dataflow/demanded-bits10-invalid.opt
```diff
@@ -1,6 +1,6 @@
 
 
-; RUN: %souper-check %s > %t 2>&1
+; RUN: %souper-check -souper-use-alive %s > %t 2>&1
```
regehr, line 3: I think in this instance (and a bunch of others) you've made alive testing better, but you have made testing of the klee backend worse, because you're changing the query

instead of removing the existing souper-check invocation, I think you should make this test case (and the others) contain two RUN: lines, one of which targets one backend, the other of which targets the other one

regehr:
looks good other than 3 minor comments I just left

lib/Infer/AliveDriver.cpp
zhengyangl: Is there any other not-initialized-properly cases?

regehr: this should definitely be an assertion or an error check, not regular code

manasij7479:
Fixed

test/Dataflow/demanded-bits5-invalid.opt
```diff
@@ -1,6 +1,6 @@
 
 
-; RUN: %souper-check -infer-rhs -souper-only-infer-iN %s > %t 2>&1
+; RUN: %souper-check -souper-double-check -infer-rhs -souper-only-infer-iN %s > %t 2>&1
```
regehr, line 3: please don't delete the existing RUN: lines unless there's a good reason to do so

test/Dataflow/demanded-bits6.opt
```diff
@@ -1,8 +1,11 @@
 
 
+; RUN: %souper-check -souper-use-alive %s > %t 2>&1
+; RUN: %FileCheck %s < %t
 ; RUN: %souper-check %s > %t 2>&1
 ; RUN: %FileCheck %s < %t
 
+
```
regehr, line 8: look for an eliminate spurious white space changes in your patch, there are a few others

test/Dataflow/demanded-bits5-invalid.opt
manasij7479: double check just adds an extra check, not replacing anything existing.
If this test succeeds, the previous one would have succeeded.

test/Dataflow/demanded-bits6.opt
manasij7479: fixed

regehr:
thanks!

PR #793, opened 2020-10-04
Fallback to code generating if the dominance check fails

pgarba:
…to code generation

Small example to illustrate a missed optimization opportunity:

define dso_local i64 @_Z15MissingPattern4m(i64 %0) local_unnamed_addr #0 {
  %2 = xor i64 %0, -1276188105
  %3 = shl i64 %0, 1
  %4 = and i64 %3, -2552376210
  %5 = add i64 %2, 1276188121
  %6 = add i64 %5, %4
  %7 = add i64 %0, 16 ; Replacement origin that fails the dominance check
  %8 = and i64 %6, %7
  ret i64 %8
}

### DIFF
--- a/lib/Codegen/Codegen.cpp
+++ b/lib/Codegen/Codegen.cpp
@@ -90,8 +90,7 @@ llvm::Value *Codegen::getValue(Inst *I) {
       }
     }
     if (DebugLevel > 2)
-      llvm::errs() << "returning nullptr from getValue()\n";
-    return nullptr;
+      llvm::errs() << "dominance check failed. falling back into code generation\n";    
   }
 
   // otherwise, recursively generate code


pgarba:
Sorry the old test case didn't trigger the problem. This is the correct one:

```
; Function Attrs: nofree norecurse nounwind writeonly
define dso_local dllexport i64 @test(i64* nocapture %a, i64 %b) local_unnamed_addr #0 {
entry:
  %0 = add i64 %b, -3304
  %1 = shl i64 %0, 1
  %2 = xor i64 %0, -1276188105
  %3 = and i64 %1, -2552376210
  %4 = add i64 %2, 1276188121
  %5 = add i64 %4, %3
  %6 = add i64 %b, -3288
  store i64 %6, i64* %a, align 8
  ret i64 %5
}
```

The valid replacement found by souper:

```
%0:i64 = var
%1:i64 = add 18446744073709548312:i64, %0
%2:i64 = xor 18446744072433363511:i64, %1
%3:i64 = add 1276188121:i64, %2
%4:i64 = shl %1, 1:i64
%5:i64 = and 18446744071157175406:i64, %4
%6:i64 = add %3, %5
%7:i64 = add 18446744073709548328:i64, %0
cand %6 %7
```

Now when souper reaches the dominance check it fails because the origin of the candidate is `%6 = add i64 %b, -3288` which doesn't make any sense. We didn't stress test the patch but it worked for our scenario so we thought its a good starting point to find a proper solution for this problem. Can you explain why `%6 = add i64 %b, -3288` was chosen as the origin and if this is wanted or is there something going wrong when the candidate is created?

regehr:
closing since this patch will lead to generation of invalid IR, but I'll look into this later

Issue #779, opened 2020-06-26
PHI nodes not grouped at top of basic block!


regehr:
```
int a[], f[];
int b, c, d;
long e;
void g() {
  for (; a <= 6; b++) {
    c = 6;
    for (; c >= 0; c--) {
      d = 6;
      for (; d >= 0; d--)
        f[d] = f >= --e < d;
    }
  }
}
```
compiled like this:
```
SOUPER_VERIFY=1 SOUPER_DISABLE_LLVM_PEEPHOLES=true SOUPER_ENUMERATIVE_SYNTHESIS_MAX_INSTS=1 ~/souper-regehr/build/sclang -O3 foo.c
```

ends up with:
```
PHI nodes not grouped at top of basic block!
  %inc.lcssa = phi i32 [ %inc, %for.inc13 ]
label %for.cond.for.end14_crit_edge
fatal error: error in backend: module broken after (and probably by) Souper
clang-11: error: clang frontend command failed with exit code 70 (use -v to see invocation)
clang version 11.0.0 (https://github.com/regehr/llvm-project.git cbd3118ced344b1c1d7b7023a13abdbc68ba5624)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/regehr/souper-regehr/third_party/llvm-Debug-install/bin
clang-11: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang-11: note: diagnostic msg: /tmp/foo-4edb20.c
clang-11: note: diagnostic msg: /tmp/foo-4edb20.sh
clang-11: note: diagnostic msg: 

********************
```

PR #794, opened 2020-10-06
Move insert point after PHI nodes.

zoecarver:
If replacing a PHI node with another instruction, make sure the insertion point is after the last PHI node in that block. This prevents the assertion described in #779.

### DIFF
--- a/lib/Codegen/Codegen.cpp
+++ b/lib/Codegen/Codegen.cpp
@@ -99,6 +99,17 @@ llvm::Value *Codegen::getValue(Inst *I) {
   if (!V0)
     return nullptr;
 
+  // PHI nodes must be the first instructions in a basic block. If we're
+  // replacing a PHI node with another instruction, make sure it comes after the
+  // other PHI nodes.
+  if (ReplacedInst &&
+      isa<PHINode>(ReplacedInst->getNextNode())) {
+    Instruction *InsertPoint = ReplacedInst;
+    while (isa<PHINode>(InsertPoint))
+      InsertPoint = InsertPoint->getNextNode();
+    Builder.SetInsertPoint(InsertPoint->getNextNode());
+  }
+
   switch (Ops.size()) {
   case 1: {
     switch (I->K) {
--- a/test/Pass/insert-after-phis.ll
+++ b/test/Pass/insert-after-phis.ll
@@ -1,0 +1,46 @@
+; RUN: %opt -load %pass -souper -S -stats -o - %s 2>&1 | %FileCheck %s
+
+; The main thing being tested here is just that the verifier doesn't error
+; because a PHI node comes after another instruction.
+
+@c = local_unnamed_addr global i32 0, align 4
+@d = local_unnamed_addr global i32 0, align 4
+@e = local_unnamed_addr global i64 0, align 8
+@b = local_unnamed_addr global i32 0, align 4
+@a = global [1 x i32] zeroinitializer, align 4
+@f = global [1 x i32] zeroinitializer, align 4
+
+; CHECK: define void @g()
+define void @g() {
+entry:
+  br i1 icmp ule (i32* getelementptr inbounds ([1 x i32], [1 x i32]* @a, i64 0, i64 0), i32* inttoptr (i64 6 to i32*)), label %for.body.lr.ph, label %for.end14
+
+for.body.lr.ph:
+  %b.promoted = load i32, i32* @b, align 4
+  %e.promoted = load i64, i64* @e, align 8
+  br label %for.inc10
+
+for.inc10:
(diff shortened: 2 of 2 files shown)


test/Pass/insert-after-phis.ll
```diff
@@ -0,0 +1,46 @@
+; RUN: %opt -load %pass -souper -S -stats -o - %s 2>&1 | %FileCheck %s
```
zoecarver, line 1: I was having a bit of a hard time coming up with good tests. If I could tell souper to just run once, I think I could create a better test (that checks we replace a PHI instruction with an add, for example). 

Other than that, I can't come up with an input program where the PHI node is replaced with a non-phi node, and the whole thing isn't optimized away at some point later.

regehr:
ahhh... thanks for looking into this! we just haven't had time. I'll look at the code tomorrow, but I think for now it's sufficient testing to include the original bug trigger as a regression test. we'll get plenty more testing using our existing fuzzers.

regehr:
@zhengyangl do you have time to review this request?

lib/Codegen/Codegen.cpp
```diff
@@ -99,6 +99,22 @@ llvm::Value *Codegen::getValue(Inst *I) {
...
+  // other PHI nodes.
+  if (ReplacedInst &&
+      !ReplacedInst->isTerminator() &&
+      isa<PHINode>(ReplacedInst->getNextNode()) &&
+      I->K != Inst::Kind::Phi) {
+    Instruction *InsertPoint = ReplacedInst;
+    while (!InsertPoint->isTerminator() &&
+           isa<PHINode>(InsertPoint->getNextNode()))
+      InsertPoint = InsertPoint->getNextNode();
+    if (InsertPoint->isTerminator())
```
zhengyangl, line 113: Hi @zoecarver , 

Thanks for the patch! One concern:  as stated in [1], after calling IRBuilder.SetInsertPoint(Instruction *I), the new instructions will be inserted *before* instruction I. In this case, InsertPoint should be the instruction just next to the last phi in that basic block, right? However, this code would make InsertPoint the last PHI node and Instruction could still possibly be inserted between PHI nodes.

https://llvm.org/doxygen/classllvm_1_1IRBuilderBase.html#ae72654261eee259da46249dc919b029c

zoecarver: You're right. Fixed.

zhengyangl: Since IR builder is inserting before the ReplacedInst, it should work even when ReplacedInst is a terminator, right? Therefore, checking the terminator is not necessary here. 

And Souper won't create Phis, the check ```I->K != Inst::Kind::Phi``` is not necessary here.

I think 

```
if (ReplacedInst) {
  Instruction *InsertPoint = ReplacedInst;
  while (isa<PHINode>(InsertPoint) {
    InsertPoint = InsertPoint->getNextNode();
  }
  Builder.SetInsertPoint(InsertPoint);
}
```

should just do the work.

zoecarver: Good point. I'll update it with your suggestion. I think it would be good to keep the `isa<PHINode>(ReplacedInst->getNextNode())` check so that we don't unnecessarily update the `InsertPoint`, though.

zhengyangl:
Hi @zoecarver, one final note: if some replacement has Phi for ReplacedInst and a non-phi for the replacedInst->getNextNode, then we'll never reach the true branch, and by the definition of setInsertPoint() the new instruction would be inserted *before* the ReplaceInst (a phi node), which will cause trouble since we're trying to insert a non-phi instruction in the middle of the phi chunk.



`if (ReplacedInst && isa<PHINode>(Replacement))` would be bug-free.

zoecarver:
> Hi @zoecarver, one final note: if some replacement has Phi for ReplacedInst and a non-phi for the replacedInst->getNextNode, then we'll never reach the true branch, and by the definition of setInsertPoint() the new instruction would be inserted before the ReplaceInst (a phi node), which will cause trouble since we're trying to insert a non-phi instruction in the middle of the phi chunk.

Sorry for the delayed response. My reasoning is that `ReplacedInst` will always be removed after the new instruction is added. In this case, we don't want to change the `InsertPoint` unnecessarily because even though for a short period of time there will be a non-phi node before a phi node, the replaced phi node will be removed before the verifier ever runs.

regehr:
hey @zhengyangl are you OK with this patch now? it looks fine to me

zhengyangl:
LGTM

regehr:
thanks for the bugfix!

zoecarver:
> thanks for the bugfix!

Of course. Thanks (both of you) for reviewing/merging!

PR #795, opened 2020-10-07
[NFC] Include <vector> in Verification.h.

zoecarver:
Using my toolchain, none of these previously imported headers import `vector`. It seems like it makes sense to include it directly here.


manasij7479:
LGTM

PR #796, opened 2020-10-13
Alive bump

manasij7479:
Depends on phi PR

### DIFF
--- a/lib/Infer/AliveDriver.cpp
+++ b/lib/Infer/AliveDriver.cpp
@@ -128,8 +128,7 @@ public:
   template <typename T>
   void assume(T &&V) {
     auto AI =
-      std::make_unique<IR::Assume>(*std::move(V), /*if_non_poison=*/ true);
-      //TODO: FIgure out whether if_non_poison can be uconditionally true
+      std::make_unique<IR::Assume>(*std::move(V), IR::Assume::Kind::IfNonPoison);
     F.getBB("").addInstr(std::move(AI));
   }
 
--- a/build_deps.sh
+++ b/build_deps.sh
@@ -28,7 +28,7 @@ llvm_repo=https://github.com/regehr/llvm-project.git
 llvm_commit=disable-peepholes-v03
 klee_repo=https://github.com/rsas/klee
 klee_branch=pure-bv-qf-llvm-7.0
-alive_commit=92ad3f5f2f963ef5bd43f71a4137427bd6cce51b
+alive_commit=v1
 alive_repo=https://github.com/manasij7479/alive2.git
 z3_repo=https://github.com/Z3Prover/z3.git
 z3_commit=z3-4.8.8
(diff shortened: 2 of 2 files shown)


lib/Infer/AliveDriver.cpp
```diff
@@ -351,18 +351,27 @@ synthesizeConstantUsingSolver(tools::Transform &t,
...
+                                 std::vector<Inst *> ExtraInputs)
     : LHS(LHS_), PreCondition(PreCondition_), IC(IC_) {
   IsLHS = true;
   InstNumbers = 101;
   //FIXME: Magic number. 101 is chosen arbitrarily.
   //This should go away once non-input variable names are not discarded
 
+  // If alive ever supports src and tgt with different number of arguments in future, this can go away.
+  for (auto Extra : ExtraInputs) {
+    if(!translateAndCache(Extra, LHSF, LExprCache)) {
```
regehr, line 364: space after if

lib/Infer/AliveDriver.cpp
```diff
@@ -814,7 +826,114 @@ IR::Type &souper::AliveDriver::getType(int n) {
...
+
+    // Recursively call ReplacePhi, because Result might have Phi`s
+    return Result.ReplacePhi(IC, Change);
+  }
+  bool operator == (const RefinementProblem &P) const {
+    return LHS == P.LHS && RHS == P.RHS && Pre == P.Pre;
+  }
+  struct Hash
+  {
+    std::size_t operator()(const RefinementProblem& P) const
```
regehr, line 883: " &P" rather than "& P"
try to be consistent with these small things or else they distract from your code

regehr: also, overloading () deserves a comment. what's going on here?

lib/Infer/AliveDriver.cpp
```diff
@@ -814,7 +826,114 @@ IR::Type &souper::AliveDriver::getType(int n) {
...
+    {
+      return std::hash<Inst *>()(P.LHS)
+             ^ std::hash<Inst *>()(P.RHS) << 1
+             ^ std::hash<Inst *>()(P.Pre) << 2;
+    }
+  };
+
+};
+
+std::unordered_set<RefinementProblem, RefinementProblem::Hash>
```
regehr, line 893: is this function performance-sensitive? if so, allocate the set in the caller and mutate it, rather than making a copy by value on return

lib/Infer/AliveDriver.cpp
manasij7479: &P is addressed in the other PR.
overloading () for a hash struct is one idiomatic way to use custom types in STL containers like unordered_map or unordered_set.

lib/Infer/AliveDriver.cpp
manasij7479: I don't think it is performance sensitive. Solver time dominates all this.

lib/Infer/AliveDriver.cpp
manasij7479: Ah these are addressed in the other PR, one commit overlaps.

manasij7479:
Rebased.
docker needs to be updated for CI to pass from this commit onwards.

regehr:
LGTM, thanks!

PR #799, opened 2020-10-23
Translate BlockPCs

manasij7479:


regehr:
any comments @zhengyangl? it seems fine to me

lib/Infer/AliveDriver.cpp
```diff
@@ -890,20 +894,27 @@ struct RefinementProblem {
...
+                               &ConstMap, false);
+    }
 
     // Recursively call ReplacePhi, because Result might have Phi`s
     return Result.ReplacePhi(IC, Change);
   }
   bool operator == (const RefinementProblem &P) const {
-    return LHS == P.LHS && RHS == P.RHS && Pre == P.Pre;
+    return LHS == P.LHS && RHS == P.RHS &&
+           Pre == P.Pre && BPCs.size() == P.BPCs.size();
```
zhengyangl, line 908: This size() comparison is too loose and may cause tricky failures in future. Could we implement operator= for BPCs and checks the equality for each element?

lib/Infer/AliveDriver.cpp
```diff
@@ -946,23 +957,33 @@ std::unordered_set<RefinementProblem, RefinementProblem::Hash>
   std::unordered_set<RefinementProblem, RefinementProblem::Hash> Result;
 
   for (auto Change : ChangeList) {
-    Result.insert(P.ReplacePhi(IC, Change));
+    auto Goal = P.ReplacePhi(IC, Change);
+    // Consider switching to better data structures for dealing with BPCs
+    for (auto [Block, Pred] : Change) {
+      for (BlockPCMapping BPC : Goal.BPCs) {
```
zhengyangl, line 963: please use reference when possible to avoid copying

lib/Infer/AliveDriver.cpp
```diff
@@ -946,23 +957,33 @@ std::unordered_set<RefinementProblem, RefinementProblem::Hash>
...
+    Result.insert(Goal);
   }
-
   return Result;
 }
 
 }
 bool souper::isTransformationValid(souper::Inst* LHS, souper::Inst* RHS,
                                    const std::vector<InstMapping> &PCs,
+                                   const souper::BlockPCs & BPCs,
```
zhengyangl, line 978: eliminate space between & and BPcs

zhengyangl (changes requested):
Looks good in general:) A few places need to be fixed.

lib/Infer/AliveDriver.cpp
regehr: also, a couple lines above, always put the "*" next to the variable name, not next to the type, since in some circumstances that is misleading

lib/Infer/AliveDriver.cpp
manasij7479: Good point. Being slightly slower to avoid tricky failures seems reasonable.

lib/Infer/AliveDriver.cpp
```diff
@@ -868,11 +868,15 @@ struct RefinementProblem {
...
   souper::Inst *RHS;
   souper::Inst *Pre;
+  BlockPCs BPCs;
 
   RefinementProblem ReplacePhi(souper::InstContext &IC, std::map<Block *, size_t> &Change) {
     std::map<souper::Block *, std::set<souper::Inst *>> Phis;
     collectPhis(LHS, Phis);
     collectPhis(Pre, Phis);
+    for (auto &BPC : BPCs) {
+      collectPhis(BPC.PC.LHS, Phis);
```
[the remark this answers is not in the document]
manasij7479: Seems to be just True or False.

regehr:
LGTM, thanks!

PR #804, opened 2020-11-20
Precondition inference for multiple variables

manasij7479:


manasij7479:
@zhengyang92 @regehr  I think I am misusing the Solver API here.
Can I get both isValid and isSatisfiable from a single solver abstraction layer?
One deals with the validity of transformations, the other about the satisfiability of SMT queries.
Need both for this.

zhengyang92:
isSatisfiable is visible from SMTSolver, and it should not have any problem. I don't see why you can't call isValid() in abstractPrecondition()?

manasij7479:
I want to call isValid from Preconditions.cpp. I'm currently doing that by passing `this` from abstractPrecondition. But that seems like a weird thing to do.

zhengyang92:
you must forget git add Preconditions.cpp.  :)

manasij7479:
Seems to be in the PR: https://github.com/google/souper/blob/2ca2fbb5c3eb168691648778b2a9265b16c61cb8/lib/Infer/Preconditions.cpp

zhengyang92:
Oh, I was looking at the diff of b1b4ca1, sorry. What about just make inferAbstractKBPreconditions() a private method for BaseSolver?

manasij7479:
Will do that if there isn't a better solution. 
I see that we have been doing that for a while.
All the methods for the analysis verification work seem to be there for similar reasons.

regehr:
LGTM

PR #805, opened 2020-12-01
Generalization tool

manasij7479:


regehr:
ok good, I like that this is removing code from Solver.cpp

tools/generalize.cpp
```diff
@@ -0,0 +1,163 @@
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/GraphWriter.h"
+#include "llvm/Support/KnownBits.h"
+
+#include "souper/Infer/Preconditions.h"
+
```
regehr, line 6: does this blank line among souper includes serve a purpose?

tools/generalize.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+}
+
+// TODO: Return modified instructions instead of just printing out
+void RemoveLeafAndGeneralize(InstContext &IC,
+                               Solver *S, ParsedReplacement Input) {
+
+  if (DebugLevel > 1) {
+  llvm::errs() << "Attempting to generalize by removing leaf.\n";
```
regehr, line 62: indentation error

tools/generalize.cpp
```diff
@@ -0,0 +1,163 @@
...
+  // TODO: Do not generalize by removing leaf if LHS has one inst.
+
+  std::map<Inst *, std::set<Inst *>> Uses;
+
+  std::vector<Inst *> Stack{Input.Mapping.LHS, Input.Mapping.RHS};
+  // TODO: Find uses in PCs/BPCs
+
+  std::set<Inst *> Visited;
+  while (!Stack.empty()) {
+   auto Current = Stack.back();
```
regehr, line 73: indentation error.

if you're having trouble with basic code formatting, why not get into the habit of running clang-format on new files before contributing them?

test/Generalize/leaf.opt
```diff
@@ -0,0 +1,22 @@
+; REQUIRES: solver, synthesis
```
regehr, line 1: since this test is just looking for a bunch of LGTMs out of a two-step process, it's not really possible to see what behavior it's actually testing. please modify it to only run your new tool, and then check that its output is correct without invoking a second tool, so we can look at the test and understand what's going on. also do this to the other test.

manasij7479: OK, sounds good.

tools/generalize.cpp
manasij7479: Good idea, I'll do them cmake tweaking required for running `ninja format` or `make format`.

regehr: for now, please only do this on new files that are being added to the repo. doing this for existing files would be fun, but it would make it effectively impossible to ever merge any existing branch. we can only do this at a time where we're certain that no non-trivial branches exist. at least one exists right now, and probably more.

PR #809, opened 2020-12-04
souper-combine pass template

manasij7479:
Build system is intentionally decoupled from Souper.

### DIFF
--- a/tools/pass-generator/CMakeLists.txt
+++ b/tools/pass-generator/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required(VERSION 3.7)
+project(souper-combine)
+
+find_package(LLVM 11.0 REQUIRED CONFIG)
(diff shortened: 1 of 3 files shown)


regehr:
does this want to be a part of souper? seems like it's probably better to just work on it in your own repo.

manasij7479:
It is not a standalone tool, it just has the plumbing for a `blank` instcombine. 
By itself, this does not perform any transformation.
The transformation code generation (not submitted yet) depends on Souper.
That part would take a list of souper generated transforms and generate some C++ code which has to be #included/copied into this template. Maybe there will be a bootstrap script to do it automatically (and test it). 
So, I think it makes sense to put this inside the Souper repository. Don't have a strong opinion about it though.

regehr:
I'd say just do this outside of souper, if it stabilizes into something we want, we can merge it later

PR #813, opened 2021-04-21
Misc fixes to make Souper work better with SPEC 2017

manasij7479:


lib/Pass/Pass.cpp
```diff
@@ -277,7 +291,11 @@ struct SouperPass : public ModulePass {
             EC == std::errc::value_too_large) {
           continue;
         } else {
-          report_fatal_error("Unable to query solver: " + EC.message() + "\n");
```
regehr, line 280: I want you to print a strongly worded message here, this can't be silent

manasij7479: done

PR #822, opened 2021-05-18
Switch to using assume for encoding PCs

manasij7479:


regehr:
please add an example that fails before this patch, as a regression test

manasij7479:
> please add an example that fails before this patch, as a regression test

done

PR #838, opened 2021-06-02
Do not add SAddO and family as synthesis components

manasij7479:


regehr:
please add the crash trigger as a regression test

manasij7479:
> please add the crash trigger as a regression test

done

PR #840, opened 2021-06-02
Cost fudge option

manasij7479:


regehr:
is this tested? does it work as intended?

manasij7479:
The number of guesses generated checks out. We'll have to do a deeper inspection about the 'quality' of the guesses at some point. 
For this example:
%0:i32 = var
%1 = add %0, 1
%2 = and %1, 5
%3 = or %2, 42
infer %3

Here are the number of guesses generated with a 1 instruction synthesis bound.
N=0 -- 49
N=1 -- 114
N=2 -- 221
N=3 -- 328
N=4 -- 497
N=5 -- 526
N=6 -- 563 
(doesn't increase after this)

PR #843, opened 2021-06-07
Enumerator bug fixes

zhengyang92:
First, posting the harvested souper IR for the test case:

```
%0 = block 11
%1:i8 = var
%2:i1 = ne 0:i8, %1
blockpc %0 0 %2 1:i1
%3:i1 = ne 1:i8, %1
blockpc %0 0 %3 1:i1
%4:i1 = ne 2:i8, %1
blockpc %0 0 %4 1:i1
%5:i1 = ne 3:i8, %1
blockpc %0 0 %5 1:i1
%6:i1 = ne 4:i8, %1
blockpc %0 0 %6 1:i1
%7:i1 = ne 5:i8, %1
blockpc %0 0 %7 1:i1
%8:i1 = ne 6:i8, %1
blockpc %0 0 %8 1:i1
%9:i1 = ne 7:i8, %1
blockpc %0 0 %9 1:i1
%10:i1 = ne 8:i8, %1
blockpc %0 0 %10 1:i1
%11:i1 = ne 9:i8, %1
blockpc %0 0 %11 1:i1
blockpc %0 1 %1 9:i8
blockpc %0 2 %1 8:i8
blockpc %0 3 %1 7:i8
blockpc %0 4 %1 6:i8
blockpc %0 5 %1 5:i8
blockpc %0 6 %1 4:i8
blockpc %0 7 %1 3:i8
blockpc %0 8 %1 2:i8
blockpc %0 9 %1 1:i8
blockpc %0 10 %1 0:i8
%12:i8 = phi %0, 0:i8, 0:i8, 0:i8, 0:i8, 0:i8, 0:i8, 0:i8, 0:i8, 0:i8, 0:i8, 1:i8
infer %12
```

This is a rather interesting test case, note here the result value %12 does not have a dependency (through Def/Use chain) on the input %1. The input %1 is only used in blockpcs. Before this PR the synthesis candidates are only harvested through the Def/Use chain, and that's the reason why we're not seeing any guesses that uses %1. This patch makes the extractor not only harvest from LHS, but also from the PCs and BlockPCs.

Another issues that blocks Souper from synthesizing our expected result is that our naive cost model cannot capture the fact that eliminating phi is beneficial, and the LHS root is harvested as a synthesis candidate. As the consequence of those issues, the LHS is pushed into the guesses list and as it contains only a single phi (cost=1), after sorting by cost, all the non-trivial guesses will be checked after checking LHS itselt, and as a result, Souper just returns the phi itself.  To solve this issue we can either increase the cost of Phi to a larger value or just block the extractor from harvesting the LHS root. I chose the second solution for this pull request.


regehr:
I'll review soon. thanks!

PR #844, opened 2021-06-08
Default to evaluating first argument of phi

manasij7479:
Fixes the following crash.
This takes a long time to complete when it doesn't crash, so I'm not adding a regression test.
Even with the skip-solver option, just generating+pruning guesses takes minutes. This is suspicious, and has to be looked at.

`./souper-check -solver-timeout=15 -infer-rhs -souper-double-check -souper-enumerative-synthesis-max-instructions=2 foo.opt -souper-dataflow-pruning
`
%0 = block 2
%1 = block 3
%2 = block 3
%3:i32 = phi %2, 0:i32, 1:i32, 1:i32
%4:i32 = phi %1, %3, 1:i32, 1:i32
%5:i32 = phi %2, 0:i32, 0:i32, 1:i32
%6:i32 = addnw 1:i32, %5
%7:i32 = phi %1, %5, %5, %6
%8:i1 = eq 1:i32, %7
%9:i32 = zext %8
%10:i32 = addnw %4, %9
%11:i32 = phi %0, %10, 0:i32
%12:i1 = eq 2:i32, %11
infer %12

### DIFF
--- a/build_deps.sh
+++ b/build_deps.sh
@@ -25,7 +25,7 @@ ncpus=$(command nproc 2>/dev/null || command sysctl -n hw.ncpu 2>/dev/null || ec
 hiredis_commit=667dbf536524ba3f28c1d964793db1055c5a64f2
 llvm_repo=https://github.com/regehr/llvm-project.git
 # llvm_commit specifies the git branch or hash to checkout to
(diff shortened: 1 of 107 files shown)


regehr:
you need to add the test case and, if it's too slow, protect it behind one of our "slow test" flags

PR #845, opened 2021-06-09
Set available=false for overflow instructions when parsing

zhengyang92:
Note with this patch, we still can't get the result we want for the case

```
%0:i32 = var
%1:i32 = var
%2:i33 = ssub.with.overflow %0, %1 (hasExternalUses)
%3:i32 = extractvalue %2, 0:i32 (hasExternalUses)
%4:i1 = ule %3, 31:i32
pc %4 1:i1
%5:i32 = subnw 31:i32, %3
%6:i32 = lshr 1:i32, %5
%7:i1 = eq 0:i32, %6
infer %7
%8:i32 = sub %0, %1
%9:i1 = ult %8, 31:i32
result %9
```

This is because the RHS we want has the form (cmp hole, const); the enumerator is not handling such forms since no concrete widths are available when generating the root.

### DIFF
--- a/lib/Infer/EnumerativeSynthesis.cpp
+++ b/lib/Infer/EnumerativeSynthesis.cpp
@@ -323,6 +323,7 @@ bool getGuesses(const std::set<Inst *> &Inputs,
 
           if ((*I)->Width == 0 && (*J)->Width == 0) {
             // TODO: support (cmp hole, hole);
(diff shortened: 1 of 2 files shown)


regehr:
can you please add a regression test?

zhengyang92:
@regehr  regression test added :)

regehr:
thanks!

PR #855, opened 2022-04-14
Add instructions for use with Rust/Cargo

saethlin:
@regehr I suspect the equivalent of these instructions will work to get a follow-on superoptimizer working with Rust/Cargo. If anyone tries these out please do let me know how it goes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ $ ./build_deps.sh $buildtype $extra_cmake_flags
 ```
 $ mkdir /path/to/souper-build
 $ cd /path/to/souper-build
-$ cmake -DCMAKE_BUILD_TYPE=$buildtype /path/to/souper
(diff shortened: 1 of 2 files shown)


chadbrewbaker:
Not sure this is how I would go about it. Might be better to do a LLVM IR build like you do for GraalVM/Klee, then whitelist the files you want Souper to optimize? Doing Souper on cold paths doesn't make computational sense.

One feature I would really love is a --souper-verbose-asm that pretty prints transforms it found in the assembler file.

My ideal workflow is compile for profiling, run profiling load to whitelist 99% runtime hot functions, send just those to Souper with a hyperfine style report of the improvement.

Another thing I Souper may be doing is screwing up cache alignments and creating more branch mispredictions?

@ornata - is there an LLVM annotation to keep functions on cache line boundaries, or do thin-BOLT where you re-order small hot functions used together?

